PR #14, opened 2021-04-09
Make "attribute value" slot abstract

### LINKED ISSUE #12: attribute value should be abstract, doi should be something concrete
Whereas [named thing](https://github.com/microbiomedata/nmdc-schema/blob/eb0da4efe9b9796b3677ac5e776e767d81ad66b0/src/schema/core.yaml#L76) is abstract, [attribute value](https://github.com/microbiomedata/nmdc-schema/blob/eb0da4efe9b9796b3677ac5e776e767d81ad66b0/src/schema/core.yaml#L85) is not abstract. However, [doi](https://github.com/microbiomedata/nmdc-schema/blob/main/src/schema/nmdc.yaml#L1116) is the only slot in the NMDC schema with a range of attribute value, rather than a range that is_a attribute value such as [text value](https://github.com/microbiomedata/nmdc-schema/blob/eb0da4efe9b9796b3677ac5e776e767d81ad66b0/src/schema/core.yaml#L135). I recommend that attribute value be made `abstract: true` and that the range of doi be made `text value` or another concretion of `attribute value`.

dwinston:
Change sole slot with "attribute value" range to be "text value".
Can change doi slot to be another subclass of "attribute value" later.

Closes #12

### DIFF
--- a/src/schema/core.yaml
+++ b/src/schema/core.yaml
@@ -89,6 +89,7 @@ classes:
     is_a: ontology class
 
   attribute value:
+    abstract: true
     description: >-
       The value for any value of a attribute for a sample. This object can hold both the un-normalized atomic
       value and the structured value
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -550,7 +550,7 @@ slots:
 
   doi:
     is_a: attribute
-    range: attribute value
+    range: text value
 
   add_date:
     range: string


wdduncan:
I use the classes in the `nmdc.py` for ETL. So, I am not sure how this will affect ETL.  
Marking as `draft` for now.

wdduncan:
@dwinston This PR is quite old. Going to close for now. You can re-open if you wish.

PR #15, opened 2021-04-09
Remove GOLD paths from study

### LINKED ISSUE #10: conflicting gold paths in biosample and study
Currently, both biosample and study include slots for the five GOLD path fields (ecosystem, ecosystem_category, ecosystem_type, ecosystem_subtype, specific_ecosystem):
- [in biosample](https://github.com/microbiomedata/nmdc-schema/blob/88ba05666f71739c7946818a92629977dd3c5682/src/schema/nmdc.yaml#L433)
- [in study](https://github.com/microbiomedata/nmdc-schema/blob/88ba05666f71739c7946818a92629977dd3c5682/src/schema/nmdc.yaml#L473)

This is an issue because it's not clear if all biosamples must have the same values as the study, if the biosample values "override" the "default" study values, etc.

It would be better to (a) have them exclusively either in study or in biosample, or (b) to name them differently so as to clarify, e.g. prefix the slots in study with "default_", "assumed_", etc.

@cmungall @dehays

dwinston:
Removes conflict with GOLD paths given in a study's biosample records

Closes #10

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -277,11 +277,6 @@ classes:
     description: >-
       A study summarizes the overall goal of a research initiative and outlines the key objective of its underlying projects.  
     slots:
-      - ecosystem
-      - ecosystem_category
-      - ecosystem_type
-      - ecosystem_subtype
-      - specific_ecosystem
       - principal investigator name
       - doi
       - type  # custom slot that specifies object type


wdduncan:
Waiting on input from Reddy.

TBKReddy:
As Chris commented on #10 these will be valuable to have on the Study as it helps people to search for studies using these terms. 
My suggestion is to keep these terms on Study. Think about a different approach to get these values populated/refreshed based on the underlying samples. Currently we are not doing this in GOLD, but since you are at it and may have bandwidth to consider it.

turbomam:
The GOLD path slots still appear on both biosamples and studies. Based on discussion above, should I close this PR?

The last issue comment, https://github.com/microbiomedata/nmdc-schema/issues/10#issuecomment-819812309 says

> We have decided to keep them.

PR #16, opened 2021-04-09
Remove conflicting "principle investigator name" slot

### LINKED ISSUE #11: conflicting principal_investigator_name
`principal_investigator_name` is a slot in both [Study](https://github.com/microbiomedata/nmdc-schema/blob/eb0da4efe9b9796b3677ac5e776e767d81ad66b0/src/schema/nmdc.yaml#L478) and [OmicsProcessing](https://github.com/microbiomedata/nmdc-schema/blob/eb0da4efe9b9796b3677ac5e776e767d81ad66b0/src/schema/nmdc.yaml#L519).

The issue is similar to #10: must an omics processing's PI be the same as for the study? Does one "override" the other in a given context? Again, it would be better to (a) have it exclusively either in study or in omics processing (I'm guessing study makes more sense), or (b) to name them differently so as to clarify, e.g. "study PI" vs "processing PI".

@dehays @cmungall

dwinston:
Close #11

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -323,7 +323,6 @@ classes:
       - ncbi_project_name
       - omics type
       - part of
-      - principal investigator name
       - processing_institution
       - type  # custom slot that specifies object type
     broad_mappings:


wdduncan:
Marking this as draft until we have more input. We should probably discuss the Reddy and Jagadish too.

cc @dehays @cmungall @jagadishcs

turbomam:
The last issue comment https://github.com/microbiomedata/nmdc-schema/issues/11#issuecomment-819830057 says:

>  We have decided to keep them.

turbomam:
Although I have to review whether we still have soem reducancy between a principle investigator name slot and the credit association terms

PR #49, opened 2021-05-28
Include CRC32 checksum as a required attribute of data objects

jbeezley:
I'm creating this for discussion.  Let me know if there is a more appropriate venue to raise it.

As part of https://github.com/microbiomedata/nmdc-server/issues/354, I have created a [service](https://github.com/microbiomedata/nmdc-server/pull/393) for streaming zip files from data objects.  This works as is with the current metadata, but it has to compute CRC32 checksums for the zip file on the fly.  This means that it can't support http range requests.  Without http range request support, the bulk downloading service cannot support retries, which is an important feature when downloading large files potentially over the course of several days.  From my experience, the chance of a random failure just due to package loss is extremely high for such long lived requests, and without retries, the download has to start from the beginning.

The attribute is made required because the feature won't work if even one file in the archive is missing it.  This will require any data object producer to provide the CRC32 value and back-filling any existing data objects.


turbomam:
@jeffbaumes @dwinston how do you feel about moving forward with this?

I like it, but I would change the slot name to `crc32_checksum`, although the title could be 'crc32 checksum'

we still need to determine what classes should take this new slot.

turbomam:
I don't think making it a sub-property of `attribute` helps much.

turbomam:
@dwinston , @cmungall , @jeffbaumes I would like to close this PR. 

In fact, that's what I'm going to do if I don't get a response before Friday February 24th (since Chris will be out of town for another week or so)

jeffbaumes:
I'm ok with closing since it's not something we plan to tackle now. I do think it's worth keeping in the back of our minds that a checksum would make large downloads less painful as Jon suggested.

Issue #62, opened 2021-06-11
Share examples of repaired NCBI/INSDC Biosample metadata


turbomam:
Efforts for mapping categorical data from the NCBI/INSDC Biosample metadata to OBO foundry terms has been taking place in several branches or forks of [INCATools/biosample-analysis](https://github.com/turbomam/scoped-mapping) and [turbomam/scoped-mapping](https://github.com/turbomam/scoped-mapping) _at least_. That's partially because these techniques will be applied to other projects besides NMDC.

Existing efforts have been batchwise or notebook based. We are planning on moving towards a _repair API_.

For now, we need a place to store mappings to that NMDC personal can review them and suggest changes to our strategies.

CCing @cmungall @wdduncan

turbomam:
My most recent work for this is found in the mined EnvO and PO terms in the DucDB file
- https://portal.nersc.gov/project/m3408/biosamples_duckdb/
- [makefile target in external-metadata-awareness](https://github.com/microbiomedata/external-metadata-awareness/blob/7258b267014a1043d2c1448e65e1f4c23b5b5210/Makefiles/ncbi_metadata.Makefile#L57)

turbomam:
closing because this sort of thing will be dealt with in 
- https://github.com/microbiomedata/nmdc-ingest-agent

ideally without per-project mapping files

PR #63, opened 2021-06-11
Non-SSSOM, per-biosample MIxS/GOLD env paths (data only... no code?)

turbomam:
Issue #62 

Creates a folder for non-sssom mappings, moves existing non-sssom mappings out of sssom directory, and adds one new "repaired" file for soil package samples.


turbomam:
This one of many attempts to automate and or record the EnvO terms that could be used in MIxS environmental triad slots, on an environmental package basis.

turbomam:
Various early attempts to relate MIxS triad values to GOLD paths

Issue #19, opened 2021-04-01
Move PI profile images to official metadata schema


jeffbaumes:
This is currently in the portal repo. Should we consider something like gravatar? Unsure if ORCID provides public access to a profile image but that might be ideal.

wdduncan:
@dehays yes, that is what I was thinking.

dwinston:
So the plan is to
1. eliminate core/person_value class
2. give core/person class an orcid slot
3. give core/person class an image_url slot
4. remove nmdc/study class' principal_investigator_name slot
5. add nmdc/study class principal_investigator slot with range core/person

Is that right?

jeffbaumes:
@wdduncan I added you as an assignee since I don't think @jbeezley can make the actual query changes himself. Please correct this or delegate if I'm off here.

ssarrafan:
Adding comments from email exchange for reference:

Agree with your comments here David (and yours Kjiersten). I commented in parallel on GitHub, but the gist is that we need to resolve which fields we can and can't expect to require. Many "required" things for the portal could be made optional if needed.

https://github.com/microbiomedata/nmdc-schema/issues/41


On Wed, Apr 28, 2021 at 1:31 PM Kjiersten Fagnan <kmfagnan@lbl.gov> wrote:
I support the approach David laid out for what fields are required vs optional. I can add the following comments to the ticket, but we seem to be getting into this via email. 

Could we create some default values for the portal to populate the page - avatar, URL, DOI and scientific objectives would be harder if not impossible. 

In the future, when contributors are providing data to NMDC, could we also collect - photo, website URL, etc as part of the submission process - or perhaps give the PI the ability to add this themselves? This depends on having some level of access controls (different roles in the data portal than exist right now). Maybe this is part of working directly with the PIs to get their help on the study/data landing pages?

Kjiersten

On Wed, Apr 28, 2021 at 10:20 AM David Hays <dehays@lbl.gov> wrote:
Bill and Emiley said:

- https://github.com/microbiomedata/nmdc-schema/issues/41 I do not have access to the data for some of the fields that Jeff is requesting. I can give an estimate until we track the data.
> @Emiley Eloe-Fadrosh any idea of who to follow up on with to get access to the data needed?  

- https://github.com/microbiomedata/nmdc-schema/issues/19 Again, I do not have access to images of PIs.
> @Emiley Eloe-Fadrosh any idea of who to follow up on with to get access to the PI images needed?

[...]

PR #68, opened 2021-06-15
add person/profile_image

dwinston:
related to #19. cc @dehays

- move `data object` class to `core.yaml`
- add `profile image` slot to `person value`, with a range of `data object` (which has a `url`).


dwinston:
@wdduncan what's the difference between the `person` and `person value` classes in `core.yaml`? Did I use the right one here?

src/schema/core.yaml
```diff
@@ -74,6 +74,29 @@ types:
     
 classes:
 
+  data object:
+    is_a: named thing
+    in_subset:
```
wdduncan, line 82: I am pretty sure a `subsets` section will need to be added to `core.yaml` or even the the `subset` may be deleted. For example, here is the `subsets` section in `nmdc.yaml`:
```yaml
subsets:
  sample subset:
    description: >-
      Subset consisting of entities linked to the processing of samples. 
      Currently, this subset consists of study, omics process, and biosample.
  
  data object subset:
    description: >-
      Subset consisting of the data objects that either inputs or outputs of processes or workflows.
```

dwinston: yeah, moved this in a commit while you were drafting this comment. :)

Issue #74, opened 2021-06-21
run tests for invalid data


wdduncan:
For invalid datasets in `test/data`, add target to `makefile` to check that invalid data is confirmed to be invalid.

As a convention, prefix that name of the invalid dataset with `invalid_`; e.g., `invalid_study_test.json`.

cc @turbomam

wdduncan:
See this stackoverflow article for pointers:

https://stackoverflow.com/questions/2188376/how-to-have-gnu-make-explicitly-test-for-failure

turbomam:
in addition to prefixing the negative tests with "invalid_", can we also put them in an `invalid` subdirectory?

wdduncan:
> in addition to prefixing the negative tests with "invalid_", can we also put them in an invalid subdirectory?
I'd prefer not.

wdduncan:
Yes. That was going to be my approach.

wdduncan:
@turbomam I think @sujaypatil96 fixed this :)

Closing ticket. Reopen if needed.

PR #79, opened 2021-06-24
negated invalid tests pass #74

turbomam:


turbomam:
I just looked though the Makefile and test/data/* and I think everything provided by the PR is already present in the current main

Issue #72, opened 2021-06-21
addition to file_type_enum for NOM and metaB file types


dehays:
modify the file_type_enum definition in nmdc.yml (https://github.com/microbiomedata/nmdc-schema/blob/main/src/schema/nmdc.yaml - line 376).  

Add 2 enum values with descriptions:

```
data_object_type: GC-MS Metabolomics Results
description: GC-MS-based metabolite assignment results table
```

```
data_object_type: Natural Organic Matter Results
description: FT-MS high resolution molecular formula assignment results table
```

turbomam:
I added the two blocks into nmdc.yaml as is. Did not pass tests.

```
enums:
  file type enum:
    permissible_values:
      assembly_contigs.fna: 
        description: Final assembly contigs fasta.
.
.
.
      NOM:
        data object type: Natural Organic Matter Results
        description: FT-MS high resolution molecular formula assignment results table
      metaB:
        data object type: GC-MS Metabolomics Results
        description: GC-MS-based metabolite assignment results table
```

> ValueError:  Unknown argument: data object type = 'Natural Organic Matter Results'

turbomam:
@dehays fixed in issue #83 / branch https://github.com/microbiomedata/nmdc-schema/tree/closes-%2372

```
enums:
  file type enum:
    permissible_values:
      assembly_contigs.fna: 
        description: Final assembly contigs fasta.
.
.
.
      GC-MS Metabolomics Results:
        GC-MS-based metabolite assignment results table
      Natural Organic Matter Results:
        FT-MS high resolution molecular formula assignment results table
```

**Where would the predicate/property `data object type` appear now? JSON data, or in the schema?**

dehays:
@turbomam data_object_type will appear on instances of data object where data_object_type has the file_type_enum as its range.  Not sure why Bill called the property data_object_type rather than file_type following the enum naming.

wdduncan:
It was called `data object type` b/c that was the original name that Chris gave it.

wdduncan:
@dehays Do we need to implement this for the next sprint?

dehays:
@wdduncan I am going to close this one.  The change would be to update the entire file type enum definition.  And before setting the range on data object type to the enum - we need to consider the mapping implementation that Donny used for data object type on the current data objects.

PR #82, opened 2021-06-25
added NOM and metaB to file type enum for #72

turbomam:
Also addd requested `description` and `data_object_type` annotations

`data_object_type property` already existed but doesn't seem to be mentioned anywhere else in the schema yet. I'm not sure I completely understand when to use whitespaces vs underscores for delimiters.


wdduncan:
Per discussion with @dehays closing this PR.

Issue #84, opened 2021-06-28
Add links for NMDC data objects in README


turbomam:
From @cmungall

dwinston:
Can link to (or copy from) https://github.com/microbiomedata/nmdc-runtime#data-exports

PR #76, opened 2021-06-23
Added images to README.md for #73

turbomam:
NMDC image may be too big

Still need to find newer/better? schema image **or** validate the current one.

PR #85, opened 2021-06-28
Issue #84 - add links for downloading data collections to README

turbomam:
Are you guys OK with the way I have exposed the NMDC JSON downloads (to the public?)

Please read the [two notes](https://github.com/microbiomedata/nmdc-schema/blob/issue-84/README.md#collections-within-the-schema).


turbomam:
Please also see #76 . It would be nice to close or address both of these at the same time.

dwinston (changes requested):
No, these are brittle links. I suggest we link to a timestamped [DRS *bundle*](https://ga4gh.github.io/data-repository-service-schemas/preview/release/drs-1.0.0/docs/#_drs_datatypes) that has, as its `contents`, a DRS *blob* for each bzipped json lines collection.

We can have a table of summary statistics, and we can also link directly to the DRS blob for each collection.

The link can be served from the nmdc-runtime dev deployment in the short term (https://api.nmdc-runtime-dev.polyneme.xyz/docs). We should get a nicer CNAME for this, e.g. api.microbiomedata.org.

wdduncan (approved):

wdduncan:
I'll go ahead an merge.

wdduncan:
Fixes #84

turbomam:
In what ways are we still using DRS? Is it useful to end users, esp. now that we have metadata-download APIs?

dwinston:
> Is it useful to end users...?
No idea. I have not been collecting usage logs. That would be step one.

turbomam:
Would adding logging take much effort?

dwinston:
shouldn't be too bad. can add a lightweight middleware to nmdc-runtime based on e.g. <https://github.com/tom-draper/api-analytics/tree/main/analytics/python/fastapi>.

PR #91, opened 2021-07-22
WIP: first pass at schema crosswalk

cmungall:


cmungall:
based on slide 12 https://docs.google.com/presentation/d/1WIdniAInrd71UGno3121Qw3X9Xx331wM0CiBFn36JQM/edit#slide=id.ge5e799d6b4_0_0

src/schema/nmdc.yaml
```diff
@@ -324,13 +335,30 @@ classes:
...
+      - GOLD study identifiers
+      - MGnify project identifiers
     slot_usage:
       doi:
         description: >-
           The dataset citation for this study
+    exact_mappings:
+      - SIO:001066  ## study
+      - NCIT:C63536 ## study
+      - OBI:0000066 ## investigation
```
wdduncan, line 350: `OBI:0000066` is a kind of planned process.  
`NCIT:C63536` is a kind of research activity.
`SIO:001066` is a kind of process.

Our description for study doesn't seem to capture the process sense of 'study' expressed in these concepts:
```
A study summarizes the overall goal of a research initiative and outlines the key objective of its underlying projects.
```

src/schema/nmdc.yaml
```diff
@@ -324,13 +335,30 @@ classes:
...
+      - ISA:Study
+    close_mappings:
+      - NCIT:C41198 ## investigation
```
wdduncan, line 353: Why is `NCIT:C41198 ## investigation` a `close_mapping` but `OBI:0000066 ## investigation` and `exact_mapping`?

Definition for `NCIT:C41198`:
```
 The act or process of a systematic and thorough examination; research, study; the process of inquiring into or following up, intended to develop facts.
```
Definition for `OBI:0000066`:
```
A planned process that consists of parts: planning, study design execution, documentation and which produce conclusion(s).
```

src/schema/nmdc.yaml
```diff
@@ -350,11 +378,16 @@ classes:
   omics processing:
     aliases:
       - omics assay
+      - sequencing project
+      - experiment
```
wdduncan, line 382: `experiment` seems a little too broad to be an alias.

src/schema/nmdc.yaml
```diff
@@ -350,11 +378,16 @@ classes:
...
     is_a: biosample processing
     in_subset: 
       - sample subset
     description: >-
       The methods and processes used to generate omics data from a biosample or organism.
+    alt_descriptions:
+      embl.ena: >-
+        An experiment contains information about a sequencing experiment including library and instrument details.
```
wdduncan, line 390: Do you want to say an experiment "produces" information? The "contains" information sounds odd to me.

dehays:
In src/schema/nmdc.yaml
<https://github.com/microbiomedata/nmdc-schema/pull/91#discussion_r675683607>
:

> @@ -350,11 +378,16 @@ classes:
   omics processing:
     aliases:
       - omics assay
+      - sequencing project
+      - experiment

I propose we take this opportunity (while making large schema changes)
to rename 'omics processing' to 'instrument process'.

Regarding the 'experiment' alias.  I believe this is meant to be an
SRA experiment defined as 'Each SRA EXPERIMENT (SRA accession *SRX#*)
is a unique sequencing result for a specific sample.' which is
therefore equivalent to the JGI / GOLD concept of a sequencing project

In SRA terminology an experiment produces 'Runs' where 'SRA RUN is
simply a manifest of data file(s) that are derived from sequencing a
library described by the associated EXPERIMENT.'  This is equivalent
to the output list of data objects (files) from our current
omics_processing.


-David

src/schema/nmdc.yaml
cmungall: Thanks! I'm not so concerned with differences in upper ontology classification, only with whether they capture the same concept or not.

src/schema/nmdc.yaml
cmungall: NCIT has both investigation and study, exact mappings should be one-to-one

src/schema/nmdc.yaml
cmungall: That's not my definition. See https://w3id.org/linkml/AltDescription - this is a description attributed to EMBL.ENA. See https://ena-docs.readthedocs.io/en/latest/submit/general-guide/metadata.html

src/schema/nmdc.yaml
cmungall: see @dehays comment. We should probably distinguish broad/narrow/exact synonyms as we do in OBO, but this is mostly here to index searching on the schema for now

cmungall:
@wdduncan or @turbomam - can you investigate why this is not passing? I don't think this is related to my changes

(can wait til after w/e of course)

I'd like to get some of these documentation-oriented changes in soon and then we can work on some larger changes, e.g. renaming #93.

src/schema/nmdc.yaml
wdduncan: Sorry, I wasn't meaning to sound like was going down the ontological rabbit hole. I was only pointing out that the things we are mapping too may not line up well with the sense of 'study' we are using in the NMDC.

src/schema/nmdc.yaml
wdduncan: Perhaps this just points to a weakness in the NCIT? It might be better to not map to NCIT:Investigation.

src/schema/nmdc.yaml
wdduncan: Hmmm ... so, should we allow broad synonyms as aliases? Perhaps it might be better to limit aliases to exact synonyms?

src/schema/nmdc.yaml
wdduncan: I wasn't aware of the ENA descriptions. What the value gained by adding the enum instead just the alternative description? If you are wanting to match fields across schemas, wouldn't it be better to use on of the mapping slots?

PR #92, opened 2021-07-23
adding range for was-informed-by

cmungall:


cmungall:
@wdduncan this introduces a cyclic dependency - not sure how to resolve other than merge wf activity back in to core schema?

src/schema/workflow_execution_activity.yaml
```diff
@@ -52,6 +52,7 @@ classes:
         required: true
       was informed by:
         required: true
+        range: omics processing
```
wdduncan, line 55: Are you wanting the range to be just omics processing or the more general biosample processing? Also should we make biosample process and kind of prov:Activity?

wdduncan: @cmungall The PR fails b/c `nmdc` is not imported, and `omics processing` is a class in `nmdc`. Here are the imports for `workflow_execution_activity`:
```
imports:
  - prov
  - core
  - external_identifiers
  ```
However, `nmdc` also imports `workflow_execution_activity`. Here are its imports:
```
imports:
  - linkml:types
  - mixs
  - core
  - prov
  - workflow_execution_activity
  - annotation
  - external_identifiers
```

So, if `workflow_execution_activity` imports `nmdc` and `nmdc` imports `workflow_execution_activity` won't this create a cyclic dependency?

**Comment by sujaypatil96** [comment]
@cmungall @wdduncan two questions to answer here based on the above discussion:
* range of `was informed by` should be `omics processing` or `biosample processing`? understanding: `activity` which is the range of `was informed by` in the prov schema, can be informed by either an omics sequencing analysis project, or by the sample on which the analysis was run in the workflows world
* how to solve the problem of cyclic imports? move biosample processing, omics processing to a workflow specific schema like workflow_execution_activity.yaml, or perhaps a different wf sub schema altogether?

**Comment by turbomam** [comment]
src/schema/prov.yaml currently has the following under slots.was_informed_by:

```yaml
  was_informed_by:
    range: Activity
    mappings:
      - prov:wasInformedBy
```

I am going to leave things like that, open a new issue regarding making the range narrower, and then close this issue.

PR #98, opened 2021-07-30
Issue 56: generate jsonld context and use pipenv to build targets

### LINKED ISSUE #56: generate JSONLD context
Generate JSONLD context as one of the output artifacts.

cc @cmungall @turbomam

wdduncan:
Fixes #56 

Add target to the `Makefile` to generate a jsonld context.  
As part of this, I'm switching to using `pipenv`.


wdduncan:
@cmungall I added the `jsonld-context` target.  I also changed the way targets are generated to use `pipenv`. My reason for doing this was to be more in line with Harold.

PR #117, opened 2021-08-20
Tagging type field as type designator field

cmungall:


src/schema/basic_slots.yaml
```diff
@@ -57,6 +57,7 @@ slots:
     description: >-
       An optional string that species the type object.
```
wdduncan, line 58: There is a typo "species" -> "specifies".  
Not sure how I missed that up until now :(

src/schema/basic_slots.yaml
```diff
@@ -57,6 +57,7 @@ slots:
     description: >-
       An optional string that species the type object. 
       This is used to allow for searches for different kinds of objects.
+    type_designator: true
```
wdduncan, line 60: I was unaware of the `type_designatory` slot :)

wdduncan (approved):

src/schema/basic_slots.yaml
wdduncan: Can you update the [CHANGELOG](https://github.com/microbiomedata/nmdc-schema/blob/main/CHANGELOG.md)?

Issue #126, opened 2021-08-30
src/schema/mixs.yaml has no name slot


turbomam:
`linkml-runtime/linkml_runtime/utils/schemaview.py` from `linkml/linkml-runtime`
 will instantiate off of `nmdc.yaml`, but operations like `all_class` fail under the default `imports=True` condition, because `src/schema/mixs.yaml` doesn't have a required(?) `name` value

How is `src/schema/mixs.yaml` built? If it was a one time manual thing, can I just add a name?

PR #128, opened 2021-08-31
Issue 126: add mixs name

wdduncan:
See #126 

Added name and title to `mixs.yaml` and regenerated the artifacts (`make all`).


turbomam (approved):
Thanks.

I didn't look at nmdc.schema.json or nmdc.py

Issue #120, opened 2021-08-24
add name slot to person value


wdduncan:
Person values should have a `name` slot (also, perhaps,`first name`, `middle name`, and `last name` slots).

cc @cmungall @dehays

turbomam:
added `name` slot but omitted finer grained modeling.

turbomam:
In discussion with @cmungall about deprecating the currently-used `has raw value`

I've left it like this for now:

```
  person value:
    is_a: attribute value
    description: >-
      An attribute value representing a person
    slots:
      - orcid
      - profile image url
      - email
      - name
    todos:
      - add additional fields e.g for institution
    slot_usage:
      has raw value:
        description: >-
          the full name of the Investgator in format FIRST LAST
      name:
        description: >-
          will replace has raw value?```

PR #146, opened 2021-09-20
#120 `name` for `person value`

turbomam:
- included note about potential future deprecation of `has raw value` in `person value`
- updated `CHANGELOG.md`

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-linkml = ">=1.0.4"
+linkml = ">=1.0.3"
 jq = "*"
 mkdocs = "*"
 twine = "*"
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-linkml = ">=1.0.3"
+linkml = ">=1.0.4"
 jq = "*"
 mkdocs = "*"
 twine = "*"
--- a/Pipfile
+++ b/Pipfile
@@ -9,6 +9,8 @@ jq = "*"
 mkdocs = "*"
 twine = "*"
 build = "*"
+# not sure why we would need openpyxl, but 3.0.8 caused unresolvable dependcies in main
+openpyxl = "3.0.7"
 
 [dev-packages]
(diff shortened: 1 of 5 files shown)


Pipfile
```diff
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-linkml = ">=1.0.4"
+linkml = ">=1.0.3"
```
wdduncan, line 7: I forget what was added in linkml version `1.0.4`. But the higher version might be needed.

wdduncan (approved):
I think this is fine. But, please look at my comments. In particular, the description for `name` needs to be a description, not a question.

src/schema/core.yaml
```diff
@@ -137,13 +137,14 @@ classes:
       - name
     todos:
       - add additional fields e.g for institution
+      - deprecate "has raw value" in favor of "name"
     slot_usage:
       has raw value:
         description: >-
-          the full name of the Investgator in format FIRST LAST
+          The full name of the Investgator in format FIRST LAST. May eventually be deprecated in favor of "name".
```
wdduncan, line 144: Information about depreciation is something that would go in a note. I would not include it in a description. Also, you may want to in information about an optional middle name or initial. E.g.:
```
description: >-
  The name of the person.   
  It should follow the format FIRST [MIDDLE NAME| MIDDLE INITIAL] LAST, where MIDDLE NAME| MIDDLE INITIAL is optional.

File: src/schema/core.yaml [hunk 3]
```diff
@@ -141,10 +141,14 @@ classes:
...
         description: >-
-          The full name of the Investgator in format FIRST LAST. May eventually be deprecated in favor of "name".
+          The full name of the Investgator in format FIRST LAST.
+        notes: >-
+          May eventually be deprecated in favor of "name".
       name:
         description: >-
-          The full name of the Investgator in format FIRST LAST.
+          The full name of the Investgator.
+          It should follow the format FIRST [MIDDLE NAME| MIDDLE INITIAL] LAST, where MIDDLE NAME| MIDDLE INITIAL is optional.
```
wdduncan, line 150: Thanks. Looks good now :)

Issue #100, opened 2021-08-03
add email slot for PersonValue


turbomam:
Will take a string, like the orcid slot does. But maybe 0..inf cardinality?

turbomam:
"Due diligence"

```Bash
% grep -i mail src/schema/*
```

> mixs.yaml:    range: text value  ## syntax: "[attic|bathroom|closet|conference room|elevator|examining room|hallway|kitchen|mail room|private office|open office|stairwell|,restroom|lobby|vestibule|mechanical or electrical room|data center|laboratory_wet|laboratory_dry|gymnasium|natatorium|auditorium|lockers|cafe|warehouse] "
mixs.yaml:    pattern: "[attic|bathroom|closet|conference room|elevator|examining room|hallway|kitchen|mail room|private office|open office|stairwell|,restroom|lobby|vestibule|mechanical or electrical room|data center|laboratory_wet|laboratory_dry|gymnasium|natatorium|auditorium|lockers|cafe|warehouse]"
mixs.yaml:    range: text value  ## syntax: [attic|bathroom|closet|conference room|elevator|examining room|hallway|kitchen|mail room|office|stairwell]
mixs.yaml:    pattern: "[attic|bathroom|closet|conference room|elevator|examining room|hallway|kitchen|mail room|office|stairwell]"

turbomam:
the `orcid` slot is define in `core.yaml`

turbomam:
I'd like to give this a domain, but I think entities other than people can have an email address. Should I assert a domain of 'named thing', or is that implicit? Is there something with the right specificity between 'named thing' and 'person value'?

turbomam:
Should I assert a format/pattern or get anything started to enable validation? I know that's a slippery slope when it comes to email addresses. 'orcid' is defined with a string range and no pattern.

turbomam:
the following email slot has been merged into `core.yaml`

```  email:
    description: >-
      An email address for an entitity such as a person
    range: string
    notes: >-
      More specific domain? Acceptable pattern? https://emailregex.com/ ?
```

possible followups:

-  'entity' misspelled as 'entitity'
- remove note
- improve range with a regular expression pattern: https://emailregex.com/ ?
- add domain?

turbomam:
reopened branch on remote

turbomam:
Add `foaf:mbox` IRI? Would need a prefix definition?

turbomam:
@cmungall suggests using https://schema.org/email.

[...]

PR #147, opened 2021-09-20
addition of email slot

turbomam:
local make all OK now


wdduncan (approved):
This looks fine to me. Where you going to add the email slot to the person value class?
See #100

PR #156, opened 2021-09-22
Add ess_dive_datasets to study

### LINKED ISSUE #140: Study slot for ESS-DIVE dataset list
Looking at https://github.com/microbiomedata/nmdc-schema/blob/main/src/schema/external_identifiers.yaml

My plan would be to add a 'ess dive dataset identifiers' type to https://github.com/microbiomedata/nmdc-schema/blob/main/src/schema/external_identifiers.yaml and then use that as the range for a 'ess dive datasets' slot on study

@cmungall @wdduncan 

We have values to populate this with now.  They are all DOIs.

dehays:
Closes #140

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -357,10 +357,6 @@ classes:
       doi:
         description: >-
           The dataset citation for this study
-      ess dive datasets:
-        multivalued: true
-        description: >-
-          List of ESS-DIVE dataset DOIs
     exact_mappings:
       - SIO:001066  ## study
       - NCIT:C63536 ## study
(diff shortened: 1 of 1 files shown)


src/schema/nmdc.yaml
```diff
@@ -356,6 +357,10 @@ classes:
       doi:
         description: >-
           The dataset citation for this study
+      ess dive datasets:
```
cmungall, line 360: this should go under the top level "slots"

dehays: moved this from the slot usage section on study to the top level slots section

PR #166, opened 2021-09-30
Issue 163: biosample "part of" slot_usage produces schema validation

### LINKED ISSUE #163: biosample "part of" slot_usage produces schema validation
In `nmdc.yaml`, the `biosample` class defines `part of` in the `slot_usage` section:
```
slot_usage:
  ...
  part of:
    required: false
    range: study
    description: >-
      Relates the biosample to the study for which the sample was collected.
```
During ETL, this results in a "part of" (note the space) in json output. E.g.:
```
{
  "id": "gold:Gb0115217",
  ...
  "part of": [
      "gold:Gs0114663"
    ]
}
```
This is a schema violation. It should be "part_of" (with an underscore).

I think the solution is to include `part of` in `slots` section of the `biosample` class.

cc @cmungall @dwinston

wdduncan:
The problem has nothing to do with linkml. I was on the translation side :(

Fixes #163

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -300,7 +300,7 @@ classes:
         required: true
       env_medium:
         required: true
-      id_prefixes:
+    id_prefixes:
       - GOLD
     exact_mappings:
       - OBI:0000747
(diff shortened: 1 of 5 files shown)


src/schema/nmdc.yaml
```diff
@@ -299,23 +300,12 @@ classes:
         required: true
       env_medium:
         required: true
-    id_prefixes:
+      id_prefixes:
```
cmungall, line 303: you introduced a syntax error here

Issue #178, opened 2021-10-12
expand regex for functional annotations


wdduncan:
Not all functional annotations are conforming to our [regexs](https://github.com/microbiomedata/nmdc-schema/blob/582883ba745642676158e383635f2609c72ddcfd/src/schema/annotation.yaml#L198). As @dwinston pointed out, the pattern `3-hydroxyacyl-[acyl-carrier-protein] dehydratase` doesn't conform (it doesn't have a prefix). He is currently [filtering out](https://github.com/microbiomedata/nmdc-runtime/blob/7abe1cb2fbe430762480bb57f0b7be15823944b8/nmdc_runtime/site/ops.py#L426) whatever doesn't match the schema.

@cmungall I thought we wanted all annotations to have prefixes ... but I may be misremembering.

cc @dehays @turbomam

cmungall:
Here is what is happening here. Thinking in terms of regexes is not so useful here. What we have is a distinction between two kinds of annotations

 1. Annotations of genes with a gene product name, e.g. 3-hydroxyacyl-[acyl-carrier-protein] dehydratase
 2. Annotations of genes with a term for a vocabulary/ontology/database, e.g. KEGG.KO, Pfam, EC, MetaCyc, GO, etc

With type 1 annotations, there is no ID (and hence no ID regex), with type 2 there is an ID (and hence regex makes sense). But the key issue here is accommodating a system that allows both kinds of annotations

you'll see we have a todo anticipating exactly this scenario:

```yaml
  functional annotation term:
    aliases:
      - function
      - functional annotation
    is_a: ontology class
    description: >-
      Abstract grouping class for any term/descriptor that can be applied to a functional unit of a genome (protein, ncRNA, complex).
    abstract: true
    todos:
      - decide if this should be used for product naming
```

It looks like in 7301f4d6 @turbomam changed the range of `has_function` in FunctionalAnnotation from FunctionalAnnotationTerm to string, presumably in order to accomodate non-ID annotations.

## SHORT TERM SOLUTION (this release?)

Remove the regex from `has_function`. Keep it modeled as string (or slightly better: create a linkml type mapped to a strong called FunctionalAnnotationType), and document this as being allowed to be either a descriptive string or an ID from an allowed system

__it is counterproductive to try and make a regex for this__ we can't anticipate all the characters that might be in a product name. We might hope they conform to a subset of ascii but there are no guarantees, especially when chemical nomenclature is embedded.

[...]

PR #179, opened 2021-10-13
removed functional annotation regex pattern for #178

turbomam:
there's a functional_annotation_set_invalid.json test, but I don't that that was being run even before this change


wdduncan:
Fixes #178 
Fixes #179

PR #199, opened 2021-11-08
Issue 182: Add file type enums

### LINKED ISSUE #182: add file type enums
New file type enums are needed. These can be retrieved from the file_type_enum collection. Each object has an id, name, description. So, write script to generate the enums.

After completed, request review of names and descriptions.

Current enum list of 30 types available [here](https://gist.github.com/dwinston/bdd8aaa0461e15f67da50604a9060bd9).

cc @dehays @cmungall

wdduncan:
Fixes #182 

@scanon @hubin-keio I've added the `file enums` based on the data provided by @dwinston [here](https://gist.github.com/dwinston/bdd8aaa0461e15f67da50604a9060bd9).
Can you please review the enums for correctness. Also, let me know if there are more I need to add.

Note: many of these types were already present in the old enums. E.g.: 
old enum:
```
assembly.agp:
        description: An AGP format file describes the assembly.
```
new enum:
```
Assembly AGP:
        description: An AGP format file that describes the assembly
```
To avoid confusion, I've deleted the old file enums.

cc @dehays @cmungall

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Changes to the schema are documented in this file.
 ### Added
   - N/A
 ### Fixed
-  - N/A
+  - updated file enums for workflows (#182)
 ### Changed 
   - N/A
 ### Removed
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -25,9 +25,9 @@
         },
         "antlr4-python3-runtime": {
             "hashes": [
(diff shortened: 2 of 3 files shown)


hubin-keio (approved):
Approved changes. Very likely there will be another change to expand the enum file types for metaT once we have synchronize the metaT workflow with JGI.

Issue #233, opened 2022-01-10
'examples' from GOLD sample identifiers: https://identifiers.org/gold:GbTODO


turbomam:

turbomam:
Could we either enter a real example GOLD sample identifier, or remove the example and add a to-do?

turbomam:
in `src/schema/external_identifiers.yaml`

```yaml
  GOLD sample identifiers:
    is_a: sample identifiers
    mixins:
      - GOLD identifiers
    pattern: "^GOLD:Gb[0-9]+$"
    multivalued: true
    description:
      - identifiers for corresponding sample in GOLD
    examples:
      - value: https://identifiers.org/gold:GbTODO
```

cmungall:
can you go ahead make the PR @turbomam - this kind of thing shouldn't require a lot of discussion. Just use any valid gold id, e.g. gold:Gb0108335

PR #238, opened 2022-01-11
verified GOLD ID examples  for #233

turbomam:
move some examples to todos for #233


cmungall (changes requested):
can you not just make a real example?

turbomam:
> can you not just make a real example?

I don't know what the ids/URIs are supposed to look like for these identifiers. Where could I find some?

cmungall:
just look on the data portal or on gold itself!

wdduncan:
@turbomam Please update the changelog.

Issue #231, opened 2022-01-10
GOLD Path term desciption still "TODO"


turbomam:
- ecosystem
- ecosystem_category
- ecosystem_subtype
- ecosystem_type
- specific_ecosystem

turbomam:
in `src/schema/nmdc.yaml`

cmungall:
You can also include see_also links to

 - https://gold.jgi.doe.gov/ecosystem_classification
 - https://pubmed.ncbi.nlm.nih.gov/20653767/

turbomam:
https://microbiomedata.github.io/nmdc-schema/ecosystem/

PR #239, opened 2022-01-11
overwrote some TODO descriptions for #231

turbomam:
overwrote some TODO descriptions for #231

definitions provided by https://gold.jgi.doe.gov/help via @emileyfadrosh

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -796,27 +796,27 @@ slots:
   ecosystem:
     is_a: gold_path_field
     description: >-
-      TODO
+      An ecosystem is a combination of a physical environment (abiotic factors) and all the organisms (biotic factors) that interact with this environment. The abiotic factors play a profound role on the type and composition of organisms in a given environment. The GOLD Ecosystem at the top of the five-level classification system is aimed at capturing the broader environment from which an organism or environmental sample is collected. The three broad groups under Ecosystem are Environmental, Host-associated, and Engineered. They represent samples collected from a natural environment or from another organism or from engineered environments like bioreactors respectively. See GOLD terminology at https://gold.jgi.doe.gov/help.
 
   ecosystem_category:
     is_a: gold_path_field
     description: >-
-      TODO
+      Ecosystem categories represent divisions within the ecosystem based on specific characteristics of the environment from where an organism or sample is isolated. For example, the Environmental ecosystem is divided into Air, Aquatic and Terrestrial. Ecosystem categories for Host-associated samples can be individual hosts or phyla and for engineered samples it may be manipulated environments like bioreactors, solid waste etc. See GOLD terminology at https://gold.jgi.doe.gov/help.
 
   ecosystem_type:
     is_a: gold_path_field
     description: >-
-      TODO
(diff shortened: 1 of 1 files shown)


src/schema/nmdc.yaml
```diff
@@ -796,27 +796,27 @@ slots:
   ecosystem:
     is_a: gold_path_field
     description: >-
-      TODO
+      An ecosystem is a combination of a physical environment (abiotic factors) and all the organisms (biotic factors) that interact with this environment. The abiotic factors play a profound role on the type and composition of organisms in a given environment. The GOLD Ecosystem at the top of the five-level classification system is aimed at capturing the broader environment from which an organism or environmental sample is collected. The three broad groups under Ecosystem are Environmental, Host-associated, and Engineered. They represent samples collected from a natural environment or from another organism or from engineered environments like bioreactors respectively. See GOLD terminology at https://gold.jgi.doe.gov/help.
```
wdduncan, line 799: I think this and the other GOLD path descriptions can be shortened. For example, the `ecosystem` description can be shorted to:
```
An ecosystem is a combination of a physical environment (abiotic factors) and all the organisms (biotic factors) that interact with this environment. 
 ```
Having the shortened description will help if we want the description to show in tooltips.  
The rest of the information is good. But, I would put the following in a [comment](https://linkml.io/linkml-model/docs/comments/#slot-comments)
That is, I would put the following in a comment:
```
The abiotic factors play a profound role on the type and composition of organisms in a given environment. The GOLD Ecosystem at the top of the five-level classification system is aimed at capturing the broader environment from which an organism or environmental sample is collected. The three broad groups under Ecosystem are Environmental, Host-associated, and Engineered. They represent samples collected from a natural environment or from another organism or from engineered environments like bioreactors respectively. See GOLD terminology at https://gold.jgi.doe.gov/help.
```
@cmungall may have different opinion though.

**Review by wdduncan — CHANGES_REQUESTED** [review]

**Comment by wdduncan** [comment]
@turbomam Please update the changelog.

File: src/schema/nmdc.yaml (hunk 1 above)
Comment (turbomam) [reply src/schema/nmdc.yaml]: Good point

For some of the Gold Ecosystem Classification Levels, I don't think the first sentence alone provides enough description. But I have moves the examples into `comments`

For example I don't find 
> Ecosystem subtypes represent further subdivision of Ecosystem types into more distinct subtypes
informative at all.

I have added a statement like
> Ecosystem type is in position 3/5 in a GOLD path.
to each of the descriptions.

**Comment by cmungall** [comment]
Ideally we would have an enum for each path component, but for now just
listing a few key examples would go a long way to elucidating these

File: src/schema/nmdc.yaml [hunk 2]
```diff
@@ -796,27 +796,37 @@ slots:
...
+    comments: The abiotic factors play a profound role on the type and composition of organisms in a given environment. The GOLD Ecosystem at the top of the five-level classification system is aimed at capturing the broader environment from which an organism or environmental sample is collected. The three broad groups under Ecosystem are Environmental, Host-associated, and Engineered. They represent samples collected from a natural environment or from another organism or from engineered environments like bioreactors respectively.
+    see_also: https://gold.jgi.doe.gov/help
 
   ecosystem_category:
     is_a: gold_path_field
     description: >-
-      TODO
+      Ecosystem categories represent divisions within the ecosystem based on specific characteristics of the environment from where an organism or sample is isolated. Ecosystem category is in position 2/5 in a GOLD path.
+    comments: The Environmental ecosystem (for example) is divided into Air, Aquatic and Terrestrial. Ecosystem categories for Host-associated samples can be individual hosts or phyla and for engineered samples it may be manipulated environments like bioreactors, solid waste etc.
```
wdduncan, line 807: Interestingly, the arity for [comments](https://linkml.io/linkml-model/docs/comments/#slot-comments) is `0..*`. @cmungall Doesn't this mean the `comments` slot should take a list of values?  
It does pass validation though.

cmungall: it should. The schemaloader is liberal in what it accepts, and will convert single value to list for list types. Nevertheless we should conform.

PR #242, opened 2022-01-19
Issue 216: add slot for multiple credit roles

### LINKED ISSUE #216: permit for multiple credit roles
@dehays @emileyfadrosh would like to be to add multiple credit roles for a researcher.  Currently, the `applied role` is not multivalued. Making it multivalued  would invalidate the data presently in Mongo and would cause issues with the data portal.

I propose the following:
* Keep the `applied role` slot but deprecate it.
* Create an `applied roles` slot to capture the multiple roles an investigator may have.
* After the data portal is up to date, retire the `applied role` slot.

cc @cmungall @turbomam

wdduncan:
Fixes #216 

- deprecated `applied role` slot
- added `applied roles` (multivalued) slot

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -566,7 +566,9 @@ enums:
       Principal Investigator:
         description: principal investigator role
         meaning: OBI:0000103
-
+    comments:
+      - credit enums come from https://casrai.org/credit/
+    
 slots:
 
   ess dive datasets:
@@ -605,8 +607,6 @@ slots:
     domain: credit association
     range: credit enum
     slot_uri: prov:hadRole
-    comments:
-      - credit enums come from https://casrai.org/credit/
     deprecated: A credit association may have multiple roles. So, use the applied roles slot.
     multivalued: false
     required: false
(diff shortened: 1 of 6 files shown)


src/schema/nmdc.yaml
```diff
@@ -439,6 +439,7 @@ classes:
     slots:
       - applies to person
       - applied role
+      - applied roles
```
cmungall, line 442: so there will be two fields, `applied role` and `applied roles`? Seems a little unusual. However, go ahead and merge if you think this will work we can always revise on a future iteration

cmungall: also you have the plural form set to required: true, so there will always be a list of length 1, as well as the single valued?

wdduncan: I set the required to `false` for `applied role` (singular). I only left the `applied role` slot there b/c I didn't want to invalidate the data currently in the database that uses this slot. After the databases is updated using the `applied roles` slot, I will remove it from the `credit association`.

src/schema/nmdc.yaml
```diff
@@ -604,9 +605,18 @@ slots:
     domain: credit association
     range: credit enum
     slot_uri: prov:hadRole
-    todos: >-
-      role enums come from https://casrai.org/credit/
+    comments:
+      - credit enums come from https://casrai.org/credit/
```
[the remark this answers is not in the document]
wdduncan: Good point. I'll go ahead and move the comments.

Issue #256, opened 2022-02-04
report of all human-readable annotations on all NMDC schema terms


turbomam:
for @emileyfadrosh

turbomam:
@emileyfadrosh is it OK if I close this issue?

emileyfadrosh:
Yes, thank you!

PR #257, opened 2022-02-04
#256 slot report

turbomam:
gets saved to the local dir. could put in an assets dir instead. have published one time at https://docs.google.com/spreadsheets/d/1yTCqZojqoLPa0Y97ZYcyFuEG-u-oXf7OSCvq9Hu4dCQ/edit#gid=0

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -5,6 +5,8 @@ name = "pypi"
 
 [packages]
 linkml-runtime = ">=1.1.6"
+click_log = "*"
+jsonasobj2 = "*"
 
 [dev-packages]
 linkml = ">=1.1.13"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "2942c9a915fc3e5cec69a32e49fb3ffa5976bfca1ee9797bfd9fd7571b29750a"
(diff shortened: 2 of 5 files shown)


turbomam:
could also do a long format, and include a 'preferred replacement value' column. For example, I think there is an interest in using slot usage to assign a description value of 'a human-readable description of a thing'  to the description  slot on the study class.

wdduncan:
@turbomam I'm not sure what you mean by using `slot_usage`. You can override the descriptions for slots (of course), but I do not think you can override descriptions for classes (I could do this locally). Perhaps you make use of the [alt_descriptions](https://linkml.io/linkml-model/docs/alt_descriptions/) slot?

Also, can you open up access to the google sheet to anyone with a link?

turbomam:
> I do not think you can override descriptions for classes (I could do this locally). Perhaps you make use of the [alt_descriptions](https://linkml.io/linkml-model/docs/alt_descriptions/) slot?

yes, good points @wdduncan

turbomam:
> can you open up access to the google sheet to anyone with a link?

I opened it up for viewing to `anyone with a link`. Did you want it world editable @wdduncan ?

wdduncan:
@turbomam I was thinking it would be helpful if anyone with a link could view the google sheet.

PR #263, opened 2022-02-10
integrate poetry into nmdc-schema repo

sujaypatil96:
Create initial pyproject.toml and poetry.lock files that intend to replace analogous files created for setuptools and pipenv.

Tasks to be completed before we fully eliminate setuptools, pipenv and tox:

- [ ] Modify Makefile with poetry commands
- [ ] Modify Github Actions with poetry commands
- [ ] Automatic release management

But the above tasks are for separate issues and PRs, so this PR is good for now, I think.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,0 +1,2490 @@
+[[package]]
+name = "alabaster"
+version = "0.7.12"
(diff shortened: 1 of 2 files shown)


wdduncan:
@sujaypatil96 great! I'll check this out in a bit.

Issue #252, opened 2022-02-02
short term: programmatically tag the origin of MIxS terms when imported into NMDC schema


turbomam:
CC @cmungall @wdduncan 

[Sample MIxS term in NMDC shema](https://github.com/microbiomedata/nmdc-schema/blob/b7e656664e39ec6d528c8a0ebd66636842bc6560/src/schema/mixs.yaml#L147-L159)

see also https://github.com/microbiomedata/nmdc-server/issues/555

especially the discussion of MIxS' lat_lon field

turbomam:
CC @cmungall @wdduncan 

see also ...

turbomam:
Implement with `conforms_to` or `annotations`?

semantics too loose?
- in_subset
- see_also

probably not
- quick and dirty parse of slot_uri, from_schema, imported_from, exact_mappings etc.
- owner (deprecated)

wdduncan:
tagging @mslarae13

cmungall:
I think in_subset is good

turbomam:
I implemented with with `source: http://w3id.org/mixs/terms` in [`mixs.yaml`](https://github.com/microbiomedata/nmdc-schema/blob/main/src/schema/mixs.yaml)

When creating the Bioscales tempalte, we found that the LinkML `source` attribute conflicts with a reserved `source` annotation in DataHarmonizer, so this might not be a good long term solution.

turbomam:
[What are DataHarmonizer's reserved field names](https://github.com/cidgoh/DataHarmonizer/issues/313)

turbomam:
> Just checking, does this relate to the discussion we had yesterday about "annotating" the NMDC schema slots with "where they came from"?

@mslarae13 @ssarrafan 

Yes, but I'm not convinced I did it in a sustainable way, or that I used the most appropriate LinkML slot to make the annotations. Will look into it more in the beginning of August.

ssarrafan:
> > Just checking, does this relate to the discussion we had yesterday about "annotating" the NMDC schema slots with "where they came from"?
> 
> @mslarae13 @ssarrafan
> 
> Yes, but I'm not convinced I did it in a sustainable way, or that I used the most appropriate LinkML slot to make the annotations. Will look into it more in the beginning of August.

I'm tempted to tag this as interim fix and close it but will move it to August for you @turbomam. FYI @mslarae13

mslarae13:
[...]

PR #269, opened 2022-02-22
Issue 252 mixs subset

turbomam:
preparations for #252


turbomam:
MIxS terms imported into nmdc-schema are all in `src/schema/mixs.yaml` now. They're tagged with 

>     source: http://w3id.org/mixs/terms

although I'm not convince that `source` is the best predicate

PR #277, opened 2022-03-01
replace pipenv commands with poetry in Makefile

sujaypatil96:
* Refactor Makefile to replace pipenv commands with poetry commands
* Freeze out and add a newly updated poetry.lock file


sujaypatil96:
Note: I updated only the Github Action that gets triggered on the creation of a PR since only that needed to be updated for the PR to go through.

I will update the other Actions in an upcoming PR.

PR #282, opened 2022-03-14
Add `gen-linkml` target to Makefile

sujaypatil96:
Modeling based on other generator targets in the Makefile, add a `gen-json` target to the Makefile that uses `gen-linkml` from the linkml library to generate a JSON artifact that has the LinkML model with slots for all classes materialized as attributes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,8 +43,8 @@ nmdc-data = "nmdc_schema.nmdc_data:cli"
 [tool.poetry.dependencies]
 python = "^3.9"
 
-linkml = ">=1.2.0rc6"
-linkml-runtime = ">=1.2.0rc6"
+linkml = "*"
+linkml-runtime = "*"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -41,10 +41,10 @@ nmdc-version = "nmdc_schema.nmdc_version:cli"
 nmdc-data = "nmdc_schema.nmdc_data:cli"
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = "^3.9"
```
sujaypatil96, line 44: Since the Actions infrastructure runs the checks on Python 3.9, I thought it made sense to enforce the package requirements to have 3.9 and above as well.

cmungall (approved):

pyproject.toml
cmungall: I might have made a separate PR as this is a separate concern but OK to merge in this one

PR #285, opened 2022-03-16
First pass at export of nmdc schema to TSV for DOE sample interop project

cmungall:
First pass at export for DOE sample interoperability project with @elishawc 

Aim is for multiple projects to export their entire data dictionaries to a simple TSV format

Can be browsed here:
https://github.com/microbiomedata/nmdc-schema/blob/doe-sample-interop/sheets/nmdc.sheets.tsv

Note: lots missing will look what's up later

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -230,3 +230,7 @@ validate-%: test/data/%.json jsonschema/nmdc.schema.json
 
 validate-invalid-%: test/data/invalid_schemas/%.json jsonschema/nmdc.schema.json
 	! $(RUN) jsonschema -i $< $(word 2, $^)
+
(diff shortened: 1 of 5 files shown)


mslarae13:
@turbomam 

Could you get me an updated file for the schema and v6 changes?

turbomam:
> Could you get me an updated file for the schema and v6 changes?

OK, when there are changes to the nmdc-schema, I will rerun the step in the Makefile that generates `sheets/nmdc.sheets.tsv`

Will also try to remember some tools for finding close matches in term labels

turbomam:
@cmungall @mslarae13 @elishawc should this PR and the corresponding branch still be left open?

Issue #288, opened 2022-03-24
Display hints requested for "NCBI BioProject Title" and "Data Types", but those aren't accessible from the study class


turbomam:
See #267 and [SubmissionPortal_TermUpdates](https://docs.google.com/spreadsheets/d/1T31B5gbRgS6vuK_lIg0ZLpIyafByWhlvE2hfcCy-c4k/edit#gid=0&range=15:17) 

I have [added display hint annotations to the NMDC schema](https://github.com/microbiomedata/nmdc-schema/pull/287) for most of the rows in the `SubmissionPortal_TermUpdates` sheet. In those cases, I have provided jq paths to the hints.

However, I don't see anything [related to studies in the schema](https://github.com/microbiomedata/nmdc-schema/blob/issue-267-disp-hints/assets/study_indiced_slots.yaml) that correlates strongly with "bioproject title" or "data type"

I guess you could argue that [alternative titles](https://microbiomedata.github.io/nmdc-schema/alternative_titles/#slot-alternative-titles) could be used for "bioproject title", but the semantics are pretty weak.

Also, why would somebody enter a title when it could be looked up from the accession?

For "Data Types", is that supposed to correlate with [analysis_type](https://docs.google.com/spreadsheets/d/1WkftrJV548wO5Oh1L-K6N1BNU03btRUm2D7jvlHc7Uc/edit#gid=0&range=70:70) from our DataHarmonizer interface? I don't think that term is defined in either the NMDC or MIxS schemas yet.

turbomam:
rather hinting about how to enter something, not hinting on how to display it

tooltips

turbomam:
could column A or B be an actual DOM path? or anything verifiable?

turbomam:
why are different tooltips for PI and Contrib ORCiD

box around PI fields?

then could have tooltip on principal investigator slot

ssarrafan:
I'm going to move this one to April as it appears to be in process.  I will remove @mslarae13 as an assignee.

turbomam:
@dehays provided some useful feedback about other classes that could receive this information, esp. 'Data Types'. Omics processing?

- [ ] I don't think we should add a display hint for 'NCBI BioProject Title' or even include that on the study submission pages at all. It should be looked up in NCBI based on the BioProject ID/accession (which should also be verified

[...]

PR #290, opened 2022-03-29
Issue 288 non subclass hints

turbomam:
Issue #288 non subclass hints


turbomam (commented):
- `python/nmdc.py` will have to be regenerated before release
- there are still some placeholder display hints like this in `assets/study_indiced_slots.yaml` ?!
- rename `assets/study_indiced_slots.yaml`
- need better documentation on these new assets
- add annotation, to pair with display hints, that says where in the NMDC Portal's DOM the hint goes

```yaml
- name: has credit associations
  annotations:
    display_hint:
      tag: display_hint
      value: has credit associations slot display hint
```

Issue #291, opened 2022-03-29
mixs as a submodule


turbomam:
for moving away from the old static-ish MIxS 5 [mixs.yaml](https://github.com/microbiomedata/nmdc-schema/blob/main/src/schema/mixs.yaml) 

MIxS isn't available through w3id yet, and the LinkML YAML files aren't bundled with the PyPI package yet:

```shell
pip show -f mixs-linkml
```

> Name: mixs-linkml
Version: 0.1.1
Summary: A LinkML (https://linkml.io/) model of the MIxS standard (https://gensc.org/mixs/)
Home-page: 
Author: GSC
Author-email: 
License: CC0
Location: /Users/MAM/mixs_pypi_test/venv/lib/python3.9/site-packages
Requires: linkml, mkdocs, pandas
Required-by: 
Files:
  mixs_linkml-0.1.1.dist-info/INSTALLER
  mixs_linkml-0.1.1.dist-info/LICENSE
  mixs_linkml-0.1.1.dist-info/METADATA
  mixs_linkml-0.1.1.dist-info/RECORD
  mixs_linkml-0.1.1.dist-info/REQUESTED
  mixs_linkml-0.1.1.dist-info/WHEEL
  release/__init__.py
  release/__pycache__/__init__.cpython-39.pyc
  release/graphql/mixs.graphql
  release/jsonld/mixs.context.jsonld
  release/jsonld/mixs.jsonld
  release/jsonschema/mixs.schema.json
  release/mixs.py
  release/owl/mixs.owl.ttl
  release/prefixmap/mixs.yaml
  release/protobuf/mixs.proto
  release/shacl/mixs.shacl.ttl
  release/shex/mixs.shex
  release/sqlschema/mixs.sql

turbomam:
- _16s_recover
- _16s_recover_software
- bin name
- bin quality
- chem_treatment_method
- completeness
- contamination
- env_package
- extreme_salinity
- gene count
- gtdbtk class
- gtdbtk domain
- gtdbtk family
- gtdbtk genus
- gtdbtk order
- gtdbtk phylum
- gtdbtk species
- health_disease_stat
- heat_system_deliv_meth
- host_blood_press_diast
- host_blood_press_syst
- ihmc_ethnicity
- investigation_type
- mixs_url
- nose_mouth_teeth_throat_disord
- num 16s
- num 23s
- num 5s
- num tRNA
- number of contig
- organism_count_qpcr_info
- pres_animal
[...]

PR #292, opened 2022-03-30
Integrate MIxS 6 submodule with nmdc-schema, make new import

turbomam:
## Goal
Update the MIxS integration in NMDC soon after Apr 28th release

@turbomam: minimize changed files. Give PR and issue a better name.

Not all of the reviewers are expected to pour over all of the changes.

---

See Issue #291 

- identify slots historically provided by `src/schema/mixs.yaml`
- also call out lost slots... are they really needed/used?
- import as many of those slots from MIxS 6, into a new YAML file


turbomam:
This PR is mostly concerned with creating a new, dynamic MIxS import for the nmdc-schema. It would replace the mostly static MIxS 5-based `src/schema/mixs.yaml`

See these issue comments:
- https://github.com/microbiomedata/nmdc-schema/issues/291#issuecomment-1093003740
- https://github.com/microbiomedata/nmdc-schema/issues/291#issuecomment-1093031441

turbomam:
This PR now includes all MIxS 6 slots used by all environmental packages in the NMDC DH schema. 

## Notes

- No packages have been curated except for soil. That means that there are probably slots in here that we may choose not to include
- Although the NMDC DH schema is examined to see what slots should be imported, the imports are taken from the MIxS schema itself, and do not include any modifications present in the NMDC DH schema.

turbomam:
## Manual checks that should probably become tests 

https://github.com/microbiomedata/nmdc-schema/blob/issue-291-mixs-submod/src/schema/mixs_6_for_nmdc.yaml

- [ ] `EMSL`: omit, since it's a **dh_section**
- [ ] `gender_restroom`: yes, this is new for us (from built environment?) and should be included
- [ ] `source_mat_id`: include but without "globally unique ID" title

sujaypatil96:
Reviewing this. Will add comments soon.

sujaypatil96:
I'm reviewing this PR and the new submodule process, by running and testing the Makefile targets. Here is the list of commands that I ran one after the other:

```
git checkout -b issue-291-mixs-submod   # make sure you are on the right branch
poetry install   # install all necessary dependencies in a virtualenv

git submodule init   # initialize local configuration file
git submodule update   # fetch data from above project

make mixs_clean   # remove all mixs specific files
make reports/slot_roster.tsv   # class slot relationships from MIxS and NMDC
make src/schema/mixs_new.yaml   # NMDC specific new MIxS 6 YAML
```

The mixs_new.yaml file got generated successfully and replaced the old mixs.yaml.

sujaypatil96 (approved):

PR #296, opened 2022-04-11
bytes slot rage: xsd:int to xsd:long

dwinston:
Some data objects in the NMDC metadata have file_size_bytes > 2GB, e.g. https://api.dev.microbiomedata.org/data_objects/jgi:55a9caff0d87852b2150891e

### DIFF
--- a/src/schema/core.yaml
+++ b/src/schema/core.yaml
@@ -45,7 +45,7 @@ types:
     description: >-
       An integer value that corresponds to a size in bytes
     base: int
-    uri: xsd:int
+    uri: xsd:long
     see_also:
       - UO:0000233
   


dwinston:
See https://www.w3.org/TR/xmlschema11-2/#int

turbomam:
This just seems like term metadata, so unlikely to break anything. I'm inclined to merge this in.

src/schema/core.yaml
```diff
@@ -45,7 +45,7 @@ types:
     description: >-
       An integer value that corresponds to a size in bytes
     base: int
-    uri: xsd:int
+    uri: xsd:long
```
sujaypatil96, line 48: @dwinston: I think the `base` type might need to be modified rather than the `uri`?

turbomam: Thanks @sujaypatil96 !

I find the LinkML `type_definition` confusing! Here's two links. It seems like there is some refactoring going on in the Documentation generation and deployment, so zero, one or both may work!  

- https://linkml.io/linkml-model/docs/TypeDefinition/ 
- https://linkml.io/linkml-model/docs/types/

turbomam: Is there a good way to test this?

cmungall: Let's merge this. The base type doesn't need modified, python ints are already longs

```python
>>> 2**256
115792089237316195423570985008687907853269984665640564039457584007913129639936
>>> type(2**256)
<class 'int'>
```

It's always good to test things and in general we want more integration tests to be able to sure that any schema changes doesn't have undesired side effects on other parts of the system

However, I think this is safe. The change will be silent for any generated python or json-schema

https://json-schema.org/understanding-json-schema/reference/numeric.html

cmungall (approved):

src/schema/core.yaml
cmungall: on second thoughts, the PR should be amended to use `xsd:integer`, which has no limits

 - https://www.w3.org/TR/xmlschema-2/#integer
 - https://www.w3.org/TR/xmlschema-2/#int

turbomam (commented):
accepted

PR #297, opened 2022-04-11
add test case to check if unit is enforced on depth schema element

sujaypatil96:
Adding a test case to check if adding a depth field without a `has_unit` attribute results in a validation error. But apparently it does not. The way the schema is currently, it accepts `depth` without the `unit`, which we need to change.

CC: @cmungall @wdduncan

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -214,7 +214,8 @@ SCHEMA_TEST_EXAMPLES_INVALID := \
 	biosample_missing_required_field \
 	biosample_single_multi_value_mixup \
 	biosample_undeclared_slot \
-	study_credit_enum_mangle
+	study_credit_enum_mangle \
+	depth_unit_required
 
 # 	functional_annotation_set_invalid has invalid ID pattern but regex tests aren't applied yet? MAM 2021-06-24
 
--- a/test/data/invalid_schemas/depth_unit_required.json
+++ b/test/data/invalid_schemas/depth_unit_required.json
@@ -1,0 +1,124 @@
+{
+    "biosample_set": [
+      {
+        "id": "gold:Gb0101224",
+        "name": "Lithgow State Coal Mine Calcium nutrients (early)",
+        "description": "Bulk Aqueous phase filtered water",
+        "type": "nmdc:Biosample",
+        "part_of": [
+          "gold:Gs0128849"
+        ],
+        "depth": {
+            "has_numeric_value": 0.05
+        },
+        "env_broad_scale": {
+          "has_raw_value": "ENVO:00002030"
+        },
+        "env_local_scale": {
+          "has_raw_value": "ENVO:00002169"
+        },
+        "env_medium": {
+          "has_raw_value": "ENVO:00005792"
+        },
+        "geo_loc_name": {
+          "has_raw_value": "Lithgow"
+        },
+        "lat_lon": {
+          "has_raw_value": "-33.460524 150.168149",
+          "latitude": -33.460524,
+          "longitude": 150.168149
+        },
+        "ecosystem": "Environmental",
+        "ecosystem_category": "Aquatic",
+        "ecosystem_type": "Freshwater",
+        "ecosystem_subtype": "Groundwater",
+        "specific_ecosystem": "Coalbed water",
+        "add_date": "28-JUL-14 12.00.00.000000000 AM",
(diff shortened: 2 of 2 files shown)


sujaypatil96:
This test case needs to be fixed with changes to the schema, perhaps enforcing a `required: True` on `has_unit`?

cmungall:
I agree with this course of action, but I would first check mongo. I would *hope* that we don't have a ton of depths with implicit units but good to check

turbomam:
@sujaypatil96 Do you have access to [NERSC](https://docs.nersc.gov/accounts/) [SPIN](https://www.nersc.gov/systems/spin/) or NMDC mongodb access yet?

@dehays and Marcin helped me get my initial NERSC account and @scanon helped me get associated with the necessary projects.

I think @dwinston has to create the mongodb user and password

I can you use these resources once your credentials and project associations are complete.

cmungall:
Is Nersc/spin required for read access? Can’t this just be done via api?

dehays:
@sujaypatil96 @turbomam @cmungall 

Some clarification - you can query NMDC Mongo via API: https://api.dev.microbiomedata.org/docs#/queries/run_query_queries_run_post

You will need an API user-psw which Donny can provide.  

If you want to go to NMDC Mongo directly - you will need a NERSC account (but not a Spin account) in order to ssh tunnel through a machine that can touch DB ports on Spin.  You can request an account from NERSC - NMDC's project is m3408.

I don't think anything you want to do requires Spin / Rancher access.  If so, you need to do NERSC's Spin training or self-training and convince Cory Snavely (NERSC) that you should have access.

dwinston:
@cmungall looks like we have 5 biosamples with no explicit depth.has_unit:

https://api.dev.microbiomedata.org/nmdcschema/biosample_set?filter={%22depth%22%3A%20{%22%24exists%22%3A%20true}%2C%20%22depth.has_unit%22%3A%20{%22%24exists%22%3A%20false}}&max_page_size=1000

(from entering `biosample_set` for `collection` and `{"depth": {"$exists": true}, "depth.has_unit": {"$exists": false}}` for `filter` at https://api.dev.microbiomedata.org/docs#/metadata/list_from_collection_nmdcschema__collection_name__get

cmungall:
looks like this is cm:

https://app.geosamples.org/sample/igsn/IEWFS000I

IGSN has this as ranges but we are treating as the upper bound?

dehays:
2 comments - 

Those are the 5 Brodie samples that DID NOT come through GOLD ETL.  (They were only used at EMSL and were never in GOLD.)

With regard to representing range values for depth - the same query looking at depth2 (The GOLD and at least temporarily NMDC attribute needed for depth ranges.) shows 178 without units.

sujaypatil96:
So we need to do two things here:
* retroactively fix the records in Mongo that have `depth` recorded without unit
* modify depth slot with slot_usage on biosample class by forcing unit like:
```
classes:
   biosample:
   slots:
   - depth
    slot_usage:
    - depth:
       - range: float  
       - unit:
           - ucum_code: m
```

@cmungall: am I proposing a sensible solution here?

turbomam:
@sujaypatil96 I support this suggestion, but we should

turbomam:
I rewrote the query to hit the current production MongoDB, and it found ~ 200

https://api.microbiomedata.org/nmdcschema/biosample_set?filter=%7B%22depth%22%3A%20%7B%22%24exists%22%3A%20true%7D%2C%20%22depth.has_unit%22%3A%20%7B%22%24exists%22%3A%20false%7D%7D&max_page_size=1000

turbomam:
I think this PR should be refactored to contribute an invalid data file in

sujaypatil96:
@turbomam I agree with your solution on the way forward for this PR.

The reason I made this PR was because when @cmungall and I were reviewing certain data we noticed that _depth_ didn't have unit information associated with it, and we thought oh no, depth definitely should have the unit it was captured in tagged to it.

I would like to extrapolate this idea, and ask the question that if we're going to enforce _requiredness_ on `has_unit` for depth, then should we do it for all measurement value slots that have a range of `QuantityValue` or would that be too big a breaking change to chew off in one bite?

cmungall:
I don't have strong opinions on how you break this down but QV.has_unit
absolutely must be required!!!

Issue #317, opened 2022-05-24
integrate non-mixs classes from the NMDC DH schema 


turbomam:
- EMSL
- JGI *
- MIxS inspired

turbomam:
https://github.com/microbiomedata/nmdc-schema/tree/issue-317-dh-non-mixs

turbomam:
These are all in `src/schema/portal` now

PR #319, opened 2022-05-25
biosample slots from portal

turbomam:
aka DataHarmonizer

#317


turbomam:
merging per 2022-06-01 metadata meeting

Issue #311, opened 2022-05-19
Model Stan's Bioscales elements/nutrients terms here


turbomam:
Were originally requested in MIxS: https://github.com/GenomicsStandardsConsortium/mixs/issues?q=is%3Aissue+author%3AStantonMartin

Some have been implemented in MIxS already. Questions remain for the others.

[Copy of GA lab Data 11_17_2020 (1).xlsx](https://github.com/microbiomedata/nmdc-schema/files/8730598/Copy.of.GA.lab.Data.11_17_2020.1.xlsx) illustrates some additional new terms required for Bioscales

CC @StantonMartin

turbomam:
[carbon terms for Bioscales (plus errors) #7](https://github.com/microbiomedata/mixs/issues/7)

ssarrafan:
@turbomam reported this morning that this issue is done. Closing issue.

PR #320, opened 2022-05-25
initial bioscales commit

turbomam:
#311

### DIFF
--- a/docs/schema-slides.html
+++ b/docs/schema-slides.html
@@ -10,76 +10,73 @@
   <meta name="date" content="2020-05-13" />
   <title>NMDC Schema</title>
   <style type="text/css">
-    code{white-space: pre-wrap;}
-    span.smallcaps{font-variant: small-caps;}
-    span.underline{text-decoration: underline;}
-    div.column{display: inline-block; vertical-align: top; width: 50%;}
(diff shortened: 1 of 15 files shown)


StantonMartin (commented):
Looks reasonable.  Please advise on the best way to test this once the commit is finalized.

turbomam:
> Looks reasonable. Please advise on the best way to test this once the commit is finalized.

Thanks @StantonMartin 

My next step is telling microbiomedata/sheets_and_friends to generate a Bioscales template. We should consult with @mslarae13 regarding your plan to eliminate columns. I think we can eliminate many, but I don't want to do it in a vacuum

turbomam:
merging per NMDC metadata meeting 2022-06-01

mslarae13:
We shouldn't eliminate columns. All the standard columns that are in the soil template should be there. The point here isn't REALLY to make a 'bioscales template' but rather identify the additional terms bioscales needs to expand on the soil template & revisit those terms & determine if they're worth adding to the "normal" soil template.

Any fields irrelevant to bioscales can be left empty. If you run into any required fields you don't have terms for, let me know & we can discuss.

For your plant samples, have you looked at the plant template Stan? Are there terms you need added to that too?

Issue #328, opened 2022-06-01
create a user submitted ID field


turbomam:
The user submitted ID field could follow the spirit of other external ID fields like INDC identifier(s?)

The existing id field, inherited from named thing, should be populated with IDs minted by https://api.dev.microbiomedata.org/docs#/identifiers

turbomam:
https://api.dev.microbiomedata.org/docs#/identifiers

mslarae13:
I thought we decided we were removing the 'identifiers' slot?
id vs identifier is too confusing.

turbomam:
OK. Given the mongodb -> schema 7 migration @mslarae13 I and I have been doing, it looks like `source_mat_id` will serve the purpose I was looking for. Maybe we can clarify the slot descriptions to differentiate between the NMDC-minted identifier in `id` and `source_mat_id`

turbomam:
Good progress. We could always improve the guidance to submitters.

PR #335, opened 2022-06-06
just description and annotations on submitter_id

turbomam:
for #328

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -132,6 +132,7 @@ classes:
         Samples are associated with checklists, which define the fields used to annotate
         the samples. Samples are always associated with a taxon.
     slots:
+      - submitter_id
       - type  # custom slot that specifies object type
       - alternative identifiers
       - part of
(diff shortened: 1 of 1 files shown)


turbomam:
I don't think I have gotten enough eyes on this from the metadata team yet. I think I was intending for `submitter_id ` to replace `source_mat_id`, which is entitled 'globally unique ID' in https://github.com/turbomam/sheets_and_friends/blob/main/artifacts/nmdc_submission_schema.yaml

sujaypatil96 (approved):
It looks like you're proposing the addition of a `submitter_id` slot that will eventually replace `source_mat_id` which we are using from MIxS. And as part of this PR you haven't remove `source_mat_id` yet, so that shouldn't affect too many things upstream data collection interfaces like the DH powered data submission portal, sheets_and_friends, etc. But we should probably talk about this phase out of `source_mat_id` in favour of `submitter_id` on the metadata call.

Approving this PR.

turbomam:
Discussions with @mslarae13 have led to to believe that we should be using MIxS terms like `source_mat_id` and not minting our own slots if possible.

Issue #336, opened 2022-06-06
add the bioscales template itself (not it's slots) to nmdc.yaml


turbomam:
possible through an import statement

will require some dependent terms

turbomam:
@cmungall and @mslarae13 :

turbomam:
Oops, I created a `bioscales` class (`is_a: dh_interface`) and a `Bioscales` class (`is_a: dh_section`)

I'll have to fix that. Any suggestions?

turbomam:
I don't see any `Bioscales` class anymore. The bioscales terms are imported from `src/schema/bioscales.yaml`

That does address the issue of the potential growth of nmdc-schema.yaml when all of the submission schema stuff in merged in.

@cmungall has a schema fixer that might remove slot_usage statements that re-assert attributes that are already asserted at the schema level.

see https://github.com/linkml/linkml/issues/909

PR #340, opened 2022-06-06
#336 bioscales template IN nmdc schema

turbomam:
What would happen if we defined the DataHarmonizer templates for the submission portal **in** nmdc-schema, as opposed to building them in sheets_and_friends?

Eventually we will probably want to support the creation and of these templates with schema_sheets


turbomam:
this builds, but there are some circular linkml imports

turbomam:
I believe that everything provided by `src/schema/dh_templates.yaml` here is in the current main branch's `src/schema/portal`

It doesn't look like `semweb_context` is in the `default_curi_maps` yet, but that may not be harming anything.

turbomam:
Actually I meant that `default_curi_maps` was already in main

Issue #366, opened 2022-07-25
prepare change sheet for omics processing processing_institutions


turbomam:
using values from processing_institution_enum

PR #368, opened 2022-07-25
observed and changesheet

turbomam:
observed and changesheet files in `util`. Will tidy up later.

for #366


sujaypatil96 (approved):
Changesheet looks good to me @turbomam.

One thing I would recommend is while you're tidying up is probably creating a folder called, `changesheets` at the root of the repository, and putting all changesheets in there. We can probably create sub folders based on the month and the year and the filename with a timestamp suffix in the filename, like, `changesheets > jul_2022 > changesheet_2022_07_25`

Issue #386, opened 2022-08-05
add sample_link to Biosample modeling


turbomam:
may be related to or duplicate of some other issues?

Biosample 1..* ? Named Thing?

We're intending to use this with Study ranges

remove requirement for `part of`

PR #387, opened 2022-08-05
Biosamples must link to Studies via sample_link instead of part_of

turbomam:
#386


turbomam:
I'm surprised that the test test biosamples validated, assuming they don't have sample_link values

Issue #347, opened 2022-06-15
low hanging build/load error messages


turbomam:
## Patterns

- [ ] INFO:root:class "chemical entity" slot "chemical formula" does not reference an existing slot.  New slot was created.
- [x] INFO:root:Default_range not specified. Default set to 'string'
- [x] Unrecognized prefix: COG
- [x] WARNING:root:File "<file>", someline/somecol: Prefix case mismatch - supplied: MIXS expected: mixs
- [ ] WARNING:root:No namespace defined for URI: https://microbiomedata/schema/annotation/FunctionalAnnotation

This duplicates a few other issues

- #234 
- #232 
- #103

turbomam:
## Verbose


```
INFO:root:class "chemical entity" slot "chemical formula" does not reference an existing slot.  New slot was created.
INFO:root:class "chemical entity" slot "inchi key" does not reference an existing slot.  New slot was created.
INFO:root:class "chemical entity" slot "inchi" does not reference an existing slot.  New slot was created.
INFO:root:class "chemical entity" slot "smiles" does not reference an existing slot.  New slot was created.
INFO:root:class "database" slot "date created" does not reference an existing slot.  New slot was created.
INFO:root:class "database" slot "etl software version" does not reference an existing slot.  New slot was created.
INFO:root:class "database" slot "nmdc schema version" does not reference an existing slot.  New slot was created.
INFO:root:class "genome feature" slot "encodes" does not reference an existing slot.  New slot was created.
INFO:root:class "genome feature" slot "end" does not reference an existing slot.  New slot was created.
INFO:root:class "genome feature" slot "feature type" does not reference an existing slot.  New slot was created.
INFO:root:class "genome feature" slot "phase" does not reference an existing slot.  New slot was created.
INFO:root:class "genome feature" slot "seqid" does not reference an existing slot.  New slot was created.
INFO:root:class "genome feature" slot "start" does not reference an existing slot.  New slot was created.
INFO:root:class "genome feature" slot "strand" does not reference an existing slot.  New slot was created.
INFO:root:class "metabolite quantification" slot "highest similarity score" does not reference an existing slot.  New slot was created.
INFO:root:class "metabolite quantification" slot "metabolite quantified" does not reference an existing slot.  New slot was created.
INFO:root:class "metabolomics analysis activity" slot "has calibration" does not reference an existing slot.  New slot was created.
[...]

PR #394, opened 2022-08-09
many prefix/ns solutions & define slots outside of usage

turbomam:
#347 

will require followup for 
- example.com prefix expansion placeholders
- unexpected prefix definition requirements like nmdc_mixs
- slots that were only defined in a slot usage now have stubs at the schema level


turbomam:
also 
- some general reformatting
- changing `mixs` prefixes to `MIXS`

PR #405, opened 2022-08-12
Update workflow_execution_activity.yaml

hubin-keio:
Added description and title of read based analysis.

Shall we rename it to "read based taxonomy analysis activity?"

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -104,7 +104,7 @@ classes:
       - nom analysis activity set
       - omics processing set
       - read QC analysis activity set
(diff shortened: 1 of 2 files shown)


sujaypatil96 (approved):
Read based taxonomy analysis activity definitely sounds better.

turbomam:
Thanks for reviewing this PR, @sujaypatil96. It's really not a matter of what sounds better. It's a matter of aligning the term name with the term `description`. In this case, there is good alignment, so I will make that additional change and merge it in.

turbomam (commented):
approved

PR #415, opened 2022-08-20
Added description of metagenome assembly

hubin-keio:


src/schema/workflow_execution_activity.yaml
```diff
@@ -66,6 +66,7 @@ classes:
         required: true
 
   metagenome assembly:
+    description: A workflow execution activity that dose de novo genome assembly using metagenome sequencing reads
```
wdduncan, line 69: Should this says **does** de novo?

turbomam: good catch

turbomam:
Thanks, @bin!

Also, are all "metagenome assembly"s inherently de novo? They can never be guided by reference genomes?

Or are we saying that at this point in time, NMDC doesn't use any guided metagenome assemblies?

The name and description of the class should agree in a 1:1 fashion.

turbomam:
https://www.cs.umd.edu/sites/default/files/scholarly_papers/CepedaEspinoza.pdf

turbomam (approved):
I moved "de novo" from the description to comments because nobody could explain NMDC's usage of de novo metagenome assemblers vs reference-guided assemblers

We c an always change that in the future

hubin-keio:
Sorry for my late reply. The assembly is de novo in our workflow.

PR #418, opened 2022-08-22
schema maintenance and cleanup

sujaypatil96:
* remove unnecessary files
* update README to reflect latest content


turbomam (approved):
The changes to the README.md (including the badges) were more than I expected but otherwise this is good.

Issue #426, opened 2022-08-23
diff some v3.0.0 and v4.0.0 (or v6.0.3) JSON artifacts for Monet


turbomam:
For @corilo and Natasha

turbomam:
https://github.com/microbiomedata/nmdc-schema/releases/tag/v4.0.0

vs

PR #427, opened 2022-08-23
jd schema diff 3.2.0 v 4.0.0

turbomam:
#426 

`jd` is a system dependancy. It can be installed with `brew`


turbomam:
closes #426

PR #441, opened 2022-08-31
show middle layer elements in inheritance tree in mermaid

sujaypatil96:
This PR seeks to address one major point, and a minor point.

Major point being, we need to add some jinja code to allow for the better rendering of mermaid diagrams that show inheritance trees summarizing `is_a` relationships for schema classes that have both parents and children.

For example: [BiosampleProcessing](https://microbiomedata.github.io/nmdc-schema/BiosampleProcessing/)
Screen Shot 2022-08-31 at 2 05 31 PM

Minor point is, we need to standardize what we're calling the LinkML source sections on both:

* class pages: https://microbiomedata.github.io/nmdc-schema/BiosampleProcessing/#linkml-specification
* slot pages: https://microbiomedata.github.io/nmdc-schema/has_input/#linkml-source

Q) LinkML source or LinkML specification?

### DIFF
--- a/src/doc-templates/class.md.jinja2
+++ b/src/doc-templates/class.md.jinja2
@@ -115,7 +115,7 @@ URI: {{ gen.uri_link(element) }}
 {% endif -%}
 
 
-## LinkML Specification
+## LinkML Source
 
 ### Direct
 
--- a/src/doc-templates/class.md.jinja2
+++ b/src/doc-templates/class.md.jinja2
@@ -19,11 +19,10 @@ _{{ element.description }}_
 
 URI: {{ gen.uri_link(element) }}
 
-
-
-{% if schemaview.class_parents(element.name) %}
+{% if schemaview.class_parents(element.name) and schemaview.class_children(element.name) %}
 ```mermaid
  classDiagram
+    class {{ gen.name(element) }}
       {% for s in schemaview.class_parents(element.name)|sort(attribute='name') -%}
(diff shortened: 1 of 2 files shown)


turbomam:
Great!

I have a PR that disables class mermaid diagrams. I haven't merged it yet. If we can address these two point below, then I'll just close it.

My requests (to be included in this PR or something to follow):
- there should be a filter that says (for example) "Biosample has so many slots, they're not worth displaying in the Mermaid diagram"
- we should expand the goal of this PR to display not just middle layers from a is_a/mixin perspective, but all class-to-class relationships. So the Mermaid diagram for Biosample should also show its relation to Study.

turbomam:
Q) LinkML source or LinkML specification?

Yes, we should be consistent. What do you prefer? I have a slight preference for LinkML source

sujaypatil96:
@turbomam: to answer your requests:

> there should be a filter that says (for example) "Biosample has so many slots, they're not worth displaying in the Mermaid diagram"

Hm, I'm not sure if I have that level of control over the HTML, CSS and JS that renders these diagrams, but I can check

> we should expand the goal of this PR to display not just middle layers from a is_a/mixin perspective, but all class-to-class relationships. So the Mermaid diagram for Biosample should also show its relation to Study.

This should definitely be possible. If I understand correctly you're asking that we show the sibling relationship for a class in addition to the is_a relationship between parent and child?

> Yes, we should be consistent. What do you prefer? I have a slight preference for LinkML source

Let's go with LinkML Source

turbomam (approved):

PR #444, opened 2022-09-08
Notebooks that might be of general utility and unrelated no-mermaid Jinja2 templates

turbomam:
Some of these notebooks are probably might be useful in other repos, but this PR is a mess.


turbomam:
manually merged into
- #682

closing

Issue #428, opened 2022-08-23
lingering build/load error messages


turbomam:
closes #347

turbomam:
```

INFO:root:class "database" slot "date created" does not reference an existing slot.  New slot was created.
INFO:root:class "database" slot "etl software version" does not reference an existing slot.  New slot was created.
INFO:root:class "database" slot "nmdc schema version" does not reference an existing slot.  New slot was created.
INFO:root:class "metaproteomics analysis activity" slot "has peptide quantifications" does not reference an existing slot.  New slot was created.
INFO:root:class "pathway" slot "has_part" does not reference an existing slot.  New slot was created.
INFO:root:class "reaction participant" slot "chemical" does not reference an existing slot.  New slot was created.
INFO:root:class "reaction participant" slot "stoichiometry" does not reference an existing slot.  New slot was created.
INFO:root:class "study" slot "notes" does not reference an existing slot.  New slot was created.

INFO:root:Default_range not specified. Default set to 'string'


WARNING:root:No namespace defined for URI: https://microbiomedata/schema/workflow_execution_activity/asm_score
WARNING:root:No namespace defined for URI: https://microbiomedata/schema/workflow_execution_activity/binned_contig_num
WARNING:root:No namespace defined for URI: https://microbiomedata/schema/workflow_execution_activity/contig_bp
WARNING:root:No namespace defined for URI: https://microbiomedata/schema/workflow_execution_activity/contigs
WARNING:root:No namespace defined for URI: https://microbiomedata/schema/workflow_execution_activity/ctg_L50
WARNING:root:No namespace defined for URI: https://microbiomedata/schema/workflow_execution_activity/ctg_L90
WARNING:root:No namespace defined for URI: https://microbiomedata/schema/workflow_execution_activity/ctg_logsum
WARNING:root:No namespace defined for URI: https://microbiomedata/schema/workflow_execution_activity/ctg_max
WARNING:root:No namespace defined for URI: https://microbiomedata/schema/workflow_execution_activity/ctg_N50
WARNING:root:No namespace defined for URI: https://microbiomedata/schema/workflow_execution_activity/ctg_N90
WARNING:root:No namespace defined for URI: https://microbiomedata/schema/workflow_execution_activity/ctg_powsum
WARNING:root:No namespace defined for URI: https://microbiomedata/schema/workflow_execution_activity/gap_pct
[...]

PR #447, opened 2022-09-08
Eliminated warnings and concerning "INFO"s when loading schema

turbomam:
#428


turbomam:
- moved some comments outside of the `target/python/%.py` Makefile rule 
- sorted `imports`, `prefixes` and `default_curi_maps`
  - **ordering of `default_curi_maps` might be significant**
- some prefixes are still set to expand to http://example.com/
- all schema modules now have `default_range` and `default_prefix`
  - **nmdc prefix isn't resolvable**
  - **mixs/MIXS prefix needs work too.**
  - **prov?**. esp default_prefix
- moved some usage-only slots out to the module's outer slots: block
  - like `has part` in `src/schema/annotation.yaml`  , several in `src/schema/nmdc.yaml`, and more
  - note that this was also a rename operation from `has_part` to `has part`, to be consistent with `part of`, although I would like to normalize all slots to lower snake case

turbomam:
- ./annotation.yaml:  COG: http://example.com/
- ./annotation.yaml:  EC: http://example.com/
- ./annotation.yaml:  MetaCyc: http://example.com/
- ./annotation.yaml:  MetaNetX: http://example.com/
- ./annotation.yaml:  RetroRules: http://example.com/
- ./core.yaml:  UniProtKB: http://example.com/
- ./core.yaml:  gtpo: http://example.com/
- ./nmdc.yaml:  COG: http://example.com/
- ./nmdc.yaml:  EC: http://example.com/
- ./nmdc.yaml:  ISA: http://example.com/
- ./nmdc.yaml:  MetaCyc: http://example.com/
- ./nmdc.yaml:  MetaNetX: http://example.com/
- ./nmdc.yaml:  RetroRules: http://example.com/
- ./nmdc.yaml:  UniProtKB: http://example.com/
- ./nmdc.yaml:  gtpo: http://example.com/
- ./nmdc.yaml:  insdc.srs: http://example.com/
- ./nmdc.yaml:  mgnify: http://example.com/

grep -R ': https://example.com' | grep -v 'from_schema:' ? none

turbomam:
https://github.com/perma-id/w3id.org/blob/master/nmdc/.htaccess

> Options +FollowSymLinks
RewriteEngine on
RewriteRule ^(.*) https://github.com/microbiomedata/ [R=307,L]

So https://w3id.org/nmdc/Study -> https://github.com/microbiomedata/

But should -> https://microbiomedata.github.io/nmdc-schema/Study/

Anyway, the nmdc prefix should expand to https://w3id.org/nmdc/, at least on a term-by-term basis.

It is currently is set to 'https://microbiomedata/meta/'

turbomam:
Name-based MIxS docs page: https://genomicsstandardsconsortium.github.io/mixs/Water/

There has been some talk of switching to id-based pages. Could we have a hybrid within the NMDC schema and it's docs pages? Name base pages for NMDC and id-based for MIxS?

Is this redirection coming from either or both?
- https://github.com/perma-id/w3id.org/blob/master/mixs/.htaccess
- https://github.com/perma-id/w3id.org/blob/master/gensc/.htaccess

try 
- https://w3id.org/mixs/Water -> https://genomicsstandardsconsortium.github.io/mixs/Water/
- ~~https://w3id.org/gensc/Water~~
- ~~https://w3id.org/gensc/mixs/Water~~

therefore, prefix `mixs` should expand to https://w3id.org/mixs

Also, since the w3id namespace is lowercase `mixs`, shouldn't the corresponding LinkML prefix be lowercase, too? Check the NMDC and MIxS schemas for usage of the uppercase variant `MIXS`

turbomam:
This claimed mapping in `src/schema/mixs.yaml` doesn't currently exist:

```yaml
  env_package:
    name: env_package
    aliases:
    - environmental package
    mappings:
    - mixs:env_package
```

src/schema/annotation.yaml
```diff
@@ -4,25 +4,36 @@ title: Annotation Module for NMDC Schema
 description: >-
   This module in the schema is for representing annotations including functional annotations of proteins and other gene products,
   as well as controlled terms for describing things like metabolites
+default_range: string
 
 see_also:
   - https://github.com/microbiomedata/nmdc-metadata/issues/176
 
 default_curi_maps:
-  - obo_context
```
cmungall, line 12: what's the reason for this? obo should take precedence

src/schema/annotation.yaml
```diff
@@ -4,25 +4,36 @@ title: Annotation Module for NMDC Schema
...
   - idot_context
+  - obo_context
 
 prefixes:
-  nmdc: https://microbiomedata/meta/
+  COG: http://example.com/
```
cmungall, line 17: if we're going to make this change we should use the correct URLs

I think the linter will suggest the right URLs, cc @pkalita-lbl

pkalita-lbl: The `canonical_prefixes` rule is a pretty light-weight wrapper around https://github.com/linkml/prefixmaps. You can provide the `prefixmaps_contexts` option as a list of strings that will be passed to prefixmaps' `load_multi_context` to prioritize some prefix registries over others (the prefixmaps docs can explain it better than me 😁). But that's all there is in terms of configuration for now. What else would be helpful for you?

sujaypatil96 (approved):

PR #463, opened 2022-09-15
trim contents `nmdc_schema`

sujaypatil96:
* Filter contents of nmdc_schema to include only files that are useful when pushed to the PyPI repo.
* Update contents after latest regeneration of artifacts as in v7.0.0

Eventually I think the process needs to be updated so that we don't have this folder altogether. The package should be built and published by poetry on the fly.


sujaypatil96:
This PR was pre emptive. Closing it. Didn't realize it affected a lot of other downstream scripts.

Issue #470, opened 2022-09-22
dump biosample, study and omics processing Mongo collections inot an NMDC "Database" 


turbomam:
Since we don't have any visual explorer yet, at least people can do manual, one-stop, cross-collection searching in a document like this

brynnz22:
@turbomam this issue looks like its related to the linkml one you created: https://github.com/linkml/linkml/pull/1609?

aclum:
Can this be closed now that we have the data in graphdb?

PR #471, opened 2022-09-22
studies/biosamples fast

turbomam:
#470 

activities slow


dwinston:
@turbomam let’s make a “metadata export” endpoint in nmdc-runtime to do this much more efficiently.

turbomam:
> @turbomam let’s make a “metadata export” endpoint in nmdc-runtime to do this much more efficiently.

Agreed. And one of the acceptable formats should be an NMDC database object, in JSON or YAML. In other words, the output should validate against the schema with no pre-processing.

That's as opposed to output that has top level keys for number of pages or next page token, and wraps the actual content in a "resources" key, etc.

Would that be challenging for a large download where pagination typically would be applied?

dwinston:
this job would not need to go through HTTP requests like a normal API client. It could connect directly to the db and yes, produce a nmdc:Database JSON object.

Issue #401, opened 2022-08-12
slot 'GOLD sample identifiers' is unclear


mslarae13:
The slot 'GOLD sample identifiers' is identified as used in the class 'study'

The definition of 'GOLD sample identifiers' has value as 'Provide the GOLD study IDs associated with samples for this study.'

Few things

1. That's not a value, isn't that a description?
2. The description fits for study class. BUT, the term says SAMPLE identifiers, which are different from the GOLD study ID.

Which is this supposed to be, how can we update it to be more clear, do we need to add a term?

@turbomam 
@emileyfadrosh 
@sujaypatil96 
@pvangay

emileyfadrosh:
@mslarae13 I think I need some help finding where this is to provide a reasonable response. I don't believe GOLD has a "sample identifier" term, instead there is: GOLD Biosample ID; Biosample name; identifer (see https://gold.jgi.doe.gov/biosample?id=Gb0144001 for example). If 'GOLD sample identifiers' relates to one of the above, we should take the definition directly from GOLD and make sure we are not mixing up study and biosample terms from GOLD. I can do a PR to make an update if you point me to where this is at :)

ssarrafan:
Not assigned to anyone so I'm removing from the Aug sprint. FYI @mslarae13 @emileyfadrosh

mslarae13:
Success here looks like
GOLD sample identifiers slot is "cleaned up"
- The description is made more clear for what it's ACTUALLY identifying.
- If it is meant to be under class study, the slot is re-named to say study, not sample
- If it is meant to capture samples, it is removed from study class & put in biosample class.

turbomam:
This is a good issue, but I don't think much or any action is required. I disagree that slot `GOLD sample identifiers` is used by class `Study`. @mslarae13 , if you have a document that says otherwise, please point me towards it. I hope I didn't do something to confuse you!

First, I searched for all mentions or definitions of `GOLD sample identifiers` in the src/schema directory:

```shell
grep -R 'GOLD sample identifiers' *yaml
```

> external_identifiers.yaml:  GOLD sample identifiers:
nmdc.yaml:      - GOLD sample identifiers
nmdc.yaml:      GOLD sample identifiers:

`GOLD sample identifiers` is defined in `external_identifiers.yaml` and mentioned in `nmdc.yaml`

## Definition

```yaml
  GOLD sample identifiers:
    is_a: sample identifiers
    mixins:
      - GOLD identifiers
[...]

PR #472, opened 2022-09-22
polish prev gold_sample_identifiers edits

turbomam:
migrated slot usage out of study; edited annotations

followup on #401


mslarae13 (approved):
update looks good and clarifies the slot!

PR #478, opened 2022-09-29
Add id patterns for Study and Biosample classes

pkalita-lbl:
Resolves https://github.com/microbiomedata/issues/issues/4

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -77,6 +77,13 @@ subsets:
     description: Subset consisting of the data objects that either inputs or outputs
       of processes or workflows.
 
+settings:
+  id_nmdc_prefix: "nmdc:"
+  id_type_sample: "sa"
+  id_type_study: "st"
+  id_shoulder: "[0-9][a-z]*[0-9]"
+  id_generated: "[a-z0-9]{6,}"
+
 classes:
   Database:
     tree_root: true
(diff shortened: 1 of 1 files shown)


src/schema/nmdc.yaml
```diff
@@ -404,6 +411,11 @@ classes:
       - lbceq
 
     slot_usage:
+      id: 
+        structured_pattern:
+          syntax: "{id_nmdc_prefix}{id_type_sample}{id_shoulder}{id_generated}"
```
dwinston, line 416: is there any weirdness about the  id_shoulder being greedy or not greedy?

dwinston (commented):
what does interpolated: true mean in this context?

src/schema/nmdc.yaml
pkalita-lbl: That's a good catch; I think the `[a-z]*` in the should part should be lazy. I'll fix.

`interpolated: true` means that values from `settings` should be interpolated into the `syntax` in order to construct the final regular expression: https://linkml.io/linkml/schemas/constraints.html#structured-patterns

dwinston: thanks. On inspection, I think your fix won't be necessary because a numeric character `[0-9]` is required at the shoulder end and is in no danger of being slurped up by `[a-z]*`. My only other comment is that perhaps the prefix should be "nmdc" and the ":" put as a literal into the syntax, but none of this is blocking. I approve.

dwinston (approved):

PR #485, opened 2022-10-06
Remove id structured_patterns

### LINKED ISSUE #484: roll back new NMDC id patterns
The new new NMDC id patterns (from https://github.com/microbiomedata/nmdc-schema/pull/478) look very good, but they can't be included in `main`'s schema yet. If the `structured_patterns` in the schema were expanded with `gen-linkml` and then converted to plain `pattern`s, then none of our existing biosample or study data would pass.

 I don't think the tests would even pass.

pkalita-lbl:
Fixes #484

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -418,11 +418,6 @@ classes:
             tag: display_hint
             value: >-
               Provide the GOLD biosample IDs associated with this biosample.
-      id: 
-        structured_pattern:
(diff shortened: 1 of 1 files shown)


turbomam (approved):
Thanks. Will it be easy for you load this back in at the appointed time, from some branch you haven't deleted yet?

pkalita-lbl:
It's not in a branch, but it will be easy to track it down in git history

Issue #330, opened 2022-06-03
Update and improve identifiers


mslarae13:
Identifiers, GUIDs, sample names, alternative IDs, 

The use of these fields and their definitions needs re-worked with possible addition of new terms. 

- [x] #369
- [x] #370
- [x] #371

turbomam:
See [NMDC Biosample Slots selected cols](https://docs.google.com/spreadsheets/d/1eJAa-FExSls3YZrsVVA2NjDIGRm68FsEh00ma8rgbyw/edit#gid=916732144). Will provide more documentation soon.

ssarrafan:
@mslarae13 would you like me to create 3 separate GH issues for this based on the three tasks above?  Update and improve seems to be too vague to be able to accomplish/close an issue.

mslarae13:
@ssarrafan  I agree & think this would make it easier to plan

ssarrafan:
> @ssarrafan I agree & think this would make it easier to plan

Ok I've created 3 separate issues related to this. I moved the first one to the August sprint and have not assigned the other two to a sprint yet. Let me know how you want to schedule them and if someone else should be assigned.  I'll move this overall issue to August so we can figure it out.

mslarae13:
Related to https://github.com/microbiomedata/nmdc-schema/issues/369

ssarrafan:
@mslarae13 moving to Sept sprint but if you plan to not work on it in the next 2 weeks let me know so I can remove and add the backlog label

mslarae13:
See also
https://github.com/microbiomedata/nmdc-schema/issues/401#issuecomment-1234641862
&
https://github.com/microbiomedata/nmdc-schema/issues/369#issuecomment-1234630202

ssarrafan:
Checked in with @turbomam and moving this to the squad sprint.  @mslarae13 let me know if you're not planning to work on this the rest of this month.

mslarae13:
Can this get added to the deinococcus squad board? @ssarrafan

ssarrafan:
Sure @mslarae13. I'll remove it from the current sprint and add it to the deino squad board until you're ready to work on it again then we can add it to the next sprint.  

FYI @turbomam

mslarae13:
Oh, no.. it can be in the upcoming sprint. I'll be working on it. It's just a task that rounds out the work I'm doing in the squad and it wasn't on the board. Sorry for the confusion @ssarrafan

[...]

PR #491, opened 2022-10-20
Issue330 update identifiers

mslarae13:
I added some identifier slots we've discussed and updated some descriptions.

Please review my changes & make sure everything still looks correct!

I did NOT add emsl_study_identifier because @turbomam  and I did that in a branch he had open.


mslarae13:
#330

turbomam (commented):
`make all` was failing in this branch because the slot `gold_study_identifiers` was defined in both `external_identifiers.yaml` (which uses the **schema** `id` https://microbiomedata/schema/external_identifiers) and in `nmdc.yaml`, which uses the `id` https://microbiomedata/schema

_Those schema `id`s need to be fixed up in another future branch._

I fixed the double definition in commit [3f46d86c07f62cc73d14efa1fbb665063f6632de](https://github.com/microbiomedata/nmdc-schema/pull/491/commits/3f46d86c07f62cc73d14efa1fbb665063f6632de) 

I still have to see what the merge conflict is regarding the main branch. Then I'll propose modifications to the `Biosample` class `description` . If that's acceptable, then we can merge this into main.

turbomam (commented):
The merge won't be difficult.

I will also be removing the `structured_pattern` assertions

turbomam (commented):
I think this would benefit from some spell checking, too

turbomam (approved):
great

Issue #493, opened 2022-11-08
Add Napa Protocol `id` pattern to schema


turbomam:
cc @dwinston @cmungall @sujaypatil96 @mslarae13 @scanon

turbomam:
per-class `id` induced attributes report can be generated with `nmdc-schema/util/id_usage.py`

`id` `pattern`s haven't been asserted for any classes yet

Excerpt:

> INDUCED id ATTRIBUTE FOR Activity:
> name: id
> description: A unique identifier for a thing. Must be either a CURIE shorthand for
>   a URI or a complete URI
> from_schema: https://microbiomedata/schema/basic_slots
> multivalued: false
> identifier: true
> alias: id
> owner: Activity
> domain_of:
> - Biosample
> - Study
> - NamedThing
> - Activity
> range: string
> 
> 
> 
> CLASS Agent DOESN'T USE THE id SLOT
> 
> 
> CLASS AttributeValue DOESN'T USE THE id SLOT
> 
> 
> INDUCED id ATTRIBUTE FOR Biosample:
> name: id
> description: An NMDC assigned unique identifier for a biosample submitted to NMDC.
> from_schema: https://microbiomedata/schema/basic_slots
> multivalued: false
> identifier: true
> alias: id
> owner: Biosample
> domain_of:
> - Biosample
> - Study
> - NamedThing
> - Activity
> range: string
> required: true

dwinston:
copying from @turbomam's NMDC slack post to faciliate extended discussion here:

abstraction:
```
prefix:typecode-authshoulder-blade(.version)?(_seqsuffix)?
```

regex:
```
^nmdc:[A-Za-z0-9]{1,6}-[A-Za-z0-9]{1,6}-[A-Za-z0-9]{1,6}(\.[A-Za-z0-9]{1,6})?(_.+)?$
```

dwinston:
FYI I started a draft spec (just boilerplate for now) using https://respec.org/docs/ and put it in this repo as `docs/napa-protocol.html` ([snapshot](https://github.com/microbiomedata/nmdc-schema/blob/66b401ab30fd6e0110414813a20e879c29487c4f/docs/napa-protocol.html)), so visible as <https://w3id.org/nmdc/napa-protocol.html>. Will be good to have such a document to guide implementation.

turbomam:
I love the railroad diagram and hope I won't forget about it. But it's time to close this generally satisfied issue.

PR #499, opened 2022-11-08
strawperson global id pattern

turbomam:
for #493

### DIFF
--- a/src/schema/basic_slots.yaml
+++ b/src/schema/basic_slots.yaml
@@ -11,7 +11,7 @@ prefixes:
   linkml: https://w3id.org/linkml/
   nmdc: https://w3id.org/nmdc/
   skos: http://www.w3.org/2004/02/skos/core#
-    
+
 default_prefix: nmdc
 default_range: string
 
@@ -31,7 +31,11 @@ slots:
     description: >-
       A unique identifier for a thing.
       Must be either a CURIE shorthand for a URI or a complete URI
-    #required: false # for now we are setting this to false until we develop an id template
+    notes:
+      - "relocated YAML comment: for now we are setting this to false until we develop an id template"
+      - "Doesn't 'identifier: true' imply 'required: true'?"
+      - "abstracted pattern: prefix:typecode-authshoulder-blade(.version)?(_seqsuffix)?"
+    pattern: ^nmdc:[A-Za-z0-9]{1,6}-[A-Za-z0-9]{1,6}-[A-Za-z0-9]{1,6}(\.[A-Za-z0-9]{1,6})?(_.+)?$
 
   name:
     multivalued: false
@@ -45,7 +49,7 @@ slots:
     description: >-
       a human-readable description of a thing
     slot_uri: dcterms:description
-    
+
   type:
     deprecated: >-
       Due to confusion about what values are used for this slot, it is best not to use this slot.
--- a/src/schema/basic_slots.yaml
+++ b/src/schema/basic_slots.yaml
@@ -36,7 +36,7 @@ slots:
       - 'a minimum length of 3 characters is suggested for typecodes, but 1 or 2 characters will be accepted'
       - 'typecodes must correspond 1:1 to a class in the NMDC schema. this will be checked via per-class id slot usage assertions'
       - 'minting authority shoulders should probably be enumerated and checked in the pattern'
-    pattern: ^nmdc:[a-z]{1,6}-[0-9][a-z]{0,6}[0-9]-[A-Za-z0-9]{1,}(\.[A-Za-z0-9]{1,})*(_[A-Za-z0-9_\.-]+)?$
+    pattern: ^(?<prefix>nmdc):(?<typecode>[a-z]{1,6})-(?<shoulder>[0-9][a-z]{0,6}[0-9])-(?<blade>[A-Za-z0-9]{1,})(?<version>(\.[A-Za-z0-9]{1,})*)(?<locus>_[A-Za-z0-9_\.-]+)?$
     examples:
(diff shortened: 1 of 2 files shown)


src/schema/basic_slots.yaml
```diff
@@ -31,7 +31,11 @@ slots:
     description: >-
       A unique identifier for a thing.
       Must be either a CURIE shorthand for a URI or a complete URI
-    #required: false # for now we are setting this to false until we develop an id template
+    notes:
+      - "relocated YAML comment: for now we are setting this to false until we develop an id template"
+      - "Doesn't 'identifier: true' imply 'required: true'?"
+      - "abstracted pattern: prefix:typecode-authshoulder-blade(.version)?(_seqsuffix)?"
+    pattern: ^nmdc:[A-Za-z0-9]{1,6}-[A-Za-z0-9]{1,6}-[A-Za-z0-9]{1,6}(\.[A-Za-z0-9]{1,6})?(_.+)?$
```
dwinston, line 38: blade should be open-ended length, i.e. `[A-Za-z0-9]{1,}`

sujaypatil96:
@turbomam: great job on the regex, it's certainly beginning to look a lot like the _typed identifier with version number_ that was unanimously agreed upon by the team. An example of a metaproteomic sample identifier with contig information would certainly help refine the regex if need be. CC: @scanon @pdpiehowski 

Although I'm not fully understanding the `id_usage.py` script in this PR? Is it's purpose meant to be a utility method? Or an example of sorts?

src/schema/basic_slots.yaml
sujaypatil96: Hm, I don't think I'm understanding the purpose of the last grouping `_.+`?

sujaypatil96: I think the regex pattern for evaluating base32 encodings is `[A-Z2-7]+=*`

Base32 uses A-Z and 2-7 for the encoding, and adds a padding character = to get a multiple of 8 characters. And according to @dwinston's [gist](https://gist.github.com/dwinston/083a1cb508bbff21d055e7613f3ac02f) the blade is a base32 encoding of any 64 bit integer. So we can change the blade portion of the above regex do you think?

sujaypatil96: I think we might want to standardize our naming of minting authorities, and further constrict our regex?

turbomam: > Hm, I don't think I'm understanding the purpose of the last grouping _.+?

That's for contig identifiers, which may be part of NMDC protein identification identifiers

From Paul Piehowski: nmdc:mga0wn63_1070_c1_4056_7928

(_.+)? would correspond to _1070_c1_4056_7928 in this case

turbomam: Maybe we should constrain that `.` any character wildcard

turbomam: > I think the regex pattern for evaluating base32 encodings is [A-Z2-7]+=*

OK, this requires more discussion, esp. re. receiving and emitting characters. I would like keep the pattern as simple as possible. If we need to accept lower case from an identifier minter, then we should have a canonical convert that is applied before inserting anything into a NMDC backend.

thttps://www.crockford.com/base32.html

turbomam:
> An example of a metaproteomic sample identifier with contig information would certainly help refine the regex if need be. CC: @scanon @pdpiehowski

I pasted an example protein identifier from @pdpiehowski above

turbomam:
> Although I'm not fully understanding the `id_usage.py` script in this PR? Is it's purpose meant to be a utility method? Or an example of sorts?

Good point. It's a quick & dirty, and I didn't provide any documentation. That script just determines which of the NMDC classes use the `id` slot, and if so, what the induced usage looks like.

src/schema/basic_slots.yaml
sujaypatil96: Proposed changes to regex:

* type code
  *  all lowercase characters, no digits
  * soft acceptance of minimum length of 3 characters, but no widely accepted
  * regex: `[a-z]{1,6}`
* shoulder
  * no need to stick to `{number}{characters}{number}` pattern for shoulder since we are using hyphens for separating components of the identifiers out
  * all lowercase between padding numbers
* blade
  * open ended length `[A-Za-z0-9]{1,}`

scanon: Just to give some concrete examples... 

A contig ID might look like...

```
nmdc:mgasm-00-wn63_1070_c1
```

and a gene/protein ID might look like...

```
nmdc:mgann-00-mn1c6_1070_c1_4056_7928
```

I think the hyphen's are fine but we should confirm this doesn't cause any tool problems.

Note that in this example the shoulder is different between these even though the gene ID is coming from that contig in this example.

I'll give some examples of a MAG bin as well.

scanon: Hmm.  The MAGs/bins are kind of tricky.   We have bin IDs.  Right now we don't really do much with these in the system.  So they are just bin.1, bin.2, etc.  So we should probably at a minimum, rename these to include the activity ID...  something like....

nmdc:mgmag-00-x012.1_bin_1

In the various output files, the workflow currently uses the original contig IDs.

So they currently look like...

```
>nmdc:mga0mn1c69_7_c1
```

So I'm not sure what to do with this.  I suppose it could be renamed as well.  So..

```
> nmdc:mgmag-00-x012.1_7_c1
```

This is a little weird because we would have to different types of suffixes on this ID.

sujaypatil96:
Direction from @cmungall: hold off on merging this PR since it will render all records in the NMDC Mongo invalid. Merge needs to be coordinated with workflows team.

Issue #509, opened 2022-11-11
combine all nmdc-schema modules EXCEPT mixs into a consolidated YAML file


turbomam:
work in this branch won't reflect changes in other current branches, like [issue-486-data-to-7-0](https://github.com/microbiomedata/nmdc-schema/tree/issue-486-data-to-7-0)

changes:
- quoting of schema's `description` ?
- ordering of elements
- loses `#` comments
- verbosity
  - prefix_prefix/prefix_reference
  - tag/value annotations

turbomam:
How to isolate the mixs module?

turbomam:
See also https://github.com/linkml/linkml/issues/1137

PR #523, opened 2022-11-29
NMDC schema with only 2 modules: nmdc.yaml and mixs.yaml

turbomam:
#509


turbomam:
I believe that this PR was indented to make it easier to find elements. We have since seen  that most or all schema contributors can use some kind of find-text-in-directory tool to determine which YAML module any schema element lives in.

PR #529, opened 2022-11-30
schema and test data changes for v7 migration

turbomam:
#527

Whitespace normalization

Converted many YAML `#` comments to LinkML `comments` or `notes`

- All modules import 
  - basic_slots
  - core

- All `slot_uri` of `rdf:type` removed from NMDC `type` slot

- Created `ControlledIdentifiedTermValue` class in core.yaml
  - And assigned to slots that had a `ControlledTermValue` range in mixs.yaml, as long as '{termLabel} {[termID]}'

```
  ControlledIdentifiedTermValue:
    description: A controlled term or class from an ontology, requiring the presence of term with an id
    comments:
      - To be used for slots like env_broad_scale
    is_a: ControlledTermValue
    slot_usage:
      term:
        required: true
```

- Set `external_database_identifiers`' range to `uriorcurie` in external_identifiers.yaml
  - Want the objects of these relations to appear as IRIs in RDF

- mixs.yaml: commented out several slot definitions when the slot shared a `slot_uri` with other slots, if that slot has not been used in mongodb records 

- nmdc.yaml: added several prefixes
  - Note the addition of the `bare` prefix to account for un-prefixed records in MongoDB
  - Note the assignment of the https://unknown.to.linter.org/ to several prefixes
- Commented out usages of mixs slots for `Biosample` if they were commented out in mixs.yaml
- Added and assigned several new identifier slots
- Assigned pattern: '^nmdc:' to Biosample (should also apply that pattern to other classes)
- Commented out all ...depth2 slots and usages
- Removed `identifier` slot. Replaced with `sample_name` in test data.
- renamed `gold_sample_identifiers` slot to `gold_biosample_identifiers`

Updated test data to pass against updated schema

### DIFF
--- a/src/schema/annotation.yaml
+++ b/src/schema/annotation.yaml
@@ -38,7 +38,6 @@ default_prefix: nmdc
 imports:
   - basic_slots
   - core
-  - linkml:types
   - prov
   - workflow_execution_activity
 
--- a/src/schema/annotation.yaml
+++ b/src/schema/annotation.yaml
@@ -13,6 +13,9 @@ default_curi_maps:
   - obo_context
   - idot_context
 
+notes:
+  - Removed slot_uri of rdf:type from type slot. Is OntologyClass an appropriate range?
+
 
 prefixes:
   COG: https://unknown.to.linter.org/
@@ -28,7 +31,6 @@ prefixes:
   biolink: https://w3id.org/biolink/vocab/
   linkml: https://w3id.org/linkml/
   nmdc: https://w3id.org/nmdc/
-  #  KEGG_ORTHOLOGY: http://unknown.to.linter.org/
   KEGG.ORTHOLOGY: http://identifiers.org/kegg.orthology/
 
 default_prefix: nmdc
@@ -51,13 +53,12 @@ classes:
     slot_usage:
       seqid:
         description: The ID of the landmark used to establish the coordinate system for the current feature.
-        range: string ## make this an object
+        range: string
         required: true
-      raw_type:
+      type:
         range: OntologyClass
         description: >-
           A type from the sequence ontology
-      #        slot_uri: rdf:type
       start:
         is_a: gff_coordinate
         description: The start of the feature in positive 1-based integer coordinates
@@ -74,7 +75,6 @@ classes:
         close_mappings:
           - biolink:end_interbase_coordinate
       strand:
-        ## TODO: enum
         description: >-
           The strand on which a feature is located. Has a value of '+' (sense strand or forward strand) or '-' (anti-sense strand or reverse strand).
         exact_mappings:
@@ -188,20 +188,22 @@ classes:
     id_prefixes:
       - CATH
       - EGGNOG
-      - KEGG.ORTHOLOGY ## KO number
+      - KEGG.ORTHOLOGY
       - PANTHER.FAMILY
       - PFAM
       - SUPFAM
       - TIGRFAM
     exact_mappings:
       - biolink:GeneFamily
+    notes:
+      - KEGG.ORTHOLOGY prefix is used for KO numbers
 
   FunctionalAnnotation:
     description: >-
       An assignment of a function term (e.g. reaction or pathway) that is executed by a gene product, or which the gene product plays an active role in.
       Functional annotations can be assigned manually by curators, or automatically in workflows. In the context of NMDC, all function annotation is performed
       automatically, typically using HMM or Blast type methods
-    comments:
+    notes:
       - "move id slot usage patterns to has_function slot usage?"
     see_also:
       - https://img.jgi.doe.gov/docs/functional-annotation.pdf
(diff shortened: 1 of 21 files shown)


src/schema/annotation.yaml
```diff
@@ -50,13 +53,12 @@ classes:
     slot_usage:
       seqid:
         description: The ID of the landmark used to establish the coordinate system for the current feature.
-        range: string ## make this an object
+        range: string
```
sujaypatil96, line 56: Question: is this comment requesting for `seqid` to have some other class enforced as its range? A more constrained class perhaps, with a `pattern`?

src/schema/bioscales.yaml
```diff
@@ -28,8 +31,10 @@ default_curi_maps:
 
 
 imports:
-  - linkml:types
+  - basic_slots
   - core
+  - linkml:types
```
sujaypatil96, line 36: Is an explicit `linkml:types` import necessary? I ask because `basic_slots` imports `linkml:types` already

src/schema/bioscales.yaml
```diff
@@ -121,7 +129,8 @@ slots:
     title: nitrate_nitrogen
     examples:
       - value: 0.29 mg/kg
-    #    source: https://www.ornl.gov/content/bio-scales-0
+    notes:
+      - see https://www.ornl.gov/content/bio-scales-0. could that be the "source"?
```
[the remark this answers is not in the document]
sujaypatil96: This comment is applicable to all notes that mention this source: https://www.ornl.gov/content/bio-scales-0

src/schema/core.yaml
```diff
@@ -417,7 +416,17 @@ classes:
...
+      - add fields for ontology, branch
+
+  ControlledIdentifiedTermValue:
+    description: A controlled term or class from an ontology, requiring the presence of term with an id
+    notes:
+      - To be used for slots like env_broad_scale
+    is_a: ControlledTermValue
+    slot_usage:
+      term:
+        required: true
```
sujaypatil96, line 429: This is a super useful addition. I had one quick question. Is there a reason why we can't enforce `required: true` on `ControlledTermValue`? 

Also it looks like there's additional requests for expanding what `ControlledTermValue` can capture, i.e., ontology, branch, etc.?

Just an approach I had in mind:

```
ControlledValue
  is_a: AttributeValue

ControlledTermValue:
  is_a: ControlledValue
  slots:
    - term
   slot_usage:
     term:
       required: true

ControlledOntologyValue:
  is_a: ControlledValue
  slots:
    - ontology # to be added?
   slot_usage:
     ontology:
       required: true
```

And so on.

Please feel free to override this, I haven't made too many changes to the schema, and this is just a thought I had.

src/schema/nmdc.yaml
```diff
@@ -871,11 +898,41 @@ enums:
...
+      - add is_a or mixin modeling
+      - what are the class of IMG records?!
+
+  emsl_biosample_identifiers:
+    title: EMSL Biosample Identifiers
+    description: >-
+      A list of identifiers for the biosample from the EMSL database.  This is
+      used to link the biosample, as modeled by NMDC, to the biosample in the planned EMSL NEXUS database.
+    multivalued: true
+    is_a: sample_identifiers
```
sujaypatil96, line 922: Should we rename `sample_identifiers` to `biosample_identifiers`? Very trivial, but since it looks like most of the slots that inherit this are different types of **biosample** identifiers, it can help clear some vagueness around the word _sample_?

src/schema/prov.yaml
```diff
@@ -20,7 +20,9 @@ default_curi_maps:
 
 imports:
   - basic_slots
+  - core
   - linkml:types
```
sujaypatil96, line 24: `linkml:types` necessary? It looks like core imports it too already

sujaypatil96:
Great job on the PR @turbomam. I've left a few amateur comments as part of a review.

src/schema/bioscales.yaml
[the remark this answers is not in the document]
turbomam: This note is for internal use only. It should not be shown on end-user documentation. `see_also` probably would be good semantics for the https://www.ornl.gov/content/bio-scales-0 value.

My motivation here was to migrate YAML # comments that wouldn't survive any form of conversion or regeneration into LinkML `notes`. I would prefer to merge this PR as is and save your suggestion for a new issue. I wouldn't want to change the `notes` to `see_also`s without reviewing what kind of attribution I was trying to make, and what the best slot is for that attribution.

src/schema/annotation.yaml
turbomam: I don't know. You have a good question, but that ## comment predates me. See point about converting YAML # comments into LinkML `notes` above.

src/schema/prov.yaml
turbomam: good point

src/schema/nmdc.yaml
turbomam: good point

src/schema/bioscales.yaml
turbomam: good point

turbomam:
> Great job on the PR @turbomam. I've left a few amateur comments as part of a review.

Your comments are very professional, not amateur at all.

Do you want to spend more time researching and discussing the comments that I didn't close, or can you live with making them new issues for the future?

src/schema/core.yaml
sujaypatil96: Oh right, that makes sense. To start with modeling classes for the MIxS environmental triad. And I completely missed thinking about the existing `ControlledTermValue` records from MongoDB which have `has_raw_value` and no `id`. Enforcing `id` on `ControlledTermValue` would invalidate the existing records from MongoDB, so a new `ControlledIdentifiedTermValue` class would definitely be the way to go, but we should think about the structure for these objects in later versions.

src/schema/annotation.yaml
sujaypatil96: Awesome, was curious if you knew. But no worries, documenting in a `todo` for now should be good enough.

sujaypatil96:
Oh absolutely, this PR is good to go as is. We can address any minor changes from some comments that I've made in future PRs.

sujaypatil96 (approved):

src/schema/bioscales.yaml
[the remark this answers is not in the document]
turbomam: I did change these `notes` to `see_also`s. Thanks for the suggestion.

Issue #532, opened 2022-12-01
Convert Anastasiya's SQL DDL to LinkML


turbomam:
@anastasiyaprymolenna and Natasha are building Metabolomics and Sample Processing data models directly in SQLAlchemy. For harmonization with the NMDC schema, we should examine a LinkML representation of their model as early as possible.

A SQLAlchemy representation of the NMDC schema has already been sent to @anastasiyaprymolenna.

turbomam:
source: assets/metabref_SQL_table_dump.txt

turbomam:
Hah, I overlooked the fact that the schema automator reads from a database, not a SQL DDL file. Any SQLAlchemy compatible database is OK and/or local SQLite files

```shell
schemauto import-sql path/to/my.db
```

turbomam:
It hasn't been possible for LBL etc staff to databases hosted on PNNL

PR #533, opened 2022-12-01
metabref SQL -> LinkML

turbomam:
#532 

**For your review... may not actually need to be merged.**

I overlooked the fact that the schema automator can import an SQL schema, but only from a database connection, not from a SQL DDL file.

@anastasiyaprymolenna thanks for providing the SQL DDL statements for metabref (assets/metabref_SQL_table_dump.txt). It looks like that file is table definitions only. I had to make up some sequences and types (assets/metabref_SQL_table_dump_modified.sql).

Then I used the modified SQL DDL file to create a local Postgres database and inferred a LinkML schema from that

```shell
make assets/metabref_SQL_table_dump.yaml
```

The database creation could be easily encapsulated in the Makefile by building an SQLite database instead of a Postgres database, but I didn't see a quick way to convert the Postgres dialect to the simpler SQLite dialect.

In the future, it might be easier for someone with access to the database within PNNL to run the simple inference command

```shell
schemauto import-sql postgresql+psycopg2://<USER>:<PW>@<HOST>:<PORT>/<DB>  > metabref_SQL_table_dump.yaml
```

I think the only requirements for that are `schema-automator` and Postgres drivers

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -335,3 +335,6 @@ from_mongo_all: from_mongo_cleanup validate_vs_3_2_0 validate_vs_current
 
 assets/metabref_SQL_table_dump.yaml: assets/metabref_SQL_table_dump.txt
 	$(RUN) schemauto import-sql postgresql+psycopg2://metab_user:metab_pass@localhost:5432/metab > $@
+
+target/nmdc_sqlalchemy.py: src/schema/nmdc.yaml
+	$(RUN) gen-sqla $< > $@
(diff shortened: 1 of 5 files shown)


turbomam:
The LinkML class names are appear as lower camel case

```yaml
class MassSpectrumObject(db.Model):

    __tablename__ = 'massSpectrumObject'
```

Issue #374, opened 2022-08-04
"history/extreme events" needs more context


pvangay:
Currently you can enter the date on which an extreme event occurred, but there is no where to explain what the actual extreme event was. ESS attendees noted that "history/extreme events" (which expects a date) should be accompanied by another term that provides more context on the extreme event itself.

This may require submitting a ticket to the GSC? Please discuss then address. @turbomam @cmungall

turbomam:
Good one. There are a lot of MIxS terms like this that inefficiently address the need to identify noteworthy events and the dates on which they occurred. I'll dig up some more examples and make some suggestions.

mslarae13:
https://github.com/microbiomedata/nmdc-schema/pull/545

mslarae13:
@ssarrafan  can you add this to the sprint that starts monday the 12th

mslarae13:
This is basically done. PR approved and merged. Need to add "todo". WIll do that in a different branch

turbomam:
Since the PR is merged, can we close this. Is there an issue for the addition of todos yet?

PR #545, opened 2022-12-08
Issue 374 update extreme events

mslarae13:
#374 
added extreme_event to slot usage for biosample.
Edited MIxS format requirments / syntax and provided an example


mslarae13:
Pending approval, submit a ticket to GSC

src/schema/nmdc.yaml
```diff
@@ -476,6 +474,24 @@ classes:
         required: false
       part_of:
         required: true
+      extreme_event:
```
mslarae13, line 477: in mixs.yaml there's also 
"
    from_schema: http://w3id.org/mixs/terms
    source: http://w3id.org/mixs/terms
    see_also:
      - https://github.com/microbiomedata/nmdc-schema/blob/issue-291-mixs-submod/util/rebuild_mixs_yaml.py
   "
@turbomam  should I have left that in there?

mslarae13:
Remove repeated information. Anything that is inherited from the parent slot & not changed in usage should not be re-listed. 

Don't edit 'aliases'

Update "range: text" to range: string

src/schema/nmdc.yaml
turbomam: The new content and your following of SOPs are really good. Here's to many more.

turbomam (approved):

Issue #553, opened 2022-12-15
report mixs package/term association counts from sheets-for-nmdc-submission-schema


turbomam:
https://docs.google.com/spreadsheets/d/1cMlPKgjZh-v21aMYCm9x1TxzE5BwGQptBQcvuaYAtC8/edit#gid=1742830620

mslarae13:
@turbomam can you expand on what this issue is trying to address?

PR #554, opened 2022-12-15
mixs package term list

turbomam:
- #553 

depends on manually downloaded https://docs.google.com/spreadsheets/d/1cMlPKgjZh-v21aMYCm9x1TxzE5BwGQptBQcvuaYAtC8/edit#gid=1742830620 -> assets/sheets-for-nmdc-submission-schema_import_slots_regardless.tsv

creates a two-column table illustrating which MIxS terms are paired with which environment-based NMDC DataHarmonizer

thus, indicates which MIxS packages are "supported" in the NMDC submission portal. The availability of an MIxS environment-based template doesn't mean that it has been curated.

there are also some templates derived from a MIxS environment, like soil_emsl, or templates whose relation to a MIxS environment is not clear from the name (bioscales is a modification of the soil template)


turbomam:
manually copied selected files to 
- #682

closing

PR #579, opened 2022-12-23
added empty well clarification

mslarae13:
@turbomam  please add validation for jgi metaG and metaT plate well locations.

https://github.com/microbiomedata/issues/issues/15

### DIFF
--- a/src/schema/portal/jgi_metagenomics.yaml
+++ b/src/schema/portal/jgi_metagenomics.yaml
@@ -127,11 +127,12 @@ slots:
     recommended: true
   dna_cont_well:
     name: dna_cont_well
-    title: DNA well number
+    title: DNA plate position
     comments:
-    - Required when 'plate' is selected for container type. Corner wells must be blank/empty (A1, A12, H1, H12).
-      For partial plates, fill by columns. Leave blank if the sample will be shipped
-      in a tube.
+    - Required when 'plate' is selected for container type.
+    - Leave blank if the sample will be shipped in a tube.
+    - JGI will not process samples in corner wells, so A1, A12, H1 and H12 will not pass validation.
+    - For partial plates, fill by columns, like B1-G1,A2-H2,A3-D3 (NOT A2-A11,B1-B8).
     examples:
     - value: B2
     from_schema: https://example.com/nmdc_dh
@@ -140,6 +141,7 @@ slots:
     slot_group: JGI-Metagenomics
     required: false
     recommended: true
+    pattern: ^(?!A1|A12|H1|H12)(([A-H][1-9])|([A-H]1[0-2]))$
   dna_container_id:
     name: dna_container_id
     title: DNA container label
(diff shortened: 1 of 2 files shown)


src/schema/portal/jgi_metagenomics.yaml
```diff
@@ -129,7 +129,7 @@ slots:
     name: dna_cont_well
     title: DNA well number
     comments:
-    - Required when 'plate' is selected for container type. Corner wells must be blank.
+    - Required when 'plate' is selected for container type. Corner wells must be blank/empty (A1, A12, H1, H12).
```
turbomam, line 132: I don't think we should say that the corner wells should be blank, but rather that JGI won't assay sample in corner wells. Therefore, samples with a `DNA well number` of A1, A12, H1, or H12 won't pass validation.

Also, I don't think 'DNA well number' is a good `title`, since the allowed values aren't numbers, but rather well identifiers or positions.

src/schema/portal/jgi_metagenomics.yaml
```diff
@@ -129,7 +129,7 @@ slots:
...
       For partial plates, fill by columns. Leave blank if the sample will be shipped
```
turbomam, line 133: It looks like there are two comments here. It looks like the hyphen `-` starting character wasn't used for the second comment.

turbomam: I will address the two comments above and add [the pattern from the issue](https://github.com/microbiomedata/issues/issues/15#issuecomment-1360480467)

turbomam (changes requested):

src/schema/portal/jgi_metagenomics.yaml
turbomam: You have picked up so much about using GH for collaboration so quickly. Unfortunately we have a few poorly documented standards, like branch names. Please use this pattern:

> `issue-<number>-<brief keywords>`

Generally don't use capital letters or underscores. By convention, `issue-15-JGI-plate-layout` would have been a better branch name.

You can see other historical branch names at https://github.com/microbiomedata/nmdc-schema/branches/all

ssarrafan:
moving to current sprint per Slack message from @turbomam

src/schema/portal/jgi_metagenomics.yaml
```diff
@@ -127,11 +127,12 @@ slots:
...
   dna_cont_well:
     name: dna_cont_well
-    title: DNA well number
+    title: DNA plate position
     comments:
-    - Required when 'plate' is selected for container type. Corner wells must be blank/empty (A1, A12, H1, H12).
-      For partial plates, fill by columns. Leave blank if the sample will be shipped
-      in a tube.
+    - Required when 'plate' is selected for container type.
+    - Leave blank if the sample will be shipped in a tube.
```
mslarae13, line 133: Will the sheet validate if this is left blank? Previously, we needed to put NA in this column because it's required. SO, tubes, put NA .. Plates, put the well number.

mslarae13: Mark: we can use linkML rules. This adjusts slot validation based on the input of another slot. Is this possible with DH? Can DH implement this rule?

turbomam (approved):

Issue #615, opened 2023-01-19
assign typcodes to all classes with the `id` slot


turbomam:
- from [20230112 nmdc typecodes](https://docs.google.com/spreadsheets/d/17m7rtf66ffMN2VwIfO3nZyJ0f-4BgXngMwI43XWsHb4/edit#gid=0)
- they will go into the `slot_usage` for the `id` of each class that has the `id` slot assigned
- add in `pattern` **and** `structured_pattern` as long as `structured_pattern`s aren't being converted to `pattern`s by gen-likml

- [x] future: remove typecodes from classes for which we will never mint an `id`, either because  
  - it's a top level grouping class (abstract or not) that is used only through it's  subclasses
  - it's an entity from an external database or ontology, which comes with its own `id`

turbomam:
- Should not require creating any new settings, because the typecodes are 1:1 unique to each class
- will require updating tests

PR #616, opened 2023-01-19
#615 typecodes, patterns and tests

turbomam:
- #615

### DIFF
--- a/src/schema/core.yaml
+++ b/src/schema/core.yaml
@@ -111,6 +111,13 @@ classes:
   AnalyticalSample:
     is_a: MaterialEntity
     title: Analytical Sample
+    slot_usage:
+      id:
+        pattern: '^nmdc:ansm'
+        structured_pattern:
+          syntax: "{id_nmdc_prefix}:ansm-{id_shoulder}-{id_blade}{id_version}{id_locus}"
+          interpolated: true
+          partial_match: false
 
   Site:
     is_a: MaterialEntity
--- a/src/schema/core.yaml
+++ b/src/schema/core.yaml
@@ -107,6 +107,13 @@ classes:
       - Physical entity
     is_a: NamedThing
     title: Material Entity
+    slot_usage:
+      id:
+        pattern: '^nmdc:mate'
+        structured_pattern:
+          syntax: "{id_nmdc_prefix}:mate-{id_shoulder}-{id_blade}{id_version}{id_locus}"
+          interpolated: true
+          partial_match: false
 
   AnalyticalSample:
     is_a: MaterialEntity
--- a/src/schema/core.yaml
+++ b/src/schema/core.yaml
@@ -108,13 +108,13 @@ classes:
       - Physical entity
     is_a: NamedThing
     title: Material Entity
-    slot_usage:
-      id:
-        pattern: '^nmdc:mate'
-        structured_pattern:
-          syntax: "{id_nmdc_prefix}:mate-{id_shoulder}-{id_blade}{id_version}{id_locus}"
-          interpolated: true
-          partial_match: false
+  #    slot_usage:
+  #      id:
+  #        pattern: '^nmdc:mate'
+  #        structured_pattern:
+  #          syntax: "{id_nmdc_prefix}:mate-{id_shoulder}-{id_blade}{id_version}{id_locus}"
+  #          interpolated: true
+  #          partial_match: false
 
   AnalyticalSample:
     is_a: MaterialEntity
--- a/src/schema/core.yaml
+++ b/src/schema/core.yaml
@@ -92,13 +92,13 @@ classes:
       - name
       - description
       - alternative_identifiers
-    slot_usage:
-      id:
-        structured_pattern:
-          todos:
-          syntax: "{id_nmdc_prefix}:{id_typecode_named_thing}-{id_shoulder}-{id_blade}{id_version}{id_locus}"
-          interpolated: true
-          partial_match: false
+#    slot_usage:
+#      id:
+#        structured_pattern:
+#          todos:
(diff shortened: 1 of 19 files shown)


test/data/biosample_test.json
```diff
@@ -105,7 +105,7 @@
       "sample_collection_site": "Lithgow State Coal Mine"
     },
     {
-      "id": "nmdc:e924072f-98b5-4f88-a796-a7ba1d8ddd92",
+      "id": "nmdc:bsm-bsm-e924072f-98b5-4f88-a796-a7ba1d8ddd92",
```
aclum, line 108: Why does this have two typecodes?

test/data/biosample_test.json
```diff
@@ -51,7 +51,7 @@
       "sample_collection_site": "Lithgow State Coal Mine"
     },
     {
-      "id": "nmdc:6057d02c-664c-41c9-8486-3624ca845747",
+      "id": "nmdc:bsm-bsm-6057d02c-664c-41c9-8486-3624ca845747",
```
aclum, line 54: Why does this have two typecodes?

src/schema/core.yaml
```diff
@@ -92,35 +92,57 @@ classes:
...
+  #      id:
+  #        structured_pattern:
+  #          todos:
+  #          syntax: "{id_nmdc_prefix}:{id_typecode_named_thing}-{id_shoulder}-{id_blade}{id_version}{id_locus}"
+  #          interpolated: true
+  #          partial_match: false
 
 
   MaterialEntity:
+    abstract: true
```
aclum, line 105: Does abstract: true mean that it doesn't need a typecode? That was true for the other examples but is not how MaterialEntity is written.

test/data/biosamples_to_sites.json
```diff
@@ -91,7 +91,7 @@
       }
     },
     {
-      "id": "nmdc:biosample_placeholder_4",
+      "id": "nmdc:bsm-biosample_placeholder_4",
       "name": "Root microbial communities from poplar common garden site in Clatskanie, Oregon, USA - BESC-13-CL2_39_29 endosphere",
       "description": "Root microbial communities from poplar common garden site in Clatskanie, Oregon, USA",
       "collected_from": "GOLD:BESC-13-CL2_39_29",
```
aclum, line 97: This placeholder didn't get updated. Is that intentional? A test that is expected to fail perhaps?

test/data/biosample_test.json
turbomam: Good catch. The only direct validation I've put in place yet is `^nmdc:{typecode}`.  I have added the entire `id` pattern for each class in a `strucutred_pattern`, but I don't believe the support for expanding them is complete in LinkML yet. I will check tomorrow.

Bottom line, the test `id` of "nmdc:bsm-bsm-e924072f-98b5-4f88-a796-a7ba1d8ddd92" passes because the plain `pattern` validation isn't anchored on the right yet.

I'll fix the example.

test/data/biosample_test.json
turbomam: See above.

test/data/biosamples_to_sites.json
```diff
@@ -121,7 +121,7 @@
       }
     },
     {
-      "id": "nmdc:study_placeholder",
+      "id": "nmdc:bsm-study_placeholder",
       "name": "Rhizosphere soil microbial communities from poplar common garden site in Clatskanie, Oregon, USA - BESC-13-CL2_39_29",
       "description": "Rhizosphere soil microbial communities from poplar common garden site in Clatskanie, Oregon, USA",
       "collected_from": "GOLD:BESC-13-CL2_39_29",
```
aclum, line 127: collected from should be updated to a NMDC identifier.

test/data/biosamples_to_sites.json
turbomam: No, it's the same un-anchored right pattern that you observed in some of the other `Biosample` test instances.

src/schema/core.yaml
turbomam: `abstract: true` means that the class shouldn't be instantiated directly. (Only subclasses of `MaterialEntity` should be instantiated. See https://microbiomedata.github.io/nmdc-schema/MaterialEntity/)

Since you can't directly instantiate `MaterialEntity`, you can't assign an `id`, so there's no sense in assigning a typecode.

I see that I **did** assign a typecode for   `MaterialEntity.slot_usage.id.pattern`. That makes things ore confusing, so I'll remove it now.

test/data/biosamples_to_sites.json
```diff
@@ -183,37 +183,38 @@
...
-        "GOLD:BESC-13-CL2_39_29"
+        "nmdc:frsite-10e5281b-504b-4f56-80fe-2e50d7d00521"
       ],
       "has_outputs": [
-        "GOLD:Gb0305834",
-        "GOLD:Gb0291693",
-        "GOLD:Gb0291583"
+        "nmdc:bsm-f34aa0dd-1fbf-4c46-aa16-713d41cfdc64",
+        "nmdc:bsm-32dea169-9135-4e9b-bce2-2286ac10ac0b"
       ]
```
aclum, line 218: This had 3 GOLD ids and now only has 2 nmdc biosample ids so missing one id.

src/schema/core.yaml
turbomam: I left `abstract: true` on `MaterialEntity` and removed the `id` pattern/typcode constraint, but this comment hasn't updated to "outdated"

test/data/biosamples_to_sites.json
turbomam: The `has_outputs` of the two `CollectingBiosamplesFromSite` instances account for all six of the Biosamples now.

aclum (approved):
All comments were addressed. Pull request is approved assuming it is ok to defer the issue of referential integrity.

Issue #638, opened 2023-01-23
assert rules for Biosamples sent to JGI in plates


turbomam:
- if the `dna_cont_type` is 'tube' then the'`dna_cont_well` must be null
- if the `dna_cont_type` is 'plate' then the'`dna_cont_well` must not be null (and therefore must match the pattern?)
- [ ] will it propagate to the submission schema?

turbomam:
for `rna_`... metatranscriptomics too

turbomam:
see also 
- https://github.com/linkml/linkml/issues/1259

turbomam:
rules asserted for the `dna_` slots

PR #647, opened 2023-01-24
Rules relating sample containers and plate (well) positions

turbomam:
- #638


turbomam:
Merged into 
- https://github.com/microbiomedata/nmdc-schema/pull/682

Closing

Issue #645, opened 2023-01-24
nmdc.schema.json should have corresponding version in "version" field


dwinston:
https://github.com/microbiomedata/nmdc-schema/blob/3449ae1e362868bae787ce6301c27650c02e26cc/jsonschema/nmdc.schema.json#L5572

should be "7.2.0" at present, right?

turbomam:
Yeah. Can you help me think of a way to automate that?

The automation would inject the version into [src/schema/nmdc.yaml](src/schema/nmdc.yaml)'s `version`

dwinston:
Sure. What's the "source of truth" of e.g. "7.2.0"?

If it's a git tag, here's an example from the nmdc-runtime repo: https://github.com/microbiomedata/nmdc-runtime/blob/456742f00d82e0c1bd03068779dc664847bd5b63/tasks.py

turbomam:
Here's the [MAINTAINERS.md](https://github.com/microbiomedata/nmdc-schema/blob/main/MAINTAINERS.md) file that Bill last updated a year ago.

I probably got out of the habit of following all of the steps.

turbomam:
`util/get_gh_tag.py` from 
- https://github.com/microbiomedata/nmdc-schema/pull/653/files

contained

```python
from subprocess import run


def publish():
    completed = run(
        "git describe --tags $(git rev-list --tags --max-count=1)",
        shell=True,
        capture_output=True,
        encoding="utf-8",
    )

    vtag = completed.stdout.strip()

    print(vtag)


publish()
```

but that PR was also contaminated with now-irrelevant Makefile changes. I'm closing that PR and branch but this is still a good issue.

turbomam:
https://github.com/microbiomedata/nmdc-schema/blob/387ed3be408a87f9da6df590e43953fc55665138/project/jsonschema/nmdc.schema.json#L8393

PR #653, opened 2023-01-25
Auto-populate version in nmdc.yaml, plus additional Makefile cleanup

turbomam:
- #645 
- 
---

- `make all` inserts the GH tag into nmdc.yaml with git describe and yq
- so requires yq
- haven't decided how this fits into the build and publish process


turbomam:
`util/get_gh_tag.py` might have some good ideas in it, but this PR is contaminated with now-irrelevant Makefile changes. I'm closing that PR and branch but this is still a good issue.

see also 
- #645

PR #682, opened 2023-01-27
Interleave previous cookiecutter with good legacy branch

turbomam:


sujaypatil96 (approved):
Reviewed PR with Mark, and it looks ready to go. Lovely work on the cleanup, it was long overdue. This is going to be so helpful in the long run.

PR #700, opened 2023-01-31
assert `host_taxid` slot on `Biosample` class

### LINKED ISSUE #699: Assert `host_taxid` slot from MIxS on `Biosample` class in NMDC Schema
Assert the `host_taxid` slot which is already present in `mixs.yaml` on the `Biosample` class in `nmdc.yaml` in order to facilitate the capture of host taxonomy id information from GOLD.

See https://gold.jgi.doe.gov/biosample?id=Gb0291745 for example. Click on the _host metadata_ tab, and you'll see an NCBI taxnonomy id for the host which is an integer value of `3689`.

sujaypatil96:
Fixes #699

### DIFF
--- a/src/data/valid/Biosample-exhasutive.yaml
+++ b/src/data/valid/Biosample-exhasutive.yaml
@@ -272,7 +272,8 @@ add_date: '2021-03-31'
 community: no_example_from_mixs
 habitat: unconstrained text
 host_name: snail is an example value, but unconstrained text is accepted at this point
-location: unconstrained text. should we even keep this slot? check if it ahs been
+host_taxid: 6521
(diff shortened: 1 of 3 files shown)


turbomam (changes requested):
Thanks @sujaypatil96. Good valid example, too!

You, @cmungall, @mslarae13 and I need to have a talk about 
- which files we can modify MIxS terms in
- whether (and where) we should replace structured ranges like TextValue with flat ranges like integer

turbomam:
I'm preparing some materials to guide the discussion I suggested.

turbomam:
closing vis 
- #733

PR #704, opened 2023-02-01
update file_name_patterns

Michal-Babins:
update file_name_patterns to reflect patterns found in gold that will then be converted to be nmdc schema compliant


nmdc_schema/nmdc_schema_generated.yaml
```diff
@@ -395,8 +395,10 @@ enums:
     permissible_values:
       Metagenome Raw Reads:
         text: Metagenome Raw Reads
-        todos:
-        - add annotations
+        annotations:
+          file_name_pattern:
+            tag: file_name_pattern
+            value: '/.+?(?=fastq)/fastq.gz'
```
sujaypatil96, line 401: There's a compact way of specifying annotations in the schema, like below:

```yaml
annotations:
  file_name_pattern: '/.+?(?=fastq)/fastq.gz'
```

Which is how I think you should represent it, simply so you're consistent with the way in which the annotations are specified on the rest of the permissible values in the enum, and not because it's incorrect. It's verbose, and as correct as the compact solution, but the point I'm trying to highlight is consistency.

sujaypatil96 (changes requested):
The place where you need to be making modeling changes to the schema is to the concerned file in [src/schema](https://github.com/microbiomedata/nmdc-schema/tree/main/src/schema). So in this case, you would be modifying the `nmdc.yaml` file.

Apologies for the lack of documentation, but should should never touch any of the files in the `nmdc_schema` folder within the repo since the files are mostly autogenerated and used while creating a deployment of the package.

nmdc_schema/nmdc_schema_generated.yaml
turbomam: Excellent point, @sujaypatil96 

I suggest we prioritize adding the content and therefore move forward with this merge. We can programmatically condense the modeling later. I'll make an issue

turbomam (approved):

nmdc_schema/nmdc_schema_generated.yaml
sujaypatil96: Sounds good @turbomam. But I think @Michal-Babins still needs to update the PR so the changes are in `src/schema/nmdc.yaml` and not the current file where the changes have been made.

turbomam:
> The place where you need to be making modeling changes to the schema is to the concerned file in [src/schema](https://github.com/microbiomedata/nmdc-schema/tree/main/src/schema). So in this case, you would be modifying the `nmdc.yaml` file.
> 
> Apologies for the lack of documentation, but should should never touch any of the files in the `nmdc_schema` folder within the repo since the files are mostly autogenerated and used while creating a deployment of the package.

Ahh, I didn't see this. It is required, so I'll fix it now.

turbomam (approved):

PR #706, opened 2023-02-02
Werkflow branch

turbomam:
This branch is already in use by the workflow team. There's no real rush to merge it in, but we should have a web-accessible diff


src/schema/core.yaml
```diff
@@ -199,8 +199,13 @@ classes:
     attributes:
```
turbomam, line 199: We're not using attributes anymore. These should be defined as slots.

src/schema/nmdc.yaml
```diff
@@ -638,9 +639,13 @@ enums:
       QC non-rRNA R1:
```
turbomam, line 639: this is pretty good in general. Ideally each permissible value would have a description and an annotation showing legal file name patterns

src/schema/workflow_execution_activity.yaml
```diff
@@ -177,6 +180,14 @@ classes:
       - unbinned_contig_num
       - mags_list
 
+  MetagenomeSequencingActivity:
+    description: Initial sequencing activity that precedes any analysis.  This activity has output(s) that are the raw sequencing data.
+    title: Metagenome sequencing activity
+    is_a: WorkflowExecutionActivity
+    in_subset:
+      - workflow subset
+    slots: []
```
turbomam, line 189: If this class doesn't have any slots, how would you use it?

test/data/metagenome_seequencing_activity.json
```diff
@@ -0,0 +1,22 @@
+{
```
turbomam, line 1: thanks for the example data. everyone should provide this!

turbomam (commented):
This is pretty good overall.

The top priority for merging this in is switching from `attributes` to `slots` in `src/schema/core.yaml`. I can do that for you if it's unclear. It won't effect the way data are expressed.

test/data/metagenome_seequencing_activity.json
turbomam: I can take care of the merge conflicts

Makefile
```diff
@@ -217,6 +217,7 @@ SCHEMA_TEST_EXAMPLES := \
 	nmdc_example_database \
```
turbomam, line 217: Actually, we're not using this test framework anymore. We have switched to linkml-run-examples. I will try to port this test data.

src/schema/core.yaml
```diff
@@ -215,6 +220,8 @@ classes:
         range: integer
       num_23s:
         range: integer
+      num_tRNA:
```
turbomam, line 223: num_tRNA is not a valid camel_case slot name and should appear as `num_t_rna` instead

src/schema/workflow_execution_activity.yaml
```diff
@@ -44,6 +44,7 @@ classes:
       - has_output
       - part_of
       - type # custom slot that specifies object type
```
turbomam, line 46: the `range` of `type` should be set to an enumeration.

Makefile
turbomam: I have back-merged main into this branch and am going to keep main's Makefile but will be sure to use your new test data

src/schema/workflow_execution_activity.yaml
turbomam: ah, it inherits whatever slots it needs from WorkflowExecutionActivity

PR #722, opened 2023-02-06
Fix automated deployment of docs

turbomam:


turbomam:
This includes some autogenerated files that aren't related to document generation of deployment. 

We previously ran `make deploy` and that seemed to run without error and included the static content in the GH pages

turbomam:
now waiting for GH action-based doc generation and deployment

Issue #796, opened 2023-02-22
do modifications with sheets_and_friends not yq


turbomam:

turbomam:
or better yet, leave MIxS slots as string, not `TextValues`.

turbomam:
- https://github.com/microbiomedata/nmdc-schema/issues/325

turbomam:
re-ordered yq range-altering commands

## TextValues

```shell
yq -i '.slots.agrochem_addition.range |= "TextValue"' $@
yq -i '.slots.air_temp_regm.range |= "TextValue"' $@
yq -i '.slots.antibiotic_regm.range |= "TextValue"' $@
yq -i '.slots.aromatics_pc.range |= "TextValue"' $@
yq -i '.slots.asphaltenes_pc.range |= "TextValue"' $@
yq -i '.slots.atmospheric_data.range |= "TextValue"' $@
yq -i '.slots.avg_occup.range |= "TextValue"' $@
yq -i '.slots.bathroom_count.range |= "TextValue"' $@
yq -i '.slots.bedroom_count.range |= "TextValue"' $@
yq -i '.slots.biocide_admin_method.range |= "TextValue"' $@
yq -i '.slots.biomass.range |= "TextValue"' $@
yq -i '.slots.chem_mutagen.range |= "TextValue"' $@
yq -i '.slots.cool_syst_id.range |= "TextValue"' $@
yq -i '.slots.diether_lipids.range |= "TextValue"' $@
yq -i '.slots.elevator.range |= "TextValue"' $@
yq -i '.slots.emulsions.range |= "TextValue"' $@
yq -i '.slots.escalator.range |= "TextValue"' $@
yq -i '.slots.ext_door.range |= "TextValue"' $@
yq -i '.slots.fertilizer_regm.range |= "TextValue"' $@
yq -i '.slots.floor_count.range |= "TextValue"' $@
yq -i '.slots.fungicide_regm.range |= "TextValue"' $@
yq -i '.slots.gaseous_environment.range |= "TextValue"' $@
yq -i '.slots.gaseous_substances.range |= "TextValue"' $@
yq -i '.slots.gravity.range |= "TextValue"' $@
yq -i '.slots.growth_hormone_regm.range |= "TextValue"' $@
yq -i '.slots.hall_count.range |= "TextValue"' $@
yq -i '.slots.hall_count.range |= "TextValue"' $@
yq -i '.slots.hcr_pressure.range |= "TextValue"' $@
[...]

PR #797, opened 2023-02-22
Use sheets_and_friends for MIxS slot modification, not yq

turbomam:
- #796


turbomam:
I consider `yq` to be a core tool in nmdc-schema now, but it does seem like I'm using it excessively, especially for converting MIxS `string` ranges to `TextValue` for nmdc-schema and then back to `string` in submission-schema

PR #802, opened 2023-02-28
Notebook instantiates a ReadQcAnalysisActivity with multiple base/read counts and validates

turbomam:
- #801

PR #803, opened 2023-02-28
move slot_usage to global, make slots explicit, for several classes

dwinston:
This results in a correct materialized JSON schema build.


dwinston:
incorporated into #802

Issue #807, opened 2023-03-01
several MIxS slots expected in submission schema missing from nmdc-schema


turbomam:
like `add_recov_method`

cc @pkalita-lbl

PR #808, opened 2023-03-01
added mixs slots req by submission schema to Biosample

turbomam:
- #807


turbomam:
might not be done yet

PR #861, opened 2023-04-12
Modify range of identifiers fields to be uriorcurie

### LINKED ISSUE #821: Always use CURIEs for identifier slots in schema
A general best practice is to always use CURIEs for identifier slots on classes in the schema. Refer: https://berkeleybop.github.io/best_practice/

Some slots in the schema that are currently modelled differently from best practice:
* img_identifiers: https://microbiomedata.github.io/nmdc-schema/img_identifiers/
* alternative_identifiers: https://microbiomedata.github.io/nmdc-schema/alternative_identifiers/
* assembly_identifiers: https://microbiomedata.github.io/nmdc-schema/assembly_identifiers/
* gold_identifiers: https://microbiomedata.github.io/nmdc-schema/gold_identifiers/
* insdc_identifiers: https://microbiomedata.github.io/nmdc-schema/insdc_identifiers/
* insdc_bioproject_identifiers: https://microbiomedata.github.io/nmdc-schema/insdc_bioproject_identifiers/
* mgnify_project_identifiers: https://microbiomedata.github.io/nmdc-schema/mgnify_project_identifiers/
 
Since they are identifiers, plural, they all need to be modelled as:
```
range: uriorcurie
multivalued: true
```

sujaypatil96:
Fixes #821

### DIFF
--- a/src/data/invalid/Biosample-invalid_identifiers.yaml
+++ b/src/data/invalid/Biosample-invalid_identifiers.yaml
@@ -1,0 +1,164 @@
+biosample_set:
+  - id: nmdc:bsm-99-dtTMNb
+    gold_biosample_identifiers:
+      - Gb0101224 # missing prefix
+    name: Lithgow State Coal Mine Calcium nutrients (early)
+    description: Bulk Aqueous phase filtered water
+    type: nmdc:Biosample
+    part_of:
+      - gold:Gs0110115
+    env_broad_scale:
+      has_raw_value: ENVO:00002030
+      term:
+        id: ENVO:00002030
+    env_local_scale:
+      has_raw_value: ENVO:00002169
+      term:
+        id: ENVO:00002169
+    env_medium:
+      has_raw_value: ENVO:00005792
+      term:
+        id: ENVO:00005792
+    geo_loc_name:
+      has_raw_value: Lithgow
+    lat_lon:
+      has_raw_value: "-33.460524 150.168149"
+      latitude: -33.460524
+      longitude: 150.168149
+    ecosystem: Environmental
+    ecosystem_category: Aquatic
+    ecosystem_type: Freshwater
+    ecosystem_subtype: Groundwater
+    specific_ecosystem: Coalbed water
+    add_date: 28-JUL-14 12.00.00.000000000 AM
+    community: microbial communities
+    habitat: Coalbed water
+    location: from the Lithgow State Coal Mine, New South Wales, Australia
+    mod_date: 26-AUG-16 01.50.27.000000000 PM
+    ncbi_taxonomy_name: coal metagenome
+    samp_taxon_id:
+      has_raw_value: coal metagenome [NCBITaxon:1260732]
+    sample_collection_site: Lithgow State Coal Mine
+  - id: nmdc:bsm-99-AtTUOs
+    gold_biosample_identifiers:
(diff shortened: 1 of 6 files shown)


src/data/valid/Biosample-exhasutive.yaml
```diff
@@ -25,8 +25,8 @@ embargoed: true
 collected_from: nmdc:unconstrained_site_identifier_string
 type: nmdc:Biosample. change this to require a class name or an enumeration
 img_identifiers:
-  - any string 1
-  - any string 2
+  - prefix:any string 1
```
turbomam, line 28: Let's use a prefix that's defined in the schema. I think there are some cases in which using an undefined prefix causes problems. Maybe converting data to RDF TTL?

turbomam:
Thank you. Can you please also add a an invalid example data file that is very obvious about the fact that id prefixes are missing? One of the id values could be 'unprefixed', and the file could have a name like src/data/invalid/Biosample-unprefixed-id.yaml

turbomam (changes requested):

src/data/invalid/Biosample-invalid_identifiers.yaml
turbomam, line 1: Thanks for the invalid file. I just can't tell what the invalid parts are. Can you please base your examples on the minimal example, and only introduce one invalid property per file? That obviously may require a large number of invalid files, with more indicative names.

turbomam (changes requested):

sujaypatil96:
Question to self: `alternative_identifiers` has a number of child slots like `gold_identifiers`, `imgnify_identifiers`, etc. If `required: true` and `multivalued: true` are being applied to `alternative_identifiers` already, then ideally the children should inherit those properties as well. Make sure that this is the case and remove redundant explicit assertions of those properties in the children slots.

turbomam:
@sujaypatil96 I have added consistent range, pattern and multivalued properties to the `external identifier` type and the `external_database_identifiers` parent slot in src/schema/external_identifiers.yaml as part of https://github.com/microbiomedata/nmdc-schema/pull/964. 

I believe we can close this PR and any related issue.

```yaml
types:

  external identifier:
    typeof: uriorcurie
    description: >-
      A CURIE representing an external identifier
    see_also:
      - https://microbiomedata.github.io/nmdc-schema/identifiers.html
    pattern: '^[a-zA-Z_][a-zA-Z0-9_]*:[a-zA-Z_][a-zA-Z0-9_]*$'

slots:

  external_database_identifiers:
    abstract: true
    description: Link to corresponding identifier in external database
    is_a: alternative_identifiers
    multivalued: true
    pattern: '^[a-zA-Z_][a-zA-Z0-9_]*:[a-zA-Z_][a-zA-Z0-9_]*$'
    range: external identifier
    comments:
      - The value of this field is always a registered CURIE
    notes:
      - "had tried ranges of external identifier and string"
    close_mappings:
      - skos:closeMatch
```

turbomam:
@sujaypatil96 do you want to leave this PR and it's corresponding branch open?

sujaypatil96:
We should be good to close this PR @turbomam. The fix that you have in #964 took care of all the external identifiers slots that I was trying to update individually. Your modeling fix that updated the range of `external_database_identifiers` to `external identifier` is a more elegant one 💯

PR #864, opened 2023-04-13
Force taxonomy id slots to be CURIEs

### LINKED ISSUE #823: `samp_taxon_id` and `host_taxid` need to be CURIEs
* `samp_taxon_id` slot in schema: https://microbiomedata.github.io/nmdc-schema/samp_taxon_id/
  * This slot needs to be modelled as a CURIE since it is an identifier
  * Need to be modelled in a way such that the text portion of the samp_taxon_id and CURIE portion of it don't need to be parsed out of the raw string
    * Example: Gut Metagenome [NCBI:txid749906]
    * Gut Metagenome needs to be in a slot on some class that can model samp_taxon_id
    * NCBI:txid749906 needs to be in a slot with a range of uriorcurie on the same class
* `host_taxid` slot in schema: https://microbiomedata.github.io/nmdc-schema/host_taxid/
  * Currently a string, needs to be uriorcurie

sujaypatil96:
Fixes #823


turbomam:
@turbomam's responsibility: show how to apply the changes in this PR's mixs/yaml via yq. That still may not be the best solution.

turbomam:
In other words:

for any slot whose value should indicate some organism, the value should be in the form of a CURIE, and the prefix should come from an agreed-upon namespace.

turbomam:
[The EBI OLS shows that about 70 ontologies](https://www.ebi.ac.uk/ols/ontologies/ncbitaxon/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCBITaxon_9606) import `NCBITaxon:9606` as the CURIE for _Homo sapiens_ , where the expansion of NCBITaxon is http://purl.obolibrary.org/obo/NCBITaxon_

turbomam:
The mixs.yaml in this PR asserts the right expansion for that prefix, but we shouldn't hand-edit mixs.yaml because it is regenerated on each build

So add to nmdc.yaml

turbomam:
relevant slots, both from `mixs.yaml` (so will require on-the-fly updating)

- host_taxid
- samp_taxon_id

Issue #870, opened 2023-04-28
Enum URI links are broken on NMDC Schema Documentation pages


aclum:
From https://microbiomedata.github.io/nmdc-schema/AnalysisTypeEnum/ if you click on the URI link it tries to resolve to https://microbiomedata.github.io/nmdc-schema/AnalysisTypeEnum/AnalysisTypeEnum with an error 404 - Not found

I get the same behavior with SoilHorizonEnum so I suspect this is a general issue with enumerations in this version of linkml documentation 
![Screenshot 2023-04-28 at 3 45 15 PM](https://user-images.githubusercontent.com/8196591/235265979-8da888ad-147a-48a3-835a-6979ac7ba4f5.png)

turbomam:
@sujaypatil96 it looks like we had some regression and the documentation was built with a old version of the templates. But I don't see any record of the documentation being generated any more recently than https://github.com/microbiomedata/nmdc-schema/actions/runs/4736919134

Having said that, I don't know if that would have any relation to broken enum pages.

Can you please look into this?

turbomam:
We were experimenting with using different versions of LInkML recently because the schema tests wouldn't run under 1.5.0+. But I really don't think I regenerated the documentation site. Do you know a good way to check?

This looks like the last commit to the gh-pages branch: https://github.com/microbiomedata/nmdc-schema/commit/3f60e3a943dc0c816aa07123cec46ffd85e394a8

turbomam:
I just ran `make all` and `make test` on https://github.com/microbiomedata/nmdc-schema/commit/3f60e3a943dc0c816aa07123cec46ffd85e394a8 with

- linkml = "1.4.11"
- linkml-runtime = "1.4.9"

then I did `make testdoc` and things looked fine locally

so finally i did `mkdocs gh-deploy` and the hosted enum pages look OK to me now. It just makes me nervous that I had to do all of that manually.

aclum:
@turbomam the enumeration links are still broken, should fixed or you only deployed this locally?

turbomam:
@aclum: @sujaypatil96 has helped me reread this issue more carefully. We will work on the un-resolvable URIs on enum pages as well as some other things that came up in our reassessment of the documentation in general.

turbomam:
@sujaypatil96 I don't think we have resolved this. Can you take it on as part of your ongoing documentation generation work?

turbomam:
[...]

PR #872, opened 2023-04-29
regenerated with linkml 1.4

turbomam:
and manually redeployed docs, but they still reverted to a old/broken state

- #870


turbomam:
I don't really know what we are missing out on by sticking to the latest linkml 1.4. At least one feature of 1.5.1 should be more efficient linkml-run-examples

PR #876, opened 2023-05-05
Use Makefile targets instead of individual poetry commands

sujaypatil96:
Use Makefile targets like `make gendoc` and `make mkd-gh-deploy` instead of individual poetry commands in GitHub Actions workflows.


sujaypatil96:
More in alignment with linkml-project-cookiecutter: https://github.com/linkml/linkml-project-cookiecutter/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/.github/workflows/deploy-docs.yaml#L29-L32

turbomam (approved):
Yes, this is a better practice

PR #929, opened 2023-06-01
Seq enum updates

aclum:
These new FileTypeEnums are needed for the sequencing workflows for NEON. We'd like this merged in so Michal can test the automation. Related to https://github.com/microbiomedata/nmdc_automation/issues/6


src/schema/nmdc.yaml
```diff
@@ -1495,9 +1495,22 @@ enums:
       Metagenome Bins:
         description: Metagenome bin contigs fasta
 
+      Metagenome Bins File:
```
turbomam: I want to make sure we are differentiating between these two similarly named enumerations:

- Metagenome Bins:
- Metagenome Bins File:

src/schema/nmdc.yaml
```diff
@@ -2281,24 +2312,6 @@ slots:
   mod_date:
     range: string
     description: The last date on which the database information was modified.
-  ecosystem_path_id:
```
turbomam, line 2495: why were all of these slots removed? I have added them back in

src/schema/nmdc.yaml
```diff
@@ -1625,9 +1625,21 @@ enums:
       Metagenome Bins:
```
turbomam, line 1625: Let's make sure the intention of these two enums is clear:
- Metagenome Bins:
- Metagenome Bins File:

turbomam (approved):
I made changes necessary for merging but further discussion may still be beneficial

src/schema/nmdc.yaml
Michal-Babins: Yeah, I wouldn't mind moving Metagenome Bins File to Metagenome Bins Compression File or Metagenome Bins Archive -- something that captures a collection of many metagenome bins

src/schema/nmdc.yaml
aclum: That was an error/typo.

src/schema/nmdc.yaml
aclum: I'll update this in a new branch.

PR #932, opened 2023-06-01
Update nmdc.yaml

aclum:
Updates to distinguish between Metagenome Bins (fasta file for an individual bin) and Metagenome Bins Compression File (zip file of all bins) and binning info file

cc @scanon


aclum:
Closing this pull request since there are new changes to the 20220601_seq_file_enums branch so we need a new pull request

PR #933, opened 2023-06-01
20220601 seq file enums

aclum:
Updates to distinguish between Metagenome Bins (fasta file for an individual bin) and Metagenome Bins Compression File (zip file of all bins), add binning info file, add new readqc stats enum

cc @scanon


Michal-Babins (approved):
Metagenome Bins File change to Metagenome Bins Compression File, 
QC Statistics name corrected to reflect Extended stats

turbomam (approved):

PR #958, opened 2023-06-06
Update nmdc.yaml

aclum:
Updates to FileTypeEnums b/c NEON has their raw sequencing data not interleaved.


src/schema/nmdc.yaml
```diff
@@ -1602,6 +1602,16 @@ enums:
       Metagenome Raw Reads:
         annotations:
           file_name_pattern: '/.+?(?=fastq)/fastq.gz'
+      Metagenome Raw Read 1:
```
turbomam, line 1605: - I will add clarification about R1 = fwd, R2 = rev
- I will update the pattern annotations like ^.+R1\.fastq(\.gz)?$
- in a separate issue, we should clarify how the pattern annotations are used and whether they need to be checked more globally

turbomam (changes requested):

PR #973, opened 2023-06-08
Create "NIST Bioeconomy Lexicon Scraper" notebook

eecavanna:
This branch contains a notebook (i.e. a Jupyter Notebook using Python and Markdown) that people can use to scrape the terms and definitions from the [NIST Bioeconomy Lexicon](https://www.nist.gov/bioscience/nist-bioeconomy-lexicon) web page and display them in CSV format.

You can view the rendered notebook [here](https://github.com/microbiomedata/nmdc-schema/blob/474547c19e6dccb2c13140b9ddc683a255d00ded/notebooks/nist_bioeconomy_lexicon_scraper.ipynb).


turbomam (approved):
Nice. I moved the imports and the assignment of sting variables to the top and added a TSV writer.

I have to review the Slack thread between you and @cmungall to see whether this should be wrapped as a Click CLI application and automated in a Makefile in the short term.

If not, we should find some opportunities for you to practice those skills in this repo, because I'm sure you'll have some good stuff to contribute.

Issue #927, opened 2023-05-31
EMSL study identifiers 


aclum:
Pending discussions with EMSL create an nmdc schema slot for EMSL study identifiers (ie https://www.emsl.pnnl.gov/project/60141)

There is a curie registered at bioregistry
https://bioregistry.io/emsl.project

Related to https://github.com/microbiomedata/nmdc-schema/issues/868

mslarae13:
From @lamccue 
Check with EMSL leadership (George and Ken) about the bioregistry curie for EMSL. The emsl.project prefix that was registered is no problem from their viewpoint.  We agreed that if NMDC identifies a desire/need to have more prefixes (for samples or other things), then the process for us will be to talk to George and Ken about it first - what the prefix is and how it will be used, etc. They can decide if it is something that they also want to incorporate into the NEXUS schema, and then we decide how to proceed and who does the registering.

mslarae13:
With that, we'd like to get EMSL projects linked in the data portal on the study landing pages at the "External Resources" -> "Additional data".

@aclum or @turbomam  or @naglepuff or @dwinston Can someone tell what the class-slot is I need to put these links under? I attempted to use the API but was unsuccessful in finding the slot.

I assume this will need to be added to mongo via change sheet?

mslarae13:
@naglepuff when we do get these added, they should appear here
![Screen Shot 2023-06-01 at 7 32 13 PM](https://github.com/microbiomedata/nmdc-schema/assets/44581429/9152d79b-069d-48b7-a38b-0175306684db)

With this logo
![EML_Logo-Light](https://github.com/microbiomedata/nmdc-schema/assets/44581429/6996950b-1cd6-41cd-8946-99e035330dde)

Unless someone has another opinion? (I will check with Lee Ann)

mslarae13:
@turbomam this emsl.identifiers curie should probably have it's own slot like GOLD, 

https://microbiomedata.github.io/nmdc-schema/study_identifiers/

I believe this is where these links comes from?

aclum:
That slot hasn't been used yet. That could be used, if we do at some point it should be updated so it inherits from alternative_identifiers and be enforced as a curie now that we have one. @turbomam  let us know if you want to do that now under this ticket, that seems cleaner since we haven't used the slot yet or if this ticket should be closed and we just need a nmdc-server repo ticket to get this information displayed.

mslarae13:
[...]

PR #975, opened 2023-06-09
emsl_project_dois slot

turbomam:
- #927


turbomam:
I changed to slot name to `emsl_project_dois` (plural) because it is multivalued aka takes a list as its object

turbomam:
I wrapped the example data files in `Database`.`study_set`, because we have better test coverage over `Database`s

PR #991, opened 2023-06-13
Remove dashes from module name

### LINKED ISSUE #990: Error while trying to install latest release of schema package
When I try to do a simple `pip install nmdc-schema --upgrade` to use the latest version/release of the nmdc-schema, I get the following error:

```python
ValueError: ("EntryPoint must be in 'name=module:attrs [extras]' format", 'get_neon-nlcd-envo-mapping=nmdc_schema.neon-nlcd-envo-mapping:generate_nlcd_envo_mappings')
```

The reason for that is this particular line in the `pyproject.toml` file: https://github.com/microbiomedata/nmdc-schema/blob/main/pyproject.toml#L101

The nit is that we cannot have `-` in module names. it is okay for the _entry point_ name itself to have dashes, but the module name cannot have it.

So it would need to be rewritten like this:

```
get_neon-nlcd-envo-mapping = "nmdc_schema.neon_nlcd_envo_mapping:generate_nlcd_envo_mappings"
```

with the [neon-nlcd-envo-mapping.py](https://github.com/microbiomedata/nmdc-schema/blob/main/nmdc_schema/neon-nlcd-envo-mapping.py) file needing to be renamed as well.

sujaypatil96:
Fixes #990


turbomam (approved):
thanks. I get really confused about when to use `-` or `_`. You may have to remind me some more 😞

PR #1003, opened 2023-06-15
Corrections to NEON NMDC mappings TSV file

sujaypatil96:
Some changes made to the NEON NMDC mappings TSV file:
* Added _data_product_ and _static_value_ columns to TSV file
* MIXS ENVO fields _env_broad_scale_ and _env_medium_ have static field mappings to certain ENVO terms. _env_local_scale_ will be determined at runtime
* Use dot notation, see changed TSV file for examples, to compactly represent what columns map to slots in the nested class structure for certain object types in NMDC

CC: @brynnz22 @bmeluch

### DIFF
--- a/assets/misc/neon_nmdc_term_mappings.tsv
+++ b/assets/misc/neon_nmdc_term_mappings.tsv
@@ -139,7 +139,7 @@ DP1.10086.001	sls_metagenomicsPooling	collectDate	exact_mapping	Pooling	end_date
 DP1.10086.001	sls_metagenomicsPooling	eventID					
 DP1.10086.001	sls_metagenomicsPooling	samplingProtocolVersion	close_mapping	Pooling	protocol_link.url		inherited from sls_soilCoreCollection
 DP1.10086.001	sls_metagenomicsPooling	genomicsPooledIDList	close_mapping	Pooling	has_input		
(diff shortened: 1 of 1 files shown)


turbomam:
Good work everybody.

I may not be able to really check this until Tuesday afternoon. If if delivers the behavior you want in the conversion code, you can go ahead and merge it @sujaypatil96 .

@brynnz22 and @bmeluch , looking at the `Files changed` tab above may give you a sense of why TSV has limitations for this kind of collaborative effort. Presumably @sujaypatil96 only changed a few values in the TSV file, but its just about impossible to tell what was changed in `Files changed`. That _could_ have been be exacerbated by changing the sort order of the rows. 

The migration from Google sheets to TSV in GitHub was great. This is just a further opportunity to think about how the same knowledge could be expressed directly in the schema, where incremental changes would be more prominent in the `Files changed`

turbomam (approved):

brynnz22:
@sujaypatil96 @bmeluch and I noticed that you changed the mappings of the rawDataFile NEON fields to the `OmicsProcessing` class (instead of the `DataObject` class. These slots do not appear to be in the `OmicsProcessing` class, but to keep things consistent for you, we did the same for the Surface Water samples.

bmeluch:
PCR primer sequences have no direct source in the NEON tables. Users would have to look up the documentation for the Illumina adapter kit used. This may not be worth trying to capture unless PCR primer sequences are necessary for something NMDC internal that I don't know about.

PR #1020, opened 2023-06-28
delete old mapping neon_nmdc_term_mappings.tsv file

brynnz22:
Deleted neon_term_mappings.tsv file. neon_term_mapping.tsv (no "s") should be kept.


brynnz22:
@sujaypatil96 once this is merged, there should only be one file.

turbomam (approved):
great housekeeping

PR #1022, opened 2023-06-28
edit neon mapping tsv units and add nmdc unit column

brynnz22:
Edited the neon_mapping,tsv to create shorthand notation for units in the `neon_units` column. Also created a new column called `nmdc_units `to specify if there is a unit change (e.g. for `depth` from cm to m)


turbomam (approved):
Thanks. I haven't attempted to review that changes in the TSV file. This is really making me think that we should be using a more diff-friendly format like YAML.

any thoughts @sujaypatil96 ?

PR #1024, opened 2023-06-29
remove rawDataFiles table for soil, benthic, and surfacewater mappings

brynnz22:
I removed the rawDataFiles tables for the soil, surface water, and benthic metadata since we will be using a different file for these mappings. @sujaypatil96


sujaypatil96 (approved):

turbomam (commented):
approved and merged. Thanks @brynnz22  and @sujaypatil96

Issue #1034, opened 2023-07-05
Map NEON's nucliecAcidQuantMethod for SurfaceWater and Benthic 


brynnz22:
We missed mapping the SurfaceWater and Benthic NEON `nucleicAcidQuantMethod` field to nmdc's `nucleic_acid_concentration.instrument_name` slot.

PR #1036, opened 2023-07-05
update nucleicAcidQuantMethod mapping

brynnz22:
I updated the NEON tsv to include the surface water and benthic `nuclieicAcidQuantMethod` mappings. We missed them, but included them for soil. @sujaypatil96

### DIFF
--- a/assets/misc/neon_nmdc_term_mapping.tsv
+++ b/assets/misc/neon_nmdc_term_mapping.tsv
@@ -424,7 +424,7 @@ DP1.20281.00	mms_swMetagenomeDnaExtraction	sampleMaterial							Captured from fi
 DP1.20281.00	mms_swMetagenomeDnaExtraction	sampleMass	exact_mapping	Extraction	sample_mass.has_numeric_value		g		
 DP1.20281.00	mms_swMetagenomeDnaExtraction	samplePercent							
 DP1.20281.00	mms_swMetagenomeDnaExtraction	nucleicAcidConcentration	broad_mapping	ProcessedSample	nucleic_acid_concentration.has_numeric_value		ng/uL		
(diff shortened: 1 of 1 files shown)


turbomam (changes requested):
I think the new-ish? `InstrumentValue` had made things confusing. 

I don't understand this mapping:

nucleicAcidQuantMethod | exact_mapping | ProcessedSample | nucleic_acid_concentration.instrument_name

We should talk about this some more

brynnz22:
Its kind of a misleading field name. Its called `nucleicAcidQuantMethod`, but the input is an instrument name, Quantus. See @aclum comment here: https://github.com/microbiomedata/issues/issues/325

aclum:
@brynnz22 @turbomam  see lines 14-17 of https://github.com/microbiomedata/nmdc-schema/blob/main/src/data/valid/Database-processed_sample-extract-exhaustive.yaml for how these slots would be populated.
How Brynn has done the mapping matches our example data.

turbomam:
@sujaypatil96 is it OK if we merge this now? or should we assume that your version of assets/misc/neon_nmdc_term_mapping.tsv has diverged from this and you prefer to keep your current version?

turbomam:
this was requested by @aclum but we forgot about it for the past 5 weeks!

aclum:
It would be good to get this merged so the surface water and benthic samples, which haven't been ingested yet, can have this metadata.

turbomam (approved):

brynnz22:
related to issue #1034

turbomam:
@sujaypatil96 we are merging this based on your input that your NEON ETL code doesn't directly act on these NEON TSV mapping files. You just use them as inspiration for your coding, right?

PR #1039, opened 2023-07-06
retire doi, created dois and subproperties

brynnz22:
- https://github.com/microbiomedata/issues/issues/338

### DIFF
--- a/nmdc_schema/get_study_doi_report.py
+++ b/nmdc_schema/get_study_doi_report.py
@@ -1,0 +1,67 @@
+from pymongo import MongoClient
+import click
+import click_log
+from dotenv import load_dotenv
+import logging
+import os
+import pandas as pd
+
+logger = logging.getLogger(__name__)
+click_log.basic_config(logger)
+
+def access_database(env_file, mongo_db_name, mongo_host, mongo_port, admin_db):
+    # Load MongoDB credentials from .env file
+    load_dotenv(env_file)
+    logger.info(f"loaded {env_file}")
+    mongo_pw = os.getenv('SOURCE_MONGO_PASS')
+    mongo_user = os.getenv('SOURCE_MONGO_USER')
+    logger.info(f"{mongo_user = }")
+
+    client = MongoClient(mongo_host,
+                         port=mongo_port,
+                         username=mongo_user,
+                         password=mongo_pw,
+                         authSource=admin_db,
+                         authMechanism='SCRAM-SHA-256',  # todo should be an option
+                         directConnection=True
+                         )
+
+    db = client[mongo_db_name]
+
+    study_collection = db['study_set']
+    result = study_collection.aggregate([
+  { "$group": { "_id": { "name": "$name", "id": "$id", "doi": "$doi.has_raw_value" } 
+            } 
+   }
+])
+
+
+    return result
+
+@click.command()
+@click_log.simple_verbosity_option(logger)
+@click.option('--env-file', type=click.Path(), default='local/.env', help='Path to .env file')
+@click.option('--mongo-db-name', default="nmdc", help='MongoDB database name')
+@click.option('--mongo-host', default="localhost", help='MongoDB host name/address')
+@click.option('--mongo-port', default=27777, help='MongoDB port')
+@click.option('--admin-db', default="admin", help='MongoDB authentication source')
+
+
+def get_dois(env_file, mongo_db_name, mongo_host, mongo_port, admin_db):
+
+    doi_data = access_database(env_file, mongo_db_name, mongo_host, mongo_port, admin_db)
+
+    data_list = []
+    count = 0
+    for data in doi_data:
+        count+=1
+        study = data['_id']
+        df = pd.DataFrame(study, index=[count,])
+        data_list.append(df)
+        
+    df = pd.concat(data_list)
+    df.to_csv('doi_report.tsv', sep='\t', index=False)
+    
+  
+if __name__ == '__main__':
+    get_dois()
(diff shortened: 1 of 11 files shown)


turbomam:
Thanks @brynnz22. This is a helpful differentiation of different kinds of DOIs.

It looks like one or both of us did some global reformatting of the nmdc.yaml file. I think it looks better now, but we should both keep in mind that combining reformatting and content changes makes it harder for reviewers to focus on just the content changes.

Have you thought about why this PR is failing the checks?

brynnz22:
Thanks @turbomam Yes, it looks like because I changed the name from `doi` to `dois`, that it does not recognize the old singular slot. We also made the range a `uriorcurie` which no longer uses the `has_raw_value` slot.

brynnz22:
@turbomam I added `pattern` to the new `dois` slot and I checked it against a new valid study-set example data. The old study-set data examples no longer validate because of the old `doi` slot. Should I remove these? Or keep these?

brynnz22:
@turbomam I edited the nmdc schema (`nmdc.yaml` file) to change the old slot `doi` to have subslots of `award_dois`, `dataset_dois`, and `publication_dois` and change the parent slot `doi` to be called `dois` because it is now multivalued. It also has a range of `uriorcurie` , so the `has_raw_value` slot will no longer be used for the new `dois` slot.
- FYI @naglepuff for the study pages, the award dois will be taken from the new `award_dois` slot and it will be multivalued.

I included a python function in the `mongodb_direct_to_nmdc_Database_file.py` to do the database migration of changing the current study dois to use the slot `award_dois`, make it into a list, and change the current doi values to be curies. All of the current dois in the database are award dois. @turbomam mentioned someone will take the output json file in load into production? I am not sure what comes next for this.

There is a new data file for study_set which will validate with the updated schema (`Database-study-set-with-dois.yaml`)

Finally, I included a python script `get_study_doi_report.py` and an addition to the `project.Makefile` which will output a tsv with a list of the current dois at the request of @turbomam. I'm not sure if this is what you had in mind, but please let me know any adjustments I can make.

FYI @pkalita-lbl you mentioned updating the submission-portal-toMongo pipeline. 

@eecavanna @dwinston

brynnz22:
@mslarae13 @lamccue

mslarae13:
@turbomam  @naglepuff 

Can you please let me know when this PR goes through & when all the necessary version updating is live so I can put in a change sheet to update the award DOIs for some projects?

Additionally, @naglepuff will these multiple values be visible on the data portal? Or do we need to make any changes ?

naglepuff:
@mslarae13 this will require data portal changes

mslarae13:
> 



> @mslarae13 this will require data portal changes

@naglepuff yes, which I think we will have something from Faiza too. We can discuss Tuesday.
Will your hide and seek stuff be done?

turbomam:
I have updated all of the `doi`s in `src/data/valid/Database-nmdc-example.yaml` to `award_dois`

@brynnz22 could you please confirm that they are all indeed `award_dois`? I like the definitions we wrote for the new `..._dois`, but I think we should add examples or some commentary about how one knows whether a DOI is 

      - award_dois
      - dataset_dois
      - publication_dois

turbomam:
We also haven't normalized `emsl_project_dois`. It's also applicable to `Study`and inherits `multivalued: true` from

`is_a: study_identifiers` -> `is_a: external_database_identifiers`

Whereas the three new `..._dois` you created are `s_a: dois`

turbomam:
Found another un-updated example data file: 

`src/data/valid/Database-study-set-with-gnps-id.yaml::$.study_set[0]: Additional properties are not allowed ('doi' was unexpected)`

I'll change that to an `award_dois` too. @brynnz22 please confirm.

turbomam:
@brynnz22 please see 
- https://github.com/microbiomedata/nmdc-schema/issues/1046

turbomam:
@brynnz22 I am really enjoying this review of your contribution to the schema and the migration script. I apologize that I didn't do it sooner.

I am getting the following output from the

```shell
poetry run linkml-validate \
  --schema nmdc_schema/nmdc_schema_accepting_legacy_ids.yaml local/selected_mongodb_contents.json
```

phase of `make dump-validate-report-mongodb`

> ✗ 'doi:10.25585/1487763' is not of type 'array' in $.study_set[0].award_dois
> ✗ 'doi:10.25585/1487765' is not of type 'array' in $.study_set[2].award_dois
> ✗ 'doi:10.25585/1488096' is not of type 'array' in $.study_set[9].award_dois
> ✗ 'doi:10.25585/1488099' is not of type 'array' in $.study_set[1].award_dois
> ✗ 'doi:10.25585/1488160' is not of type 'array' in $.study_set[4].award_dois
> ✗ 'doi:10.25585/1488224' is not of type 'array' in $.study_set[8].award_dois
> ✗ 'doi:10.46936/10.25585/60000017' is not of type 'array' in $.study_set[11].award_dois
> ✗ 'doi:10.46936/10.25585/60000762' is not of type 'array' in $.study_set[10].award_dois
> ✗ 'doi:10.46936/10.25585/60001061' is not of type 'array' in $.study_set[5].award_dois
> ✗ 'doi:10.46936/10.25585/60001198' is not of type 'array' in $.study_set[6].award_dois
> ✗ 'doi:10.46936/10.25585/60001289' is not of type 'array' in $.study_set[7].award_dois
> ✗ 'doi:10.46936/intm.proj.2021.60141/60000423' is not of type 'array' in $.study_set[12].award_dois
> ✗ Additional properties are not allowed ('jgi_portal_study_identifiers' was unexpected) in $.study_set[11]

I don't think the complaint about `jgi_portal_study_identifiers` comes from this PR ser se, but the other ones do. I'm going to work on that now. If you see this in the mean time, do you understand what the problem and the solution are?

turbomam:
My last question was Socratic. This one's for real: does it surprise you that not all of the Studies have `award_dois`?

brynnz22:
Yes, I believe so! I forgot to make the slots a list, I can fix now!

brynnz22:
@turbomam I fixed the function in `mongodb_direct_to_nmdc_Database_file.py` to be multivalued. Does it work now? 

Also, yes it does surprise me that not all of the Studies have `award_dois`.

brynnz22:
@turbomam I added examples for the rest of the dois except for `emsl_project_dois` as I am meeting with Montana and Bea later this afternoon to get clarification on this. @mslarae13 is double checking that all of the current dois are indeed award dois.

Issue #1041, opened 2023-07-11
`jgi_portal_identifiers` isn't attached to anything; make a `jgi_portal_study_identifiers` and mix in `jgi_portal_identifiers`


turbomam:
Starting with

```yaml
  jgi_portal_identifiers:
    mixin: true
    description: identifiers for entities according to JGI Portal
```

add

```yaml
  jgi_portal_study_identifiers:
    is_a: study_identifiers
    mixins:
      - jgi_portal_identifiers
```

And associate the new `jgi_portal_study_identifiers` with `Study`
cc @aclum

PR #1043, opened 2023-07-11
added jgi_portal_study_identifiers slot and example data

turbomam:
- #1041


aclum (commented):
Only suggestion is to make the pattern match what bioregistry has so \d+$ instead of ^jgi.proposal:[a-zA-Z0-9_][a-zA-Z0-9_\-\/\.,]*$

aclum (approved):

PR #1050, opened 2023-07-27
clairfy Study description to not include links

brynnz22:
Addresses the [Study Refresh issue ](https://github.com/microbiomedata/issues/issues/340). I updated the `Study` `description` slot to specify that links should not be included. I added `comments` to hint at which slots can be used for links. I also added `description` under the list of `slots` for `Study` as it was missing.


brynnz22:
@aclum @mslarae13 @pkalita-lbl, @turbomam and I are curious about when to merge and release schema updates, like this one, that do not require a migration?

aclum:
There is no restriction on merges that don't require a data migration. We'd talked about having a more official release cycle but I'm not sure where those discussions are at. It is okay w/me to merge this in.

turbomam (approved):
Good changes. Let's continue the discussion about some shared reformatting rules.

PR #1055, opened 2023-08-01
remove abstract slot globally from schema

brynnz22:
I removed the `abstract` slot globally from the schema. It was only be used in the `Study` class and was not used at all in MongoDb. So it does not get confused with the `description` slot in the future, where it is interchangeable, we are removing it.


brynnz22:
@turbomam I added an invalid example data file that includes the `abstract` slot. I think this is good to merge now too.

turbomam (approved):

PR #1058, opened 2023-08-02
update funding_sources to be more specific

brynnz22:
For [microbiomedata/issues#339](https://github.com/microbiomedata/issues/issues/339), I added a description to funding_sources, comments that specify exactly what we would like to see in the value, along with two examples that [@lamccue](https://github.com/lamccue) [@mslarae13](https://github.com/mslarae13) specified were good examples. I also add closing_mappings with a value of the funding agency ncit ontology term that I primarily got the description from. Finally, I added a pattern that limits the value to 150 characters. Let me know if this is too strict.


brynnz22:
@turbomam I added an example invalid file with a `funding_sources` value that is more than 150 characters. I think this is ready for merging.

Issue #1060, opened 2023-08-04
Add documentation for the legacy MongoDB contents migrator


turbomam:

turbomam:
Well documented in
- https://github.com/microbiomedata/nmdc-schema/tree/main/nmdc_schema/migrators

PR #1061, opened 2023-08-04
first steps in migration notes

turbomam:
- #1060


Shalsh23 (approved):

Shalsh23:
I have tried the currently listed first four steps, and they are working for me. I was able to setup a tunnel via dtn01.nersc.gov node to the production MongoDB server.

PR #1066, opened 2023-08-15
Assert `type` slot on NEON classes

### LINKED ISSUE #1027: Add type on class `PlannedProcess`
Children of class [PlannedProcess](https://microbiomedata.github.io/nmdc-schema/PlannedProcess/) (Pooling, Extraction, LibraryPreparation) are being used extensively in the NEON metadata ingest pipeline to create instances of above mentioned processes.

It is useful to know the _type_ of instances being created, like how we indicate for _Biosamples_ that they are of type `nmdc:Biosample`. Similarly, it is useful to annotate instances of PlannedProcess objects (and their children) with type as well. For example, Pooling process instances will have type `nmdc:Pooling`, Extraction process instances will be of type `nmdc:Extraction` and so forth. This is useful redundancy to have, because it helps while writing queries against the database.

So we are requesting a modeling change where you assert the slot [type](https://microbiomedata.github.io/nmdc-schema/type/) on PlannedProcess, which will carry over into its child classes as well.

sujaypatil96:
Fixes #1027

### DIFF
--- a/src/schema/core.yaml
+++ b/src/schema/core.yaml
@@ -101,12 +101,15 @@ classes:
       - dna_concentration
       - external_database_identifiers
       - nucleic_acid_concentration
+      - type
     slot_usage:
       id:
         required: true
         structured_pattern:
           syntax: "{id_nmdc_prefix}:procsm-{id_shoulder}-{id_blade}{id_version}{id_locus}"
           interpolated: true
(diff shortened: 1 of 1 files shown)


sujaypatil96:
Closing this - incorrect syntax

Issue #1030, opened 2023-07-03
Add slot type to nmdc planned process classes


aclum:
@sujaypatil96  suggested at the metadata meeting last week that we add slot `type` to the new PlanndedProcess classes and ProcessedSample
-Pooling
-Extraction
-LibraryPreparation
-ProcessedSample

This will make the classes more consistent and will make it easier for API users to understand what class a json record is for.

cc @sujaypatil96 @mslarae13

PR #1067, opened 2023-08-15
`type` slot assertion on planned NEON process classes and processed sample

### LINKED ISSUE #1027: Add type on class `PlannedProcess`

sujaypatil96:
Fixes #1027 and #1030

### DIFF
--- a/src/schema/core.yaml
+++ b/src/schema/core.yaml
@@ -101,6 +101,7 @@ classes:
       - dna_concentration
       - external_database_identifiers
       - nucleic_acid_concentration
+      - type
     slot_usage:
       id:
         required: true
@@ -142,6 +143,7 @@ classes:
       - protocol_link
       - start_date
       - end_date
+      - type
       # - participating_agent
 
   OntologyClass:


sujaypatil96:
Agenda item for 8/16 call with @turbomam:

- decide how we want to constrain the _range_ of type slot on the NEON PlannedProcess and ProcessedSample classes.

aclum:
@turbomam  can we get this reviewed and merged? @sujaypatil96  is there corresponding migration code?

turbomam:
This can be closed, but we need to refactor "type" slots ASAP. I have an issue for this but haven't found it yet.

aclum:
@turbomam  are you saying you'd like to close instead of merge this pull request where @sujaypatil96  has addressed some of the type slot issues?

turbomam:
I was hoping to 
- prioritize applying `type`, as an inherited type-designating slot (whose range would be identifiers of nmdc-schema classes) to root classes
- remove all other NMDC slots with "type" in their name. (We would keep any MIxS slots that ahve "type" in the name)

aclum:
If we can address your first bullet 'prioritize applying type, as an inherited [type-designating slot](https://linkml.io/linkml/schemas/type-designators.html) (whose range would be identifiers of nmdc-schema classes) to root classes' with the december refactor we should do that in favor on merging this in. let's leave this open until we decided @lamccue @mslarae13

mslarae13:
I'll defer to @lamccue for the prioritization & if there's time to target `type` for December. 

As for getting it done, @turbomam can we get a list of all the slots that have "type" in the name/title that shouldn't so we can evaluate how many slots we need to change?

lamccue:
I don't have a good understanding of the level of effort required, or the impact the change would have within the schema, of the first point, i.e.,: 

1. prioritize applying type, as an inherited [type-designating slot](https://linkml.io/linkml/schemas/type-designators.html) (whose range would be identifiers of nmdc-schema classes) to root classes

aclum:
Just a note that we can separate concerns here wrt to what gets done during the December refactor, we can add slot 'type' on all classes and set a range to be a class independent of changing the name of slots which currently have the word type in them (ie data_object_type).

aclum:
Im going to close this PR since @turbomam took care of what is addressed in this PR globally in the berkeley-schema-fy24 fork. I unlinked the nmdc-schema ticket so that would remain open.

Issue #1069, opened 2023-08-15
SPARQL queries over NMDC metadata in RDF


turbomam:
May require
- dumping from MongoDB (via direct connection or API)
- conversion to RDF
- loading into local database?
- accessing shared database, possibly with features like a query IDE, text searching, etc.

see also
- #1068

turbomam:
what local query engine to use? how to install?
- arq
  - https://jena.apache.org/download/
  - https://dlcdn.apache.org/jena/binaries/apache-jena-fuseki-4.9.0.tar.gz
  - https://www.bobdc.com/blog/jenagems
- robot

turbomam:
This issue is in the RDF/SPARQL/triplestore cluster. The active cleanup issue tracking removal of RDF generation and triplestore code is https://github.com/microbiomedata/nmdc-schema/issues/2970 — Remove remaining RDF generation and triplestore remnants.

Related open issues in this cluster:
- https://github.com/microbiomedata/nmdc-schema/issues/1069 — SPARQL queries over NMDC metadata in RDF
- https://github.com/microbiomedata/nmdc-schema/issues/1121 — list of RDF ideas
- https://github.com/microbiomedata/nmdc-schema/issues/1245 — document ways to run SPARQL queries
- https://github.com/microbiomedata/nmdc-schema/issues/2749 — review and clean up RDF/OWL generation
- https://github.com/microbiomedata/nmdc-schema/issues/500 — external_database_identifiers RDF conversion
- https://github.com/microbiomedata/nmdc-schema/issues/507 — NMDC types in RDF artifact
- https://github.com/microbiomedata/nmdc-schema/issues/838 — term IDs lacking prefixes in TTL output

turbomam:
Closing as not planned. The team is moving away from SPARQL/triplestore/RDF infrastructure in nmdc-schema (tracked in https://github.com/microbiomedata/nmdc-schema/issues/2970).

PR #1071, opened 2023-08-16
#1069 local sparql

turbomam:


turbomam:
need to clarify
- when is SPRQL not the ideal tool
- when to run it on a local file with robot
- when to run it on a local file with jena arq
- when to load it into a local TDB database and query it with jena arq
- when to query a remote RDF triplestore database interactively, or with very generic programmatic methods
- when to use nmdc_schema/class_sparql.py

turbomam:
closing in anticipation of improvements to [some GitHub action](https://github.com/microbiomedata/nmdc-schema/blob/main/.github/workflows/migration.yaml) 

note that [mongo_dump_api_emph.py](https://github.com/microbiomedata/nmdc-schema/blob/main/nmdc_schema/mongo_dump_api_emph.py), accessed through the [pure-export](https://github.com/microbiomedata/nmdc-schema/blob/main/pyproject.toml#L100) CLI can retrieve MongoDB data wihtout a PyMongo connection now, so that should be emphasized in any workflow that leads to a SPARQL query

Issue #1078, opened 2023-08-17
Increase funding_sources character limit


brynnz22:
To avoid a database migration.

turbomam:
A discussion on slack led to the conclusion to not limit the number of characters in `funding_sources`.

PR #1079, opened 2023-08-17
increase character limit of funding_sources to 200

brynnz22:
`funding_sources` needs a character limit of 200 (instead of 150) so it will not require a database migration. For issue #1078


turbomam:
thanks. waiting for checks to complete.

turbomam (approved):

PR #1095, opened 2023-08-30
This PR decreases the minimum cardinality for `Pooling.has_input` to  1

turbomam:
down from 2

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -151,7 +151,7 @@ classes:
     slots:
     slot_usage:
       has_input:
-        minimum_cardinality: 2
+        minimum_cardinality: 1 # was 2, which make intuitive sense, but some NEON data says "I pooled one sample"
       has_output:
         minimum_cardinality: 1
         maximum_cardinality: 1


turbomam:
which do we really want?

turbomam:
@aclum this will be an easy issue and PR for me to close, if you feel confident about the minimum and or maximum number of inputs into a `Pooling` process.

aclum:
@turbomam  Where is the example data driving this change? Conceptually sample does not go through pooling for a sequencing lab workflow unless there are at least two samples so I think we should leave the minimum cardinality as 2.

turbomam:
@aclum this PR was opened at the end of August. I think we tried setting a minimum cardinality of three on `Pooling.has_input` and then found that there were some Neon `Pooling`s with only two inputs, so we just completely removed the minimum cardinality constraint. We can leave it off, or change the minimum cardinality to 1 or 2.

But I don't have any motivating data and would be happy to close this PR without merging if you think that's best.

aclum:
@turbomam Am i reading the schema incorrectly? I read [this](https://github.com/microbiomedata/nmdc-schema/blob/9e4c58099b2fe064b6b08d3e24cfaf36458cbefb/src/schema/nmdc.yaml#L201C6-L201C31) as the minimum cardinality for slot has_input is currently enforced as 2 in the main branch of the nmdc schema for class Pooling.

turbomam:
Sorry for dragging you into this, @aclum. Yes, I see what you mean and suggest that we just close this PR without merging.

aclum:
@turbomam  agree

Issue #1101, opened 2023-09-08
Modify `Study` to account for data collection consortia, while reusing infrastructure 


turbomam:
see also
- https://github.com/microbiomedata/nmdc-schema/issues/1097

cc @brynnz22

turbomam:
Leave the class that has modeled studies _and will now also model data collection consortia_ named as `Study`. That will retain all of the external integration with MongoDB, the API, and the DataPortal.

Add a required slot that takes enumerated values that distinguish between true hypothesis driven studies and data/sample collection motivated research consortia.

That new slot will require migration. Kitware/nmdc-server will have to start interrogating the new slot in order to determine whether to draw a study page or a consortium page, but otherwise the impact to the NMDC ecosystem should be minimal.

turbomam:
- https://github.com/microbiomedata/nmdc-schema/pull/1098

uses slot `initiative_type` and enum `InitiativeTypeEnum` to drive the differentiation.

Let's reassess those names for the right level of specificity and also double check the preferred capitalization for `enum`s and `PermisibleValues`. I always forget!

We can also check the annotations an mappings for those new elements.

turbomam:
Regarding naming, do we want to paint ourselves into a corner of never being able to use `initiative_type` and `InitiativeTypeEnum` in any other context?

We could name the enum `ResearchInitiativeTypeEnum` and then allow different enums if `initiative_type` needs to be used in other classes in the future.

Possible disadvantages? My proposed enum name is longer. Anything else?

turbomam:
We could give some thought to what the real parent class of Study is, whether that class is currently modeled in the nnmdc-schema or not yet.

Is a process? Or a group of people? Something else?

turbomam:
[What slots are allowed on EnumDefinition](https://linkml.io/linkml-model/latest/docs/EnumDefinition/#slots)? We don't have to use them all!

turbomam:
I'd like to start micro-crediting, so that people who have contributed will get credit even if they don't make the PR.

I think this will require making an ORCID prefix. https://orcid.org/ ?

turbomam:
We should be in the habit of updating not just he valid examples, but also the invalid examples. Each invalid example should illustrate one single deviations from the requirements, and the file name should state that deviation.

As new constraints are put on classes, those shouldn't become additional deviations, but rather one new, well-named invalid data file should be created to illustrate the new constraint.

turbomam:
[...]

PR #1103, opened 2023-09-08
ResearchInitiativeTypeEnum model, examples and tests

turbomam:
- #1101


turbomam:
This could be improved even more by changing the title of `Study` to something like "A goal-oriented scientific effort that at least includes collection of samples that meet some criteria. May also include processing and analysis of those samples, along with reporting of the results against some hypothesis."

The current "A study summarizes the overall goal of a research initiative and outlines the key objective of its underlying projects." isn't really that good. We want to say what things are, not what they do.

Issue #1115, opened 2023-09-14
remove `default_curi_maps` from nmdc.yaml


turbomam:
`make squeaky-clean all test rdf-all` must still work.

Scrutinize the `rdf-all` output and logs for prefixes that aren't obtained from nmdc.yaml + imports, but rather requires external definition (in ???)

Document the findings.

PR #1116, opened 2023-09-14
Prefixes standardization (no default curi maps) and making some slot_usage attributes global

turbomam:
- #1115


turbomam:
Approved by @cmungall in personal Slack message.

Issue #93, opened 2021-07-24
rename omics processing -> instrument process


cmungall:
from https://github.com/microbiomedata/nmdc-schema/pull/91#issuecomment-885764948

wdduncan:
I think it might be better to have `instrument run` as a subclass of `omics processing`. We may have other kinds of `omics processing` we need to represent; e.g., kinds of sample preparation.

Also, although I prefer the label `instrument run` to `omics processing`, depending on how the above question is answered, it might be easier to add 'instrument run` as an alias to `omics processing`.

cc @dehays @turbomam

dehays:
I would prefer to drop the term 'omics processing' entirely in favor of using more descriptive terms.  Even analysis activity could be 'omics processing'.  'omics' is too general. So the aim in my recommending 'instrument process' is in making the term clearer in our use.

That relates to the term we control - now following OBI, you have 'biosample processing' as a OBI material processing (OBI:0000094) : planned process which results in physical changes in a specified input material.  And you have 'omics processing' is_a biosample_processing and map it to OBI assay (OBI:0000070):  planned process with the objective to produce information about the material entity

I agree on the mapping of 'omics_processing' - what I am asking be renamed 'instrument process' - to OBI assay. It takes a physical material as input and produces information as output.  But I do not think it is_a biosample_processing (OBI material processing) because it does not produce a physical output.  If I am reading OBI correctly, assay and material processing are siblings, both share planned process as a parent.

I do not think that sample preparation would be a kind of what we currently call 'omics processing', an OBI assay.  I do think it would be a kind of what we call biosample processing (OBI material processing), specifically it would be OBI 'sample preparation for assay', OBI:0000073 which has OBI material processing as its parent.

Happy to discuss - also curious why OBI (and now NMDC) use gerunds in places of available nouns for terms?  Is there some reasoning there?  Why 'processing' instead of 'process' in some cases?

wdduncan:
@dehays good questions! 
 - Re: gerunds: I cannot answer this. @cmungall is responsible for the creation of `omics processing`, but I suppose the "processing" bit is meant to draw attention to the notion that something is happening to the sample rather than that the material is doing something (e.g., material process may connote that the material is performing some process).
 - I agree with your analysis of the OBI mappings. The question (for me) is how much "ontological" rigor do we want in our mappings?  Broadly speaking, I think the idea was:
     - biosample processing -> process performed on samples
     - omics processing -> biosamples processings that produce omics information

[...]

Issue #24, opened 2020-12-08
biosample processing and omics processing should not be in a subclass relationship


cmungall:
biosample processing is documented to be a process that takes samples as inputs and generates samples as outputs; ie sample->sample

```yaml
  biosample processing:
    is_a: named thing
    description: >-
      A process that takes one or more biosamples as inputs and generates one or as outputs.
      Examples of outputs include samples cultivated from another sample or data objects created by instruments runs.
    slots:
      - has input
    slot_usage:
      has input:
        range: biosample
```

we can map this to to [material processing in OBI](https://www.ebi.ac.uk/ols/ontologies/obi/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FOBI_0000094)

this can be used to represent graphs of arbitrary depths of processes on samples, from treatments through to subsampling, making aliquots. In practice these graphs may be flat especially for retrospective samples

omics_processing doesn't have a sample as output - it generates omics data, i.e sample->data. Yet it is a subclass of biosample processing

```yaml
 omics processing:
    is_a: biosample processing
    in_subset: 
      - sample subset
    description: >-
      The methods and processes used to generate omics data from a biosample or organism.
    comments:
      - The IDs for objects coming from GOLD will have prefixes gold:GpNNNN
    slots:
      - part of
      - has output
      - omics type
    slot_usage:
      id:
        description: >-
          The primary identifier for the omics processing. E.g. GOLD:GpNNNN
      name:
        description: >-
          A human readable name or description of the omics processing.
      alternate identifiers:
        description: >-
          The same omics processing may have distinct identifiers in different databases (e.g. GOLD and EMSL, as well as NCBI)
      part of:
        range: study
      has output:
        range: data object
```

I think this can be mapped to [assay in OBI](https://www.ebi.ac.uk/ols/ontologies/obi/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FOBI_0000070)

I don't think this is quite coherent, if OP inherits from BP then it should inherit the condition that the output is a sample

It may be more straightforward to have these be siblings

If we want we can have a grouping class for both of these

brynnz22:
[...]

PR #1118, opened 2023-09-14
Issue 24 omics processing move

brynnz22:
Made the class `OmicsProcessing` to be a subclass of `PlannedProcess` (no longer a subclass of `BiosampleProcessing`) and made `BiosampleProcessing` a subclass of `PlannedProcess` with subclasses `LibraryPreparation`, `Extraction`, and `Pooling`. 

I also added some mappings and definitions. I think I saw an issue floating around somewhere that you wanted to add more descriptions to the schema, so I figure I'd add them as they come up! 

Addresses: https://github.com/microbiomedata/nmdc-schema/issues/24 and issue https://github.com/microbiomedata/nmdc-schema/issues/93 (although I don't think we want to rename `OmicsProcessing` `InstrumentProcess`), but it sounds like this PR could close out both of these issues?

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -1459,7 +1459,7 @@ classes:
     is_a: PlannedProcess
     description:
       A process that takes one or more biosamples as inputs and generates
-      one or more materials as outputs. An example of an output include samples cultivated from another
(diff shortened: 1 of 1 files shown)


turbomam (approved):
let's look at the class hierarchy in the web documentation and plan next steps

Issue #1149, opened 2023-09-29
update `pure-export` phase in project.makefile for pure prod and pure dev modes


turbomam:

turbomam:
see notes on PR
- https://github.com/microbiomedata/nmdc-schema/pull/1150

turbomam:
Related: https://github.com/microbiomedata/nmdc-schema/issues/2969 — Replace dump-validate chain with direct MongoDB validation. The work proposed in 2969 would make the pure-export targets less central.

turbomam:
Closing as not planned. The team is moving away from SPARQL/triplestore/RDF infrastructure in nmdc-schema (tracked in https://github.com/microbiomedata/nmdc-schema/issues/2970).

PR #1150, opened 2023-09-29
`local/mongo_as_unvalidated_nmdc_database.yaml` target needs pure dev input and pure prod input modes

turbomam:
- #1149


turbomam:
I'm putting the emphasis now on better documentation

turbomam:
I merged a PR that 
- added some commentary in the project.Makefile 
- added some more options on the  `pure-export` CLI like `--client-base-url`

but run modes that hit 
- both a PyMongo connection to the dev data base and the dev API
- or both a PyMongo connection to the prod data base and the prod API

have not been created yet

Note that `pure-export` can run in the absence of a PyMongo connection now if `--skip-collection-check` is asserted. That mode could benefit from checking the number of documents in a collection with the API.

PR #1154, opened 2023-10-02
polymorphic sample prep collection

turbomam:
- #1153


turbomam:
Note that I didn't define any of the data entities that are inputs into or outputs from the processes I modeled. That's legal in LinkML now, because there's no referential integrity checking in the most common data formats like YAML or JSON. Referential integrity checking will probably be added to the LinkML validation soon, at least as an option. @pkalita-lbl may be able to give us an update on that at some point.

JamesTessmer:
Is this how we should reformat the SampleOperations sets?
I think this covers what we need and seems better than having sets for every sample operation type.
I think we would still need a set for each protocol like the gravimetricWaterContent,MicrobialBiomass, etc.

turbomam:
> I think we would still need a set for each protocol like the gravimetricWaterContent,MicrobialBiomass, etc.

Why's that? Can you show me how those protocols are defined now?

JamesTessmer:
> > I think we would still need a set for each protocol like the gravimetricWaterContent,MicrobialBiomass, etc.
> 
> Why's that? Can you show me how those protocols are defined now?

Right now they're defined as classes, there's the MicrobialBiomassDetermination and then we have MicrobialBiomassDetermination_set in the sample_opeartions_merge_prep branch.

There needs to be some way to identify what protocol is being used/followed so if we can add something to identify that I suppose we wouldn't need a set for each one, but I would also like to see what Anastasiya thinks about that.

anastasiyaprymolenna:
@JamesTessmer @turbomam James is correct. The schema structure outlined here has no way to indicate the type of protocol being used. We need to allow a `type_of` slot to be present that indicates the type of protocol. There needs to be a level of organization between SamplePreparation class and the sample_preparation_set slot. https://github.com/microbiomedata/nmdc-schema/pull/1154/files#diff-31491f32a70f5c80544a7da274f28cdbe03304c4a34c0f7a1cbc900c0b4e2eb9

turbomam:
Thanks for sharing you insights @JamesTessmer and @anastasiyaprymolenna 

I'm closing this PR in favor of https://github.com/microbiomedata/nmdc-schema/pull/1238

We can still refer to the comments in this PR for future discussion.

Issue #1192, opened 2023-10-16
document deletions between commits bacbe21d4 and 43d31a60b


turbomam:
https://github.com/microbiomedata/nmdc-schema/compare/bacbe21d4...43d31a60b

Most of this is fine, although my close review raised some new issues
- [x] discuss plan for auto-generated release notes vs curated `CHANGELOG.md`
- [x] delete empty `nmdc_schema/get-element-names.py`, since it would be easy to recreate
- [ ] what is our use case for `data_portal_subset`? Right now its annotations are vague and it's just associated with a single slot
- [ ] check `publication_dois` pattern

turbomam:
## `project.Makefile`

```
# recommended setup:
#   1. . ~/sshproxy.sh -u {YOUR_NERSC_USERNAME}
#   2. ssh -i ~/.ssh/nersc -L27777:mongo-loadbalancer.nmdc.production.svc.spin.nersc.org:27017 -o ServerAliveInterval=60 {YOUR_NERSC_USERNAME}@dtn01.nersc.gov

#   		--selected-collections functional_annotation_agg \ # huge, no publicly available reference data (kegg)
#   		--selected-collections metaproteomics_analysis_activity_set \ # next slowest
```

the standard `--client-base-url` is  https://api.microbiomedata.org

routine `--selected-collections` 

```
		--selected-collections biosample_set \
		--selected-collections data_object_set \
		--selected-collections extraction_set \
		--selected-collections field_research_site_set \
		--selected-collections library_preparation_set \
		--selected-collections mags_activity_set \
		--selected-collections metabolomics_analysis_activity_set \
		--selected-collections metagenome_annotation_activity_set \
		--selected-collections metagenome_assembly_set \
		--selected-collections metagenome_sequencing_activity_set  \
		--selected-collections metatranscriptome_activity_set \
		--selected-collections nom_analysis_activity_set \
		--selected-collections omics_processing_set \
		--selected-collections pooling_set \
		--selected-collections processed_sample_set \
		--selected-collections read_based_taxonomy_analysis_activity_set \
		--selected-collections read_qc_analysis_activity_set \
		--selected-collections study_set
```

these are some default settings that could be included for clarity:

```
		--env-file local/.env \
		--mongo-db-name nmdc \
		--mongo-host localhost \
		--mongo-port 27777 \
```

`--skip-collection-check` should not be used routinely

```
# from riot:
#   WARNING: java.io.tmpdir directory does not exist

[...]

PR #1193, opened 2023-10-17
restore some lines deleted between commits bacbe21d4 and 43d31a60b

turbomam:
- #1192


turbomam:
the `migration-with-java-action` from `.github/workflows/migration.yaml` should run after this PR is merged

PR #1197, opened 2023-10-18
Polymorphic "microbial biomass" collection: ranges can't be abstract and `--include-range-class-descendants` validation option required

turbomam:
see fresh-microbioal_biomass_schema.Makefile

not to be considered urgent


src/schema/nmdc.yaml
```diff
@@ -2401,6 +2404,23 @@ slots:
       Applies to a property that links a database object to a set of objects.
       This is necessary in a json document to provide context for a list, and to allow
       for a single json object that combines multiple object types
+
+  process_set:
+    mixins: object_set
+    range: PlannedProcess
```
sujaypatil96, line 2410: I was looking through the schema, and saw that `PlannedProcess` (and `MaterialEntity`) were both abstract classes.

I think the issue as to why JSON Schema validation isn't going through is because by asserting `range: PlannedProcess` you're saying that it's okay to create and initialize `process_set` with instances of `PlannedProcess` which we cannot do?

Issue #1198, opened 2023-10-18
Merge SubSampling and Extraction classes, with data that illustrates the relationship, into nmdc-schema


turbomam:
Update this issue with other people who will be consulted like
Consulted:
- @mslarae13 (will attend the meeting)
- @cmungall 
- @sujaypatil96 

Add SubSampling and Extraction class activities to the schema. Update what @JamesTessmer  & @anastasiyaprymolenna  have created & address @turbomam 's concerns with the modeling.

Completion criteria:
- The 2 classes will be added 
- Valid and invalid data will be added for test. 
- Classes modeling and examples will be updated

mslarae13:
I wish GitHub tracked WHO updated another persons comment or issue description. But @turbomam  I tried to add some more context. Does that sound good?

turbomam:
yes thank you @mslarae13

anastasiyaprymolenna:
@mslarae13 The slot "extractant" needs to be reviewed. Could you fill in where you are defining this from?

mslarae13:
@anastasiyaprymolenna  do you still need me to look at extractant?

PR #1199, opened 2023-10-18
Merge in SubSamplingProcess and supporting changes

turbomam:
- #1198 

will add commits soon

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -1529,6 +1529,28 @@ classes:
     broad_mappings:
       - OBI:0000094
 
+  SubSamplingProcess:
+    description: >
+      Separating a sample aliquot from the starting material for downstream activity.
+    see_also:
+      - http://purl.obolibrary.org/obo/OBI_0000744
+    todos:
+      - add intermediate class between PlannedProcess and this (MaterialProcessing potentially)
+      - Montana to visit slot descriptions
+      - write data that uses all the slots
+      - add 'has_output' usage for processed sample - James
+    notes:
+      - A subsample may be (a) a portion of the sample obtained by selection or division; 
+        (b) an individual unit of the lot taken as part of the sample; (c) the final unit of multistage sampling. 
+        The term 'subsample' is used either in the sense of a 'sample of a sample' or as a synonym for 'unit'. 
+        In practice, the meaning is usually apparent from the context or is defined.
+    is_a: PlannedProcess 
+    slots:
+      - container_size
+      - container_type
+      - volume
+      - temperature
+
   OmicsProcessing:
     aliases:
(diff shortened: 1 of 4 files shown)


turbomam:
see 
- https://github.com/microbiomedata/nmdc-schema/pull/1201

JamesTessmer:
Added:
SubSamplingProcess class to [nmdc.yaml](https://github.com/microbiomedata/nmdc-schema/blob/1198-merge-subsampling-and-extraction-classes-with-data-that-illustrates-the-relationship-into-nmdc-schema/src/schema/nmdc.yaml)
Valid [data file](https://github.com/microbiomedata/nmdc-schema/blob/1198-merge-subsampling-and-extraction-classes-with-data-that-illustrates-the-relationship-into-nmdc-schema/src/data/valid/SubSamplingProcess-minimal.yaml)

Modified:
[sample_prep.yaml](https://github.com/microbiomedata/nmdc-schema/blob/1198-merge-subsampling-and-extraction-classes-with-data-that-illustrates-the-relationship-into-nmdc-schema/src/schema/sample_prep.yaml) due to slot definition conflicts, and previous discussion that we would redefine slots from sample_prep.

src/schema/nmdc.yaml
```diff
@@ -2781,3 +2808,22 @@ slots:
...
+  container_size:
+    range: QuantityValue
+    description: The volume of the container an analyte is stored in or an activity takes place in
+
+  container_type:
+    range: ContainerTypeEnum
+    description: >-
+      A type of container. i.e. test tube, falcon tube, etc.
+
+  volume:
```
cmungall, line 2821: give this a name that resolves ambiguity as to what volume is being referred to

src/schema/nmdc.yaml
```diff
@@ -2781,3 +2808,22 @@ slots:
...
+    range: QuantityValue
+    description: The volume of the solvent/solute being used, not the input.
```
cmungall, line 2823: Here we are defining a very general schema-wide slot `volume`; we can imagine this being used many places, e.g. volume of a container, of a piece of soil. But the description is very specific. Either do one of:

1. Make a more specific slot such as solvent_volume, and keep the definition
2. Keeping a generic volume, and amending this definition to be more generic, and putting the specific description on slot_usage in SubSamplingProcesss

cmungall (changes requested):

turbomam:
Thanks James. Ina addition to @cmungall's request, could you please also add micro-crediting for yourself or others into the elements you want to merge (classes, slots, enumeration).

JamesTessmer:
Added the change to make volume more general, and add a more specific description in the slot usage. Also added credits for contributors.

PR #1202, opened 2023-10-18
Add backticks around file path to format it as code

eecavanna:


eecavanna:
I jumped the gun on this. I would rather it be formatted as a hyperlink since it refers to a document in the same repo.

turbomam (approved):
thanks!

PR #1256, opened 2023-10-31
1110 add glbrc aliases

mslarae13:
@turbomam  can you review the slot_usage I added in nmdc.yaml for wind_seed.

If you can confirm I selected the right attributes, we'll move ahead with the rest of the slots the already exist in NMDC but need aliases to connect to GLBRC.

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -1360,6 +1360,12 @@ classes:
         todos:
           - Currently, the comments say to use UUIDs. However, if we implement assigning NMDC identifiers with the minter we dont need to require a GUID. It can be an optional field to fill out only if they already have a resolvable ID.
 
+      wind_speed:
+        aliases:
+          - Wind_Speed_Mean
+        alt_descriptions:
+          -"Wind speed measured at 10 m every 10 seconds, integrated and stored hourly. Arithmetic mean per sampling day reported."
+
     exact_mappings:
       - OBI:0000747
       - NCIT:C43412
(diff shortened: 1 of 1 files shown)


mslarae13:
@bmeluch FYI

turbomam:
This is a great start. Thanks @mslarae13. I'm checking to see why the PR checks failed now.

mslarae13:
@turbomam  did you figure out what was wrong? I expect I forgot `value: ` somwhere that I didn't know it was needed or something like that?

turbomam:
> did you figure out what was wrong

I added a review with explanatory comments, which you should be able to see above. I also added a commit that fixed the problem.

@mslarae13 let me know if you want to discuss this more, or if I should just wait for you to add more `aliases` or `alt_descriptions` in the structure I illustrated above

mslarae13:
After discussion, it was decided not to prioritize adding the GLBRC descriptions and alias' until confirmation of expanded use. 

I can confirm that GLBRC does have some consistency... Adina sent me this table ... https://lter.kbs.msu.edu/datatables/12
(download full .csv tables for more details and descriptions)

mslarae13:
@turbomam can you explain why you closed this PR, deleted the branch, and then commented on the content?
I think it's fine to delete, and these changes aren't going to happen but we should add some better details and tracking.

turbomam:
@mslarae13 Please see https://nmdc-group.slack.com/archives/CFVF3G3H7/p1729196160120289 for an explanation of why this was closed. (Although the thought you shared above are helpful answers to the questions I asked in Slack.)

There are probably a few more of these closed via `main-old` PRs that the two of us or others should revisit

mslarae13:
Ok so it was just cuz the branch was pre-berkeley and IF the work is to be finished we should do so in a new branch & PR.
Makes sense! 
Thanks!

PR #1266, opened 2023-11-01
487 Add slot description

mbthornton-lbl:
This PR adds a description slot and value to Metagenome Raw Reads in nmdc.yaml


aclum (changes requested):
The data objects in mongo that use the enum are all paired end.  Can you see who made the enum 'Raw Compressed Metagenome Interleaved Paired-End Reads', based on usage that and 'Metagenome Raw Reads' are redundant so we should deprecate one of them. There is no use of 'Raw Compressed Metagenome Interleaved Paired-End Reads' yet in mongo prod.

turbomam (approved):

aclum (approved):

PR #1275, opened 2023-11-02
DO NOT MERGE MongoDB OK with schema v9 after two migrations

turbomam:
- #1272


turbomam:
try running `poetry update && poetry install && make squeaky-clean all test brute-force-all`

Issue #1288, opened 2023-11-03
add OmicsProcessing as an applicable Class for insdc_bioproject_identifiers


aclum:
GOLD links insdc_bioproject_identifiers identifers to their project records. Currently only applicable class for this slot is Study, GOLD uses these identifiers both at the study and project level. It would be easier to pull the insdc_bioproject_identifiers directly instead of doing a lookup from insdc_bioproject_identifiers to sra to populate [insdc_experiment_identifiers](https://microbiomedata.github.io/nmdc-schema/insdc_experiment_identifiers/).

@turbomam  let us know what you think about this.

cc @mbthornton-lbl

PR #1296, opened 2023-11-06
allow slot insdc_bioproject_identifiers on class OmicsProcessing

turbomam:
- #1288


turbomam (commented):
Since the least commit didn;t trigger the CH actiosn checks, I'm doing `make squeaky-clean all test validate-filtered-request-all validate-polymorphic locally` now and will merge if that passes.

aclum (approved):

Issue #1253, opened 2023-10-27
Edit `FunctionalAnnotationAggMember ` to work with MetaP data


aclum:
We added a nmdc schema Class for the gene aggregation for the sequencing workflows, see https://microbiomedata.github.io/nmdc-schema/FunctionalAnnotationAggMember/, but not for the metaproteomics. I believe we wanted all the collections in mongo to have a slot on class Database, in order to do that we need a Class for the metaproteomics gene function aggregation. The collection in prod mongo that current contains this data is 'metap_gene_function_aggregation'

related to 
https://github.com/microbiomedata/nmdc-schema/issues/1252

Based on that we need class with slots (based on how the collection is currently populated)

metaproteomic_analysis_id
gene_function_id (slot already exists -update domain_of for this slot)
count (slot already exists - update domain_of for this slot)
is_best_protein (need new slot)

all slots should be required

Then Class Database needs a slot 'metap_gene_function_aggregation' with a range of the new metap gene function aggregation Class

@SamuelPurvine @mslarae13 @pdpiehowski @picowatt please confirm requirements, in particular if the slots should be required or not.

cc @turbomam 

- [x] update schema (Alicia)
- [x] write migration code for mongo prod (Michael)

aclum:
This was discussed today at the metap meeting and folks agree on adding this as a Class to the schema with the slots as required. I talked with @cmungall about this last week in person and IGB and he was supportive as well.

mslarae13:
@aclum is this a "WorkflowExecution" subclass based on the new schema structure we're doing?

SamuelPurvine:
And, as mentioned over in issue 1309, the best_protein slot in this aggregation class should be renamed to something like has_best_protein or is_best_protein and be given range boolian false/true.

aclum:
@mslarae13  no, this is an aggregation table from the output of the [MetaproteomicsAnalysisActivity](https://microbiomedata.github.io/nmdc-schema/MetaproteomicsAnalysisActivity/)/ 
i am setting this up the same way that @turbomam created the [FunctionalAnnotationAggMember ](https://microbiomedata.github.io/nmdc-schema/FunctionalAnnotationAggMember/) class to aggregate the sequencing based annotation results.

aclum:
@naglepuff  or @marySalvi  Could you comment on what fields from the metap_gene_function_aggregation collection the data portal ingest code uses? We have to change the name of best_protein since it is already being used in a different way within the schema.

[...]

Issue #1252, opened 2023-10-27
make slots for Class FunctionalAnnotationAggMember required


aclum:
I think all slots for this Class should be required. For the data portal to be able to have the kegg search work properly they need to know the term id, count, and what WorkflowExecutionActivity it was derived from.
slots in question are 
- metagenome_annotation_id
- gene_function_id
- count

These slots are currently all optional.
@naglepuff  @scanon please comment.

naglepuff:
Sounds good to me

ssarrafan:
@aclum this is not assigned to anyone, and no updates since Oct so I'm removing from the sprint and adding the backlog label. 

Please assign and add to new sprint if it's an active issue.

aclum:
This is in review, it mixed in with a pull request (https://github.com/microbiomedata/nmdc-schema/pull/1309) that is failing tests that needs to be debugged which addresses related issue https://github.com/microbiomedata/nmdc-schema/issues/1253. Its fine to leave this in the backlog for now since I'm not sure we'll get the related issues resolved this sprint.

aclum:
We uncoupled the two issues. The PR to address just this issue should get merged in today.

PR #1309, opened 2023-11-08
Make FunctionalAnnotationAggMember slots required

aclum:
Updating slots to required for class FunctionalAnnotationAggMember, adding new class and slots for the metap aggregation collection.
See #1252 and #1253


aclum:
@turbomam Let me know if this looks okay. Also, I wasn't sure how to add best_protein as required just for the new Class MetaPFunctionAggMember since best_protein applies to other classes currently.

aclum:
I need some input here, the example data failed validation. I modeled this new class based on the collection that currently exists in mongo.  It uses an existing slot, [best_protein](https://microbiomedata.github.io/nmdc-schema/best_protein/). However the slot is used differently. Based on the current schema the range for best_protein is GeneProduct, with applicable classes of PeptideQuantification and ProteinQuantification. The way best_protein is used in the current metap_gene_function_aggregation collection it is populated with a true/false boolean (@scanon  please confirm). 
If we make a new slot name this will require a migration and the aggregation code will have to be updated.

Please weigh in here @SamuelPurvine  @pdpiehowski @turbomam 

example metap_gene_function_aggregation document
```
{
  "_id": {
    "$oid": "640fa0d0c239df02c4d727c7"
  },
  "metaproteomic_analysis_id": "nmdc:ec0db6c4103faf47d3393ec27d541639",
  "gene_function_id": "KEGG.ORTHOLOGY:K02358",
  "count": 17,
  "best_protein": true
}
```

turbomam:
The linkml/linkml repo has some issues that are relevant to this topic: when is [monotonic specialization](https://github.com/linkml/linkml/issues?q=is%3Aissue+is%3Aopen+monotonic) of a schema element required, recommended or optional. See especially @cmungall's [Clarify semantics of overrides/non-monotonicity](https://github.com/linkml/linkml/issues/1730) 

The most common definitions of montonic are about mathematical functions: f(x) can only have one value. In the case of data modeling, it means (among other things) that a slot _can have different ranges_ in different situations, but **all of the ranges must be included within some common parent**.

Allowing `best_protein` to have two completely incompatible and unrelated ranges when used with two different classes is about as non-monotonic as you can get. The `GeneProduct` class and the boolean type have no common parent.

From my perspective, a new slot is required.

SamuelPurvine:
metap_gene_function_aggregation is not a thing we created, it was made after our data were put into the system. best_protein is the term we are using for the protein that is most likely to be associated with the peptide that was identified, using parsimony logic. I can see where "best_protein" might have been used in a quick tossing together of the aggregation function to meet the needs, but minimally this should have been named something like is_a_best_protein to indicate whether the functional category is associated with a best_protein entry, which should have the same kind of data as a Gene_Product. We at PNNL certainly would not have tried to set best_protein to have those two ranges as that would just crerate problems down the roads, but again, we didn't write that aggregation function (to the best of my knowledge). So, if a new class needs to be created, please make it the boolian in the aggregation function and call it something that makes sense, such as the suggestion above.

aclum:
@mbthornton-lbl  and @eecavanna wrote the migration code this week so @turbomam  this ready to be reviewed and merged in. This can wait until beginning on January.

src/schema/basic_slots.yaml
```diff
@@ -217,7 +217,11 @@ slots:
   biomaterial_purity:
     domain: ProcessedSample
     range: QuantityValue
-
+  
+  is_best_protein:
+    description: Slot to store a boolean value. If populated with a  value of true the gene function is the best protein hit, if false it is not the best protein hit.
```
eecavanna, line 222: There are two spaces between "a" and "value" here.

src/schema/nmdc.yaml
```diff
@@ -2324,17 +2346,32 @@ slots:
...
     inlined_as_list: true
-
+  
+  # for metaPFunctional agg  
+  metap_gene_function_aggregation:
+    range: MetaPFunctionAggMember
+    multivalued: true
+    inlined: true
+    inlined_as_list: true
+  metaproteomic_analysis_id:
```
eecavanna, line 2371: I noticed "metaproteomic" is singular (or an adjective) in `metaproteomic_analysis_id`, but plural in `MetaproteomicsAnalysisActivity`. I don't know whether that was intentional or not.

src/schema/nmdc.yaml
```diff
@@ -2324,17 +2346,32 @@ slots:
   metagenome_annotation_id:
     domain: FunctionalAnnotationAggMember
     range: WorkflowExecutionActivity
+    required: true
+    description: The value for slot id for the analysis activity (WorkflowExecutionActivity) that generated functional annotation
   gene_function_id:
     range: uriorcurie
+    required: true
+    description: The identifier for the gene function, ie KEGG.ORTHOLOGY:K00627
```
eecavanna, line 2354: Assuming `KEGG.ORTHOLOGY:K00627` is an example value, I'd recommend:

```diff
- description: The identifier for the gene function, ie KEGG.ORTHOLOGY:K00627
+ description: The identifier for the gene function; e.g., KEGG.ORTHOLOGY:K00627
```

Reference: https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/

src/schema/nmdc.yaml
```diff
@@ -2324,17 +2346,32 @@ slots:
   metagenome_annotation_id:
     domain: FunctionalAnnotationAggMember
     range: WorkflowExecutionActivity
+    required: true
+    description: The value for slot id for the analysis activity (WorkflowExecutionActivity) that generated functional annotation
```
eecavanna, line 2350: Proposal:

```diff
-    description: The value for slot id for the analysis activity (WorkflowExecutionActivity) that generated functional annotation
+    description: The id of the WorkflowExecutionActivity that generated this FunctionalAnnotationAggMember
```

Also, I find it confusing that the slot name is `..._annotation_id`, but the description says it's an identifier of a `WorkflowExecutionActivity` (as opposed to some kind of "annotation activity" or "annotation"). My background is in software development, not biology—maybe it would be clear to someone with a background in biology.

src/schema/basic_slots.yaml
```diff
@@ -217,7 +217,11 @@ slots:
   biomaterial_purity:
     domain: ProcessedSample
     range: QuantityValue
-
+  
+  is_best_protein:
```
eecavanna, line 221: I like that this boolean slot's name starts with `is_...`. 👏

src/schema/basic_slots.yaml
eecavanna: I think "Slot to store a boolean value" is redundant, given the `range: boolean` below.

eecavanna:
I finished updating the migrator code to be compatible with the `main` branch (i.e. the changes that were merged into the `main` branch via commit https://github.com/microbiomedata/nmdc-schema/commit/42e24bd67e8ce7487907120efb3f9ac9879a43e7).

src/schema/basic_slots.yaml
turbomam: Yes, this needs to be rewritten. What does it mean for a function to be a best protein hit? When I think of "best protein hit" I think of matching raw data, like mass spectra or nucleotide sequences, to some database. The assignment of a function also has uncertainly of its own, based on inference with something like a HMM.

What class(es) are intended to be in the domain of `is_best_protein`?

If this slot is just about identifying the protein inference that best matches the observed data, the description should be something like "True when the named protein is the best explanation of teh observed data" or something like that.

src/schema/nmdc.yaml
turbomam: Well, the range of the slot **is** `WorkflowExecutionActivity`. The description doesn't really need to say that's it's value is an `id`,  but the textual `description` does agree well with the logical definition, specifically the `domain` and `range`. I would either leave this is as-is, or improve it by changing the slot `name`. Unfortunately 'generated_by' is already taken!

turbomam (commented):
Idealistically speaking, I'm still disappointed by the dissimilarity between 

`functional_annotation_agg` and `FunctionalAnnotationAggMember` vs

src/schema/nmdc.yaml
```diff
@@ -143,13 +143,36 @@ subsets:
       - Assignment of the data_portal_subset is incomplete in the schema.
 
 classes:
-
+  MetaPFunctionAggMember:
+    description:
+      A class for the metaproteomics aggregation table, used to enable NMDC's data portal KEGG search
```
turbomam, line 148: ideally, `description`s say what an element is (in reality), they don't reiterate an element's type (class) or how it will be used. (That second part is better suited for a comment)

An example of this would be 'A count of the number of times a sample (or gene or whatever) has been annotated with the specified gene function"

src/schema/nmdc.yaml
eecavanna: Schema concepts aside (I'll defer to @turbomam on those), I think there's a word (e.g., "the", "a", "this") missing in:
```diff
- that generated functional annotation
+ that generated this functional annotation
```

src/schema/nmdc.yaml
aclum: I did not make the name of the workflow execution activity subclass or the value of the field in mongo. metaproteomic_analysis_id is a field in a documents that are already in a collection in mongo, this PR is just to make sure the collection has a matching Database slot, class and schema slots that correspond to what already exists. The exception to that is `is_best_protein` because that slots can't have two different meanings.

src/schema/nmdc.yaml
aclum: Moved to use examples syntax.

eecavanna:
Thanks for making those changes. I don't have any concerns (all of the ones I had have been resolved).

mbthornton-lbl (approved):

turbomam (changes requested):
I could be convinced to merge this as is, but please consider moving the `required`s to global definitions,

src/schema/basic_slots.yaml
```diff
@@ -217,7 +217,11 @@ slots:
...
+    description: The protein that is most likely to be associated with the peptide that was identified, using parsimony logic.
+    range: boolean
```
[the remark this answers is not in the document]
aclum: outdated

turbomam (approved):

PR #1312, opened 2023-11-08
Return the transformed study from the transformation function

### LINKED ISSUE #1311: Two transformation functions do not return the transformed dictionary
### Summary

While running a transformation function named `remove_doi_slots` from within a Python notebook that `import`s the containing Python class from the `nmdc-schema` Python package on PyPI, Python raised an `AttributeError` exception because a value of `None` doesn't have an attribute named `get`.

### Screenshot

image

### Related snippets

This is where the exception was raised:

https://github.com/microbiomedata/nmdc-schema/blob/918056b61508b2094433abb15fbb0e1578fd014f/nmdc_schema/migration_recursion.py#L382-L384

These two transformation functions don't return the transformed dictionary (i.e. they implicitly return `None` since they don't have `return` statements):

https://github.com/microbiomedata/nmdc-schema/blob/918056b61508b2094433abb15fbb0e1578fd014f/nmdc_schema/migration_recursion.py#L316-L327

https://github.com/microbiomedata/nmdc-schema/blob/918056b61508b2094433abb15fbb0e1578fd014f/nmdc_schema/migration_recursion.py#L368-L376

As a result, transformation functions "further down the pipeline" won't receive a dictionary as input.

### Proposal

End each transformation function with:

```py
return study
```

eecavanna:
Single commit to fix #1311

### DIFF
--- a/nmdc_schema/migration_recursion.py
+++ b/nmdc_schema/migration_recursion.py
@@ -324,6 +324,8 @@ class Migrator_from_8_1_to_9_0(MigratorBase):
                 }
                 study.setdefault('associated_dois', []).append(new_doi)
 
+        return study
+
     def fix_award_dois(self, study: dict):
         """Moves current DOIs in the award_dois slot to the new associated_dois slot.
          It also changes some DOIs that were incorrectly labeled as award to dataset DOIs.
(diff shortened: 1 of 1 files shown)


brynnz22 (approved):
Thanks @eecavanna! I saw this a while ago and I thought I added the return statements, but I guess I didn't. Sorry for the hiccup!

PR #1324, opened 2023-11-09
committed to `emsl_project_identifiers`

turbomam:
@brynnz22 can you please add a description?

- #927


brynnz22:
@turbomam I added a description. Also, regarding your other `todos`, @lamccue mentioned that currently, the identifier is always a 5 digit integer.  We can't guarantee that it will not change...we have no control over that. But to her knowledge, there is no plan make any changes and we can expect 5 digit integers for the foreseeable future.

turbomam:
Thanks @brynnz22. I am going to merge this, but I hope we can elaborate on the description to make it more informative

Issue #1331, opened 2023-11-10
delete all instrument/device classes and start over, distinguishing between named/anonymous


turbomam:
pass this off to @anastasiyaprymolenna

turbomam:
add in data generation vs preparation after this

brynnz22:
Here is a metadata schema a working group for RDA developed that might be useful for inspiration and the "I" in FAIR. I was hoping they might have a controlled vocabulary for instrument type, but it does not look like it

PR #1332, opened 2023-11-10
Refactor `LabDevice`, `Instrument` and `InstrumentValue`

turbomam:
- #1331 

will add similar classes in in subsequent commit


JamesTessmer:
These look good to me, but I'd also like to see if everyone agrees before approving.

JamesTessmer:
> 

Oops, I misread the 'added as reviewer' email as 'ready for review'

turbomam:
I should add an example of some `PlannedProcess` instance relating to a named or anonymous instrument. The relating slot would use some language like "used", but not `used`, because that has an incompatible `range`.

anastasiyaprymolenna:
We need to include more permissible values in the instrument enums. See this commit for the sample processing enums https://github.com/microbiomedata/nmdc-schema/commit/b2e8c21ad57d8b9c769ed66d678749ab2d2853b3

turbomam:
better names for NamedInstrument:
- analytical instrument?
- data generating instrument

turbomam:
what kinds of things can have a relationship with XInstruments, via what slots?

not `used`

Issue #1310, opened 2023-11-08
Transformation function raises `FileNotFoundError` regarding data YAML file


eecavanna:
### Summary

While running a transformation function named `fix_award_dois` from within a Python notebook that `import`s the containing Python class from the `nmdc-schema` Python package on PyPI, Python raised a `FileNotFoundError` exception because it couldn't find a file at the path specified.

### Screenshot

![image](https://github.com/microbiomedata/nmdc-schema/assets/134325062/863f3fa2-8a94-474e-a8e1-50f5e2529eb7)

### Related snippets

Invocation of `load_yaml_file` function:

https://github.com/microbiomedata/nmdc-schema/blob/918056b61508b2094433abb15fbb0e1578fd014f/nmdc_schema/migration_recursion.py#L332-L333

Definition of `load_yaml_file` function:

https://github.com/microbiomedata/nmdc-schema/blob/918056b61508b2094433abb15fbb0e1578fd014f/nmdc_schema/migration_recursion.py#L90-L94

### Proposal

Reference the file in a way that works regardless of where the function is being invoked.

Suggestion (untested):

```diff
+ from importlib import resources

# ...

 def load_yaml_file(filename): 
     """Loads a YAML file into a Python dict.""" 
-    with open(filename, "r") as f: 
+    with resources.open(filename, "r") as f: 
        data = yaml.safe_load(f) 
     return data 
```

Docs:
- Official: https://docs.python.org/3/library/importlib.resources.html#module-importlib.resources
- Official: https://docs.python.org/3/library/importlib.resources.abc.html#importlib.resources.abc.Traversable.open
- Third-party: https://dev.to/bowmanjd/easily-load-non-python-data-files-from-a-python-package-2e8g

eecavanna:
As a workaround for the current migration, I can download the referenced file from the `nmdc-schema` GitHub repo and put it at the path `assets/misc/study_dois_changes.yaml` relative to the Python notebook.

![image](https://github.com/microbiomedata/nmdc-schema/assets/134325062/477a7707-509f-4ee7-9ac9-a74834a9cf00)

![image](https://github.com/microbiomedata/nmdc-schema/assets/134325062/4268904f-ce19-4c35-8c12-d6ef55735067)

turbomam:
Thanks @eecavanna 

@brynnz22 would be it be OK for me to change the assignee to you?

eecavanna:
[...]

Issue #1291, opened 2023-11-04
Add doctests to example transformation functions


eecavanna:
`doctest` is in the Python standard library.

Docs: https://docs.python.org/3/library/doctest.html

These tests can be used to unit test the transformation functions.

PR #1340, opened 2023-11-12
Refactor migrator classes into own modules and fix YAML file loading

eecavanna:
### Summary

- Created a new package named `migrators` that contains the migration-specific classes
- Moved `MigratorBase` class from `migration_recursion.py` into its own module within the `migrators` package
- Moved each migration-specific class from `migration_recursion.py` into its own module within the `migrators` package
- Moved a YAML file used by one migration-specific class into a new `assets/` directory within the `migrators` package
- Moved CURIE-related helper functions directly called by `migration_recursion.py` into a new `CurieMigrator` class
- Removed the unused `force_prefix` property from the `MigratorBase` class (it still exists on the `CurieMigrator` class)
- Capitalized names of regex pattern "constants" per [PEP 8](https://peps.python.org/pep-0008/#constants) style guide
- Configured Poetry so consumers of the PyPI package can `import` things residing in `nmdc_schema/migrators/` (instead of only things residing at the top level of `nmdc_schema/`).
    - I tested this new configuration by running `pip install /path/to/local/clone/of/nmdc-schema` and confirming the following `import` statement worked:
      ```py
      from nmdc_schema.migrators.migrator_from_8_1_to_9_0 import Migrator_from_8_1_to_9_0 as Migrator
      ```
     - I did **not** test it on PyPI (I did not publish anything to PyPI)
- Fixed bug where, in general, a YAML file could not be loaded except for in the `nmdc-schema` development environment
    - Bug: https://github.com/microbiomedata/nmdc-schema/issues/1310


eecavanna:
I finished refactoring the migration-related classes into a subfolder. 

✅ I confirmed that the `migration_recursion.py` script still runs without error. For example, this runs without error:
```shell
poetry run migration-recursion \
  --salvage-prefix nmdc \
  --output-path /tmp/output.yaml \
  --schema-path nmdc_schema/nmdc_schema_accepting_legacy_ids.yaml \
  --input-path local/mongo_as_unvalidated_nmdc_database.yaml \
  -v DEBUG \
  --migrator-name Migrator_from_8_1_to_9_0
```

🔥 However, since making these changes, I have **not** been able to `import` any of the migration-related classes (no longer defined in `migration_recursion.py`) from a Python notebook in `nmdc-runtime` (from which I had been able to import the classes before making these changes). 

I have tried several things in an attempt to be able to `import` them, but none of the things I tried worked. You can see the things I tried in the commits that begin with `WIP:` (all of which have since been reverted).

Note: Since making these changes, I have only tried installing the package via **GitHub** (via the command below). Maybe the `import` issue is related to the fact that the code in this branch is "source" code, as opposed to something "built" by Poetry.

```shell
pip install git+https://github.com/microbiomedata/nmdc-schema.git@1250-move-each-migration-class-into-its-own-module-in-a-migrations-folder
```

eecavanna:
In case this PR is still open on Tuesday, I can present the changes during the migration squad meeting that day.

brynnz22 (commented):
@eecavanna this looks great to me! A lot easier to digest and understand than one big file. I'll let @turbomam approve this since he knows the structure of the repo much better than me. Is there a way we can include dumping the output of a migration(s) into a yaml file as we did with the `migration_recursion.py` file so that we can test the migrations while writing them?

eecavanna:
Thanks, @brynnz22! The `migration_recursion.py` script still behaves the same way (from the outside) that it does in the current `main` branch; i.e. it can still be used to migrate data residing in a YAML file. Does that answer your question?

---

On a side note about testing, after this branch has been merged in, I'm planning to add some `doctests` to the transformation functions (https://github.com/microbiomedata/nmdc-schema/issues/1291). I'll add a note to tomorrow's squad meeting agenda to demonstrate doctests.

eecavanna:
@eecavanna : Emphasize in click help that the class name is NOT the module name / filename.

@eecavanna : Emphasize X.Y as default instead of X.Y.Z for migrations.

eecavanna:
I updated the `click` command help. Here's a snippet from the latest help:

```plain
  --migrator-name TEXT   The name of a migrator class (i.e. a class that
                         inherits from `MigratorBase`) that you want to use to
                         migrate data.
                         
                         Those classes are defined in `nmdc_schema/migrators/`
                         and their names begin with `Migrator_`. If you
                         specify this option multiple times, the specified
                         classes will be used in the order specified.

```

eecavanna:
Hi @turbomam, I finished making one of the [changes](https://github.com/microbiomedata/nmdc-schema/pull/1340#issuecomment-1811127872) we discussed in today's pair review. I want to hold off on the other one until a future PR, where I make other changes to the documentation and example migrator class.

turbomam (approved):

PR #1352, opened 2023-11-14
1337 add slot bulk elect conductivity

bmeluch:
Added "bulk_elect_conductivity" slot definition to nmdc.yaml, also added it to the list of slots in the Biosample class definition. This slot is being added for EMSL metadata compatibility.


src/schema/nmdc.yaml
```diff
@@ -2873,3 +2874,16 @@ slots:
     contributors:
       - ORCID:0009-0001-1555-1601 #Anastasiya Prymolenna
       - ORCID:0000-0002-8683-0050 #Montana Smith
+
+  bulk_elect_conductivity:
+    comments:
+      - Compatible with EMSL metadata
```
turbomam, line 2880: what does this mean? are you implying that any slot that doesn't have this comment isn't compatible with EMSL metadata? What **is** EMSL metadata anyway?

src/schema/nmdc.yaml
```diff
@@ -2873,3 +2874,16 @@ slots:
...
+    description: 
+      Electrical conductivity is a measure of the bulk soil ability to carry electric 
+      current, which is mostly dictated by the chemistry of and amount of soil water. 
+      Provide the value output of the field instrument.
```
turbomam, line 2884: well written `description`. are we sure that `bulk_elect_conductivity` only applies to soil samples?

instructions to users shouldn't be included in `description`.  `comments` would probably be better.

src/schema/nmdc.yaml
```diff
@@ -2873,3 +2874,16 @@ slots:
...
+    title: bulk electrical conductivity
+    examples:
+      - value: 0.017 mS/cm
```
turbomam, line 2887: @brynnz22 has been writing the examples for slots in their structured formats, as opposed to flattened strings. This flattened style is fine, but please check in with @brynnz22 if you get a chance.

src/schema/nmdc.yaml
```diff
@@ -1081,6 +1081,7 @@ classes:
       - sample_link
 
       ## EMSL slots
+      - bulk_elect_conductivity
```
turbomam, line 1084: will you be moving all existing EMSL slots underneath this comment?

src/schema/nmdc.yaml
```diff
@@ -2873,3 +2874,16 @@ slots:
...
+    comments:
+      - Compatible with EMSL metadata
+    description: 
+      Electrical conductivity is a measure of the bulk soil ability to carry electric 
+      current, which is mostly dictated by the chemistry of and amount of soil water. 
+      Provide the value output of the field instrument.
+    title: bulk electrical conductivity
+    examples:
+      - value: 0.017 mS/cm
+    is_a: attribute
```
turbomam, line 2888: i think that we have been inconsistent in saying that `Biosample` slots are `attribute`s. THis is fine, but we should start a new issue do do it more consistently.

turbomam (changes requested):
Thanks, @bmeluch . The is a good PR. Please revise it based on my comments, or explain why you don't think they're necessary (or could be dealt with at a future time)

src/schema/nmdc.yaml
brynnz22: @bmeluch here is a link the linkml info for the `range` of `examples` https://linkml.io/linkml-model/latest/docs/Example/. So you can include the `value` and the `description` too. Here's an example of multiple examples for `associated_dois`: https://github.com/microbiomedata/nmdc-schema/blob/7097fc7a236628cdef21978889b0f685babdd388/src/schema/nmdc.yaml#L2749. Also, I saw that the linkml defintion says `value_description`, but that doesn't work, its just `description` for some reason.

turbomam: Thanks @brynnz22 !  The `value_description` vs `description` is an annoying case of slot name mangling. I'm glad you caught that. I'm going to raise an issue about that in the LinkML repo. It's been discussed a lot so maybe it will be left like that for the time being.

turbomam: - https://github.com/linkml/linkml/issues/1733

src/schema/nmdc.yaml
mslarae13: This comment can go away. We don't need to include that. EMSL metadata just means it's in the EMSL metadata template but isn't in NMDC. This is a MONet measurement & is needed to support user facility operations.

src/schema/nmdc.yaml
mslarae13: I am not sure! Let's take out soil and just say measure of the sample's ability

src/schema/nmdc.yaml
mslarae13: I think it's useful to organize, so we know where it came from. The new slots will go here.. most of the slots are MIxS & already exist

src/schema/nmdc.yaml
bmeluch: Removed comment

src/schema/nmdc.yaml
bmeluch: Changed language to be general for sample type and moved instruction sentence to comments

src/schema/nmdc.yaml
bmeluch: I think I have broken out the example value into its structured format. Thank you @brynnz22 !

src/schema/nmdc.yaml
bmeluch: I removed "is_a: attribute" since it doesn't seem to be used consistently

mslarae13 (approved):

PR #1357, opened 2023-11-16
Add `doctest`s to example transformation function

eecavanna:
### Summary of changes

- Added `doctest`s to the example transformation function in the tutorial class
- Added details to the "Creating a migrator" instructions


turbomam (approved):
@cmungall was very happy that you have added this practice @eecavanna

PR #1358, opened 2023-11-16
Fix URL in schema `id`

### LINKED ISSUE #1313: Broken schema URI link on documentation landing page
URI: https://w3id.org/nmdc/nmdc

sujaypatil96:
Fixes #1313


turbomam (changes requested):
I disagree with this edit. https://w3id.org/nmdc/ is not an identifier of this module, it is an identifier of the entire schema (and resolves to the documentation page)

Maybe we need t make some changes to the w3id.org/nmdc redirection? Although we ahven't done that for any other module.

Issue #1368, opened 2023-11-17
change MIxS import to 6.2


turbomam:
possible duplicate issue

specified as a hard-coded variable in nmdc_schema/generate_import_slots_regardless.py

currently https://raw.githubusercontent.com/microbiomedata/mixs/1da849346a80b717810a02d7c8ed74a22bcd84de/model/schema/mixs.yaml

change to https://raw.githubusercontent.com/GenomicsStandardsConsortium/mixs/v6.2.0/src/mixs/schema/mixs.yaml

turbomam:
trigger with `make mixs-yaml-clean`

turbomam:
would need multiple source classes

turbomam:
ValueError: No such slot host_family_relation as an attribute of MimsSoil ancestors or as a slot definition in the schema

now [host_fam_rel](https://genomicsstandardsconsortium.github.io/mixs/0000872/)

turbomam:
- ValueError: No such slot `salinity_meth` as an attribute of MimsSoil ancestors or as a slot definition in the schema
- not used in MongoDB yet
- is mentioned in [sheets_and_friends/tsv_in/import_slots_regardless.tsv](https://github.com/microbiomedata/submission-schema/blob/0d7118ece7fe739093dd7d790fdb1d2b28e91e21/sheets_and_friends/tsv_in/import_slots_regardless.tsv)
- no replacement currently available

turbomam:
- ValueError: No such slot `samp_collec_device` as an attribute of MimsSoil ancestors or as a slot definition in the schema

turbomam:
- ValueError: No such slot `samp_collec_method` as an attribute of MimsSoil ancestors or as a slot definition in the schema
- not used in MongoDB yet
- appears in assets/misc/neon_nmdc_term_mapping.tsv
- replaced with `samp_collect_method`

turbomam:
- ValueError: No such slot `soil_text_measure` as an attribute of MimsSoil ancestors or as a slot definition in the schema
- not used in MongoDB yet
- replaced with `[soil_texture](https://genomicsstandardsconsortium.github.io/mixs/0000335/)` ???

turbomam:
ValueError: Conflicting URIs (https://raw.githubusercontent.com/microbiomedata/nmdc-schema/main/src/schema/mixs.yaml, https://w3id.org/nmdc/core) for item: sequencing

a subset in both mixs.yaml and core.yaml

[...]

PR #1386, opened 2023-11-20
1368 change mixs import to 62

turbomam:
- #1368


turbomam:
Hi @pkalita-lbl and @mslarae13 

This is a version of nmdc-schema that imports slots form GSC's MIxS 6.2, as opposed to the static, GSC v6.0-like https://github.com/microbiomedata/mixs/blob/main/model/schema/mixs.yaml

I have extensive notes in the accompanying issue #1386 but they may not be crystal clear yet

The thing I want to emphasize here is that I have removed all `yq` customization of MIxS slots except for the ones that are used in MongoDB as of now. So that means some MIxS slots may be presented to submission-schema now in a slightly different way, so we may need to make minor changes to the `sheets_and_friends` configuration.

pkalita-lbl:
I took a look at the changes here and read through the comments in https://github.com/microbiomedata/nmdc-schema/issues/1368. I'm not sure I follow every single detail, but that's probably okay. In broad strokes, the goals make sense to me. If there's something you want me to do re: the submission schema before this is merged and/or released, let me know.

mslarae13:
Overall, I'm on the same page as @pkalita-lbl . I understand the want and I support it! I think having this static MIxS file is a source of incompatibility and error.

Issue #1314, opened 2023-11-08
Proposed deprecation life-cycle


turbomam:
## When any element is to be deprecated
The following annotations should be added or updated
- [deprecated_element_has_exact_replacement](https://linkml.io/linkml-model/latest/docs/deprecated_element_has_exact_replacement/), [deprecated_element_has_possible_replacement](deprecated_element_has_possible_replacement) and/or [deprecated](https://linkml.io/linkml-model/latest/docs/deprecated/), in that order of preference:
- [last_updated_on](https://linkml.io/linkml-model/latest/docs/last_updated_on/)
- [modified_by](https://linkml.io/linkml-model/latest/docs/modified_by/), with an ORCID value

## When a  class is to be deprecated
- That class should not be in the domain or range of any slot
- No class should claim the deprecated class as its `is_a` parent or as a mixin

## When a  slot is to be deprecated
That slot should be removed from all classes

## Deprecated elements should stay in the schema for one release
The annotations listed above will aid in deleting the deprecated elements after one release cycle.

turbomam:
Here are the last merged PRs where "deprec" can be found on the modified files page 

```
v9.0.4
 https://github.com/microbiomedata/nmdc-schema/pull/1282/files
  ess_dive_datasets:
    multivalued: true
    # this is moot now that it's deprecated... but what was this slot's range? and ws it really intended to be inlined_as_list?
    #    inlined_as_list: true
    description: List of ESS-DIVE dataset DOIs
    deprecated_element_has_exact_replacement: nmdc:associated_dois
  has_credit_associations:

v9.0.0
  https://github.com/microbiomedata/nmdc-schema/pull/1208/files (several)
```

Since we're on v9.1.0 now, I propose that we delete all deprecated elements back to [v9.0.0](https://github.com/microbiomedata/nmdc-schema/releases/tag/v9.0.0) now

opinions @brynnz22 @cmungall ?

brynnz22:
@turbomam This sounds good to me. When I have been deprecating slots I have been commenting out the slot form the class so the schema still shows where it was used for a release cycle. I am thinking we should continue doing this and then remove it completely for the next release so the schema still retains all the information about the slot and where it was used until it is completely removed.

brynnz22:
@turbomam I am happy to work on the documentation for this. I had this on my list of things to do this week.

cmungall:
Looks good! Often we also add a reason for deprecation plus link to issue

[...]

PR #1393, opened 2023-11-20
add element deprecation documentation

brynnz22:
Regarding: https://github.com/microbiomedata/nmdc-schema/issues/1314

I added @turbomam words almost exactly from the issue with some minor adjustments.
- I included @cmungall suggestion of including a reason for the deprecation and a link to the corresponding issue. I suggested doing this using the `deprecated` slot (since it takes a `string`) in addition to the replacement slots (they take `uriorcurie`). Let me know if you think this should be done another way.
- I added commenting out all element usages prior to fully removing them, so we can still understand where they are being used for one release cycle in case it becomes important.


src/docs/schema_element_deprecation_guide.md
```diff
@@ -0,0 +1,26 @@
+# Deprecating NMDC schema elements
+
+When deprecating schema elements, it is a two step process:
+
+1. For one release cycle, the schema will include deprecation annotations to the element's definition and commenting out any usages of the slot (e.g. in a class, domain, range, etc) including a deprecation date (e.g. `# doi_awards deprecated on 11/12/2023` in a class's usage)
```
turbomam, line 5: let's encourage our teammates to use iso8601/international dates like 2023-11-12

turbomam (changes requested):
thanks, this is very valuable

exposing new content like this also requires that the filename is added to the `nav` section of `mkdocs.yml`

src/docs/schema_element_deprecation_guide.md
brynnz22: I added this.

turbomam:
I am going to check this out, add `src/docs/schema_element_deprecation_guide.md`  to the `nav` section of `mkdocs.yml`, and then merge.

turbomam (approved):

Issue #1350, opened 2023-11-14
Create no doi pattern in websites slot


brynnz22:
Per the discussion in the Study Refresh Squad, we need to create a pattern in the `websites` slot (used by `principal_investigator`) that does not allow any urls with "doi". 

- [x] create no doi pattern in `websites` slot
- [x] Add a comment that says where is a better place to add a doi (`associated_dois`).

PR #1400, opened 2023-11-20
add regex `pattern` to `websites` slot: DOI-like values not allowed

brynnz22:
regarding: https://github.com/microbiomedata/nmdc-schema/issues/1350

- [x] add regex pattern to `websites` that specifies it should start as `https://` and should not include the string `doi`. 
- [x] add comment to `websites` indicating that `associated_dois` should be used for any DOIs and not `websites`.
- [x] write migration to move websites that dois into the `associated_dois` slot and removed from the `websites` slot
- [x] update `CHANGELOG.md` with the unreleased addition
- [x] Include `websites_dois.yaml` file with the new information about the migrated dois.


turbomam:
Thanks @brynnz22 

The PR checks passed, so your new pattern is compatible with our test data.

I'm going to assume that you haven't run all of our MongoDB through the schema in this branch yet. So I'll check it out and run `make make-rdf`. That's also possible now as a manually launch-able GH action, [migration-with-java-action](https://github.com/microbiomedata/nmdc-schema/actions/workflows/migration.yaml). I'm thinking of renaming it because it applies all of the migrations to a data file, but it **doesn't really do a migration** in any database.

turbomam:
This is a very good PR but it will require a migration. Fortunately that's limited to `Study` instances, and you have good understanding of them and know who to double check with!

From local/mongo_as_nmdc_database_validation.log:

- [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.25982/109073.30/1895615' does not match '^https://(?!.*doi).*$' in /study_set/7/websites/4
- [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.5281/zenodo.8071247' does not match '^https://(?!.*doi).*$' in /study_set/16/websites/1
- [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.5281/zenodo.5816355' does not match '^https://(?!.*doi).*$' in /study_set/16/websites/2
- [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.5281/zenodo.6533064' does not match '^https://(?!.*doi).*$' in /study_set/16/websites/3
- [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.6073/pasta/4d5f03a4619e834c031ab4a6a121de12' does not match '^https://(?!.*doi).*$' in /study_set/17/websites/2

turbomam (changes requested):
good improvement for making sure people understand the distinct intentions for slots that could be confused with each other.

this will require a migration.

brynnz22:
@turbomam I think this could just require a changesheet. There is only one `website` that has a doi in the results. And not websites that do not start with https:// But I can write the migration process too if you would prefer?

turbomam:
Did you mention doing a changesheet instead or writing migration code? I think that would be fine too, but maybe we should get a third opinion from @pkalita-lbl  or @eecavanna

brynnz22:
Yes I did, but I thought it was only one thing that needed to be changed and then I saw your earlier comments and saw that it was several. So I think I agree with you on the migration.

eecavanna:
Regarding the schema change, itself, I'm confused about it. There may be some websites people want to link to that are only accessible via HTTP (as opposed to HTTPS). Also, there may be some websites people want to link to that contain the substring "doi". I am wondering whether there is a different way we can accomplish whatever prompted this schema change, that doesn't have those implications.

brynnz22:
@eecavanna I can change the regex pattern to accept https or http? As fars as dois, we don't want them in this `websites` slot because we have a slot specifically for DOIs. But right now, only the `Study` class uses it.

pkalita-lbl:
Fully agree with Eric's concerns about the pattern itself.

Regarding changesheets vs migrations I would say always prefer migrations for changes that are in support of a new schema version. A changesheet is run once but you have no idea what new data might come in between then and when the runtime adopts the new version. Migrations vastly reduce that time of uncertainty.

eecavanna:
Regarding the necessity of a migration, here's how I think of it:

```mermaid
flowchart LR
    A["Draft a schema change"] --> C
    C{Could that<br>schema change<br>make existing<br>data invalid?}
    C -->|Yes| D[Implement a migrator]
    C -->|No| E[Done]
  
```

brynnz22:
@pkalita-lbl @eecavanna please see my comment above. Would that address your concerns?

eecavanna:
Hi @brynnz22, thanks for explaining what prompted the schema change.

I'm in favor of allowing "http", "HTTP", "https", and "HTTPS" (I also wouldn't object to also allowing "Http", "HtTp", or any other letter casing). Either of those would address my concerns about HTTP/HTTPS.

Regarding rejecting DOIs, I'd prefer the rejection rule be more strict in terms of what it rejects. For example, if DOIs always begin with `https://doi.org/10.` (I don't know whether that's the case—I'll look it up), I'd prefer the regex check for that, specifically, instead of something more general.

brynnz22:
That's a good point. I believe all the old doi formats do start with doi.org. I will add that to that pattern along with the casing you gave

eecavanna:
I tried to find the official spec for DOIs and ran into a paywall! I'll keep searching...

image

eecavanna:
This looks good to me.

image

Thanks for making these changes, @brynnz22 !

turbomam:
@brynnz22 I see you have added an asset to inform the necessary migration, but I don't see any new migration methods yet. Are those still pending?

Do you fell comfortable running something like `make squeaky-clean all make-rdf` now to see how the data in MongoDB validate against the new patterns?

src/schema/core.yaml
```diff
@@ -674,7 +674,9 @@ slots:
   websites:
     range: string
     multivalued: true
+    pattern: ^[Hh][Tt][Tt][Pp][Ss]?:\/\/(?!.*[Dd][Oo][Ii]\.[Oo][Rr][Gg]).*$
     description: A list of websites that are associated with the entity.
+    comments: DOIs should be included as websites. Instead, use the associated_dois slot.
```
eecavanna, line 679: Possible typo:

```diff
-    comments: DOIs should be included as websites. Instead, use the associated_dois slot.
+    comments: DOIs should not be included as websites. Instead, use the associated_dois slot.
```

brynnz22: Ah thanks for the catch!

brynnz22:
Hey @turbomam yes, that migration is still pending. I am almost finished. I just need to add the doc string. I will run the make `squeaky-clean all make-rdf` soon.

eecavanna:
For future reference, migrator methods can print things via `self.logger`. For example:

https://github.com/microbiomedata/nmdc-schema/blob/bc3edda1bb6f7ac1e798a4c6287850e7ff081b5d/nmdc_schema/migrators/migrator_from_8_0_to_8_1.py#L29

nmdc_schema/migrators/migrator_from_9_1_to_9_2.py
```diff
@@ -0,0 +1,43 @@
+from nmdc_schema.migrators.migrator_base import MigratorBase
+from nmdc_schema.migrators.helpers import load_yaml_asset
+
+class Migrator_from_9_1_to_9_2(MigratorBase):
+    """previously: Migrates data from schema 9.1.0 to 9.2.0"""
```
eecavanna, line 5: I think the `previously: ` prefix is something @turbomam added to a couple existing classes' docstrings when he removed the "patch" number (i.e. `A_B_C` --> `A_B`) from their names. I'd prefer it be omitted here.

```diff
- """previously: Migrates data from schema 9.1.0 to 9.2.0"""
+ """Migrates data from schema 9.1.0 to 9.2.0"""
```

brynnz22:
@turbomam @eecavanna this is ready for review. It passed the `squeaky-clean all make-rdf` when I ran it. Thank you!

eecavanna (approved):
LGTM! Thanks for including a doctest.

turbomam:
@brynnz22 

I just ran `make make-rdf` and there are still some values that need to be migrated:

> [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.25982/109073.30/1895615' does not match '^[Hh][Tt][Tt][Pp][Ss]?:\\/\\/(?!.*[Dd][Oo][Ii]\\.[Oo][Rr][Gg]).*$' in /study_set/7/websites/4
> [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.5281/zenodo.8071247' does not match '^[Hh][Tt][Tt][Pp][Ss]?:\\/\\/(?!.*[Dd][Oo][Ii]\\.[Oo][Rr][Gg]).*$' in /study_set/16/websites/1
> [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.5281/zenodo.5816355' does not match '^[Hh][Tt][Tt][Pp][Ss]?:\\/\\/(?!.*[Dd][Oo][Ii]\\.[Oo][Rr][Gg]).*$' in /study_set/16/websites/2
> [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.5281/zenodo.6533064' does not match '^[Hh][Tt][Tt][Pp][Ss]?:\\/\\/(?!.*[Dd][Oo][Ii]\\.[Oo][Rr][Gg]).*$' in /study_set/16/websites/3
> [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.6073/pasta/4d5f03a4619e834c031ab4a6a121de12' does not match '^[Hh][Tt][Tt][Pp][Ss]?:\\/\\/(?!.*[Dd][Oo][Ii]\\.[Oo][Rr][Gg]).*$' in /study_set/17/websites/2

turbomam (changes requested):
@brynnz22 

I just ran `make make-rdf` and there are still some values that need to be migrated:

> ```shell
> [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.25982/109073.30/1895615' does not match '^[Hh][Tt][Tt][Pp][Ss]?:\\/\\/(?!.*[Dd][Oo][Ii]\\.[Oo][Rr][Gg]).*$' in /study_set/7/websites/4
> [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.5281/zenodo.8071247' does not match '^[Hh][Tt][Tt][Pp][Ss]?:\\/\\/(?!.*[Dd][Oo][Ii]\\.[Oo][Rr][Gg]).*$' in /study_set/16/websites/1
> [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.5281/zenodo.5816355' does not match '^[Hh][Tt][Tt][Pp][Ss]?:\\/\\/(?!.*[Dd][Oo][Ii]\\.[Oo][Rr][Gg]).*$' in /study_set/16/websites/2
> [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.5281/zenodo.6533064' does not match '^[Hh][Tt][Tt][Pp][Ss]?:\\/\\/(?!.*[Dd][Oo][Ii]\\.[Oo][Rr][Gg]).*$' in /study_set/16/websites/3
> [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.6073/pasta/4d5f03a4619e834c031ab4a6a121de12' does not match '^[Hh][Tt][Tt][Pp][Ss]?:\\/\\/(?!.*[Dd][Oo][Ii]\\.[Oo][Rr][Gg]).*$' in /study_set/17/websites/2
> ```

---

_Edit: @eecavanna added "code fences" around the error message snippet. He then put the fenced code into a block quote._

eecavanna:
@turbomam will you update the most recent comment so the error messages are within "code fences" (i.e. triple backtick, newline, ...error messages..., newline, triple backtick)?

The error messages contain some characters that GitHub doesn't render (I assume GitHub's Markdown renderer is treating them as formatting characters):

image

brynnz22:
Hey @turbomam this worked on friday. I am guessing that the `project.Makefile` needs to be updated so that the `--migrator-name` is specified. Do you want to push that as part of this PR? Or would you rather specify it yourself when testing in the local?

turbomam:
I'll try adding that migrator. Thanks. I think that should be the standard practice: PRs that require a migration should include all necessary migrator classes as `--migrator-name` arguments, and no unnecessary `--migrator-name`s

turbomam:
> @turbomam will you update the most recent comment so the error messages are within "[code fences](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#fenced-code-blocks)"

@eecavanna I put code in code fences and output in code blocks, using `>`

turbomam:
> [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.25982/109073.30/1895615' does not match '^[Hh][Tt][Tt][Pp][Ss]?:\\/\\/(?!.*[Dd][Oo][Ii]\\.[Oo][Rr][Gg]).*$' in /study_set/7/websites/4
> [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.5281/zenodo.8071247' does not match '^[Hh][Tt][Tt][Pp][Ss]?:\\/\\/(?!.*[Dd][Oo][Ii]\\.[Oo][Rr][Gg]).*$' in /study_set/16/websites/1
> [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.5281/zenodo.5816355' does not match '^[Hh][Tt][Tt][Pp][Ss]?:\\/\\/(?!.*[Dd][Oo][Ii]\\.[Oo][Rr][Gg]).*$' in /study_set/16/websites/2
> [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.5281/zenodo.6533064' does not match '^[Hh][Tt][Tt][Pp][Ss]?:\\/\\/(?!.*[Dd][Oo][Ii]\\.[Oo][Rr][Gg]).*$' in /study_set/16/websites/3
> [ERROR] [local/mongo_as_nmdc_database_rdf_safe.yaml/0] 'https://doi.org/10.6073/pasta/4d5f03a4619e834c031ab4a6a121de12' does not match '^[Hh][Tt][Tt][Pp][Ss]?:\\/\\/(?!.*[Dd][Oo][Ii]\\.[Oo][Rr][Gg]).*$' in /study_set/17/websites/2

eecavanna:
Hi @turbomam, in the comment, the output was in a quote block (`>`) as opposed to a code block (triple backtick). You can combine the two, like this (example "C" below):

A:

```
Code
```

B:

> Quote

C:

> ```
> Quoted code
> ```

eecavanna:
>Do you remember what characters weren't rendered properly?

It was an asterisk (the original rendered HTML result is in the screenshot in this [comment](https://github.com/microbiomedata/nmdc-schema/pull/1400#issuecomment-1830977558)).

project.Makefile
```diff
@@ -466,6 +466,7 @@ local/mongo_as_unvalidated_nmdc_database.yaml:
 local/mongo_as_nmdc_database_rdf_safe.yaml: nmdc_schema/nmdc_schema_accepting_legacy_ids.yaml local/mongo_as_unvalidated_nmdc_database.yaml
 	date # 449.56 seconds on 2023-08-30 without functional_annotation_agg or metaproteomics_analysis_activity_set
 	time $(RUN) migration-recursion \
+		--migrator-name Migrator_from_9_1_to_9_2 \
```
turbomam, line 469: thanks @brynnz22

turbomam (approved):

project.Makefile
brynnz22: Sounds good. I'll do this in the future. Thanks!

PR #1410, opened 2023-11-22
Redirect `/home` to `/` on documentation website (mkdocs)

eecavanna:
### Changes

- Added `mkdocs-redirects` (https://github.com/mkdocs/mkdocs-redirects) as a "dev" dependency
- Defined a single redirect, which makes it so `/home` redirects to `/`


turbomam (approved):

eecavanna:
Merged. Thanks for reviewing it so promptly!

Issue #1411, opened 2023-11-22
reduce the direct subdirectories of src/data to `valid`, `invalid` and `problem`


turbomam:
part of the solution is to convert the files currently in the `uncertain` and `unexpected-pass` directories into truly valid and truly invalid files and place them into the `valid` or `invalid` directories

turbomam:
in fact, there should be any long-lived `problem/` example data files. files that either pass or fail in an unexpected way should be moved into single-issue repos.

PR #1413, opened 2023-11-22
no special polymorphic dir or target

turbomam:
- #1411 (partially)


turbomam:
- also removed some commented lines from project.Makefile
- accidentally pushed auto-generated project/ files

PR #1416, opened 2023-11-22
Remove stray white space in MixingProcess description.

JamesTessmer:
This should be done now, quickly removed it in the online editory.


turbomam (approved):
Thanks. I _think_ `linkml-lint` would have caught that. We should all be in the habit of running that more routinely. By default it will report a lots of problems. We could write a custom conflagration or even our won rules.

https://linkml.io/linkml/schemas/linter.html

Issue #1421, opened 2023-11-27
temporarily add `host_disease_stat` to src/schema/portal/mixs_inspired.yaml


turbomam:
while waiting for this to merge:
- https://github.com/microbiomedata/nmdc-schema/pull/1386

see also:
- https://github.com/microbiomedata/nmdc-schema/issues/1336

PR #1425, opened 2023-11-27
host_disease_stat workarround plus PATO exp

turbomam:
- #1421


turbomam:
I made this so complicated. I think the old mostly-dynamic method for importing terms into our lock mixs.yaml was working. I just forgot to commit the generated mixs.yaml.

Issue #1427, opened 2023-11-28
try to fix QC  Report PR 1334 in new branch


turbomam:
- https://github.com/microbiomedata/nmdc-schema/pull/1334

cc @aclum

Issue #1333, opened 2023-11-10
expanded use of quality_control_report


aclum:
We have a need to indicate pass/fail for some of the WorkflowExecutionActivity subclasses. We already made quality_control_report but that currently only has Extraction as an applicable class. I would like to expand this slot to other PlannedProcess and WorkflowExecutionActivity classes. 

I would also like to expand the slots on Class [QualityControlReport](https://microbiomedata.github.io/nmdc-schema/QualityControlReport/). The current slots are 'status' and 'name'. JGI has a concept of failure_what failure_where, populated with controlled vocabulary. 
ie 
failure_where failure_what
Sequencing	substandard run output
PMO                 assembly size below min threshold
multiple_failure_modes Sequencing_qPCR error;Sequencing_no data produced

We could model something similar where failure_where has a range of Activity or PlannedProcess classes (soon to be only PlannedProcess after the Dec schema refactor)

I'll open a draft pull request with some example data.

aclum:
FYI @Michal-Babins @hubin-keio @mslarae13 @anastasiyaprymolenna @scanon @turbomam @sujaypatil96

ssarrafan:
@aclum appears to be active. I'll move to the next sprint but please remove if it's not being worked on.

ssarrafan:
Last day of sprint. Removing from sprint, adding backlog label. Please update or add to next sprint if still active.

aclum:
This is in progress this week. Ive updated the modeling and need help from Brynn on the linkml rules, then the migration code needs to be written. This will likely span into January.

aclum:
Brynn and Patrick confirmed there is a bug in the linkml rules, see https://github.com/linkml/linkml/issues/1524. Patrick said he'd try and work on this on Friday. Will pause on this until January. Targeting a solution by the end of the first sprint in January so this can make it into the January end of month release.

aclum:
To get the value_presence rules to work correctly I had to update linkml to 1.6.6 on Patrick's recommendation, then Patrick helped identify an issue where rules don't inherit to subclasses. Patrick will make a ticket for the linkml issue, this should be released in early January. Workaround would be to copy rules to each subclass but leaving for now since the next update to runtime would be end of January.

aclum:
Patrick fixed the rules in linkml, i've updated the schema to use linkml 1.6.8 which made the linkml rules inheritance work as expected. Now the build passes.

[...]

PR #1428, opened 2023-11-28
QC Report branch with test refinement

turbomam:
- #1427 
- #1333


turbomam:
@brynnz22 @aclum can we just remove the rule and the invalid examples? It looks like it will build then.

aclum:
@turbomam  I will test that now.

aclum:
@turbomam I need some help getting the mixs.yaml conflicts resolved.

turbomam:
You can delete it from the branch. It gets regenerated by make all, if necessary.

Or I can do it in about two hours.

aclum:
@turbomam  the build is failing b/c the example valid data is missing has_output. I thought you were going to add other tests/change how the model was done so this data would pass.

turbomam:
oh yeah. A reasonable solution would be to assert  `recommended: true` instead of `required: true` on `has_output`'s `slot_usage` in `WorkflowExecutionActivity`. I would especially want @scanon or someone else he designates to sign off on that change.

aclum:
@brynnz22 @turbomam  doing the rule the other way is a lot less desirable since qc_status is not currently required. I would much prefer if we could get the linkml rules working properly. @cmungall  can this be a prioritized activity?

aclum:
To get the value_presence rules to work correctly I had to update linkml to 1.6.6, then @pkalita-lbl helped identify an issue where rules don't inherit to subclasses. Patrick will make a ticket for the linkml issue, this should be released in early January. Workaround would be to copy rules to each subclass.  cc @turbomam @brynnz22

pkalita-lbl:
LinkML issue: https://github.com/linkml/linkml/issues/1802

aclum (approved):

src/schema/nmdc.yaml
```diff
@@ -223,7 +226,7 @@ classes:
       - extraction_method
       - extraction_target
       - input_mass
-      - quality_control_report
+      - volume
```
turbomam, line 229: this doesn't' really have anything to do with QC reports does it? I wonder if I let it sneak in?

src/schema/nmdc.yaml
```diff
@@ -2247,6 +2247,44 @@ enums:
       V-bottom conical tube:
       falcon tube:
 
+  FailureWhatEnum:
```
turbomam, line 2250: I think we should make a note somewhere about these permissible values that share names with classes. It's especially interesting in the OWL serialization, becasue the bases of the URIs are quite different

src/schema/workflow_execution_activity.yaml
```diff
@@ -44,7 +45,47 @@ classes:
...
+        preconditions:
+          slot_conditions:
+            qc_status:
+              value_presence: ABSENT
+        postconditions:
+          slot_conditions:
+            has_output:
+              required: true
+
+      # - title: qc_status_pass_qc_where_what_slots_not_specified
```
turbomam, line 75: are we going to leave this in as commented-out code?

src/schema/workflow_execution_activity.yaml
```diff
@@ -44,7 +45,47 @@ classes:
       - part_of
       - type
       - version
-
+      - qc_status
+      - qc_comment
+      - has_failure_categorization
+    rules:
```
turbomam, line 51: I haven't scrutinized these. if there's example data to exercise all of the conditions that the rules know about, and those examples pass or fail as excepted, then they are OK with me.

src/schema/workflow_execution_activity.yaml
```diff
@@ -65,6 +106,10 @@ classes:
         structured_pattern:
           syntax: "{id_nmdc_prefix}:wf-{id_shoulder}-{id_blade}{id_version}{id_locus}"
           interpolated: true
+      qc_failure_what:
```
turbomam, line 109: is `multivalued` being asserted for these slots in a `slot_usage`? It's almost always better to assert that in the global definition of the slot. And in general, we wouldn't want a slot to be multi-valued in some cases and single valued in others.

src/schema/nmdc.yaml
aclum: @turbomam  I think this is suppose to be here. You can see commits to main from a month or two ago from @anastasiyaprymolenna  and you. See https://github.com/microbiomedata/nmdc-schema/commit/4448e41336bede06541bd3c3d2457aec6314f6ab
the slot for Class Extraction seems to have fallen out but the slot usage is there so I think this slot should be on this class. @anastasiyaprymolenna  please confirm.

src/schema/workflow_execution_activity.yaml
aclum: Deleted

Issue #1439, opened 2023-11-29
remove or refactor `nucleic_acid_concentration` slot


turbomam:
It's not used in MongoDB. It's range is asserted to be `InstrumentValue`, which doesn't make any sense. `InstrumentValue` is also going to be removed.

It was a `see_also` for the following slots
- `dna_concentration` _in `src/schema/portal/jgi_metagenomics.yaml`_
- `rna_concentration` _in `src/schema/portal/jgi_metatranscriptomics.yaml`_

It was associated with class `ProcessedSample` _in `src/schema/core.yaml`_

It is commented out as a slot on a ProcessedSample in the following. In these cases, it illustrates a hybrid `QuantityValue`/`InstrumentValue` structure
- _src/data/problem/uncertain/Database-analytical_sample-extract-exhaustive.yaml_
- _src/data/valid/Database-processed_sample-extract-exhaustive.yaml_

`assets/misc/neon_nmdc_term_mapping.tsv` also suggests that it be used in the `QuantityValue`/`InstrumentValue` style in support of NEON's `nucleicAcidConcentration` and `nucleicAcidQuantMethod`

cc @aclum @mslarae13

mslarae13:
So we did this before we had 'JGI MT Interface' and 'JGI MG Interface' in the submission schema. I agree, we make nucleic_acid_conc (and do the same for all the other rna & dna shared slots) .... and then have slot usage for the MT and MG interfaces as necessary. 

@pkalita-lbl  fyi

PR #1441, opened 2023-11-29
remove `nucleic_acid_concentration` from schema and examples

turbomam:
- #1439

### DIFF
--- a/src/schema/basic_slots.yaml
+++ b/src/schema/basic_slots.yaml
@@ -134,10 +134,6 @@ slots:
   protocol_link:
     range: Protocol
 
-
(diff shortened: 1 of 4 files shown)


aclum:
This was added in support of NEON because 1) represents a more generic slot that the corresponding dna_concentration and rna_concentration slots and 2) there was a desire to couple a measurement with what device generated that measurement. This didn't end up getting implemented in the ETL script so perhaps a moot point. Even if we drop the second use case it is worth keeping the more generic slot.

turbomam:
Sure. I have no objection to leaving it in. We just need to rewrite the slot definition and re-associate it with the classes that should use it.

Issue #1451, opened 2023-11-29
restore plate/well example data files and make problem examples either valid or invalid


turbomam:
`Database-sample-prep.yaml` will remain in `src/data/problem/`

PR #1452, opened 2023-11-29
problem examples become valid or invalid

turbomam:
- #1451


src/data/invalid/Database-nom_analysis_activity_set-non-string-ended_at_time.yaml
```diff
@@ -9,6 +9,6 @@ nom_analysis_activity_set:
       - nmdc:3
       - nmdc:4
     type: xxx
-    started_at_time: 2018-11-13T20:20:39+00:00
-    ended_at_time: "2018-11-13T20:20:39+00:00"
+    started_at_time: "2018-11-13T20:20:39+00:00"
+    ended_at_time: 2018-11-13T20:20:39+00:00
```
mslarae13, line 13: Why did the " get removed from ended at time but added to started at time?

mslarae13 (commented):
I think this is good. Just had the 1 question

PR #1468, opened 2023-12-01
Update Extraction class and add valid data examples for review

JamesTessmer:
PR for extraction class updates.
From 12/1 meeting
New slots added and some slots moved up to planned process


kheal:
**extractant** slot needs to be able to be able to handle multiple solvents (and solutes) and specify the ratio in which they are supplied.  A possible solution would be the range of the extractant be a Solution class that could be reused for other processes rather than the existing enum. A Solution class will be essential for liquid chromatography related processes down the line.

kheal:
propose to rename **sample_volume** slot on Extraction class to **input_volume** to be consistent with input_mass

kheal:
propose to rename **volume** slot on Extraction class to **extractant_volume** for readability and to differentiate from sample/input_volume

turbomam:
Thanks for the feedback @kheal 

One goal for the schema is to keep the number of classes and slots small. Another is to clearly expose in the documentation which elements have similarities without requiring a text search.

I'm inclined to agree with your suggested slot name changes (which increases the number of slots), but then they should either be asserted as `is_a` children of a upper `volume` slot, or they should be put into a `volume_slots` subset.

I think we already discussed the idea of parent slots in some prior Metabolink meeting, but I don't think anything ever got merged.

kheal:
I don't think we need the **concentration** slot on Extraction class since this is a simple combination of **input_mass** or **sample_volume** and **volume** and adds no additional information but creates the opportunity for inconsistency.

corilo:
In the context of what we need for NMDC to 1) correctly classify NOM data in the data portal and 2) enable the metadata-informed workflows, which aim to predict the molecular search space, the fundamental slots we need for the Extraction class are extraction_method, extraction_target, extractants, concentration. We should focus on these slots and consider the others mentioned later.

turbomam (approved):
I approve this PR, but I made the last edits, so I would like to see @anastasiyaprymolenna's approval before it get's merged.

anastasiyaprymolenna (approved):
This all looks in order. I approve the merge

PR #1479, opened 2023-12-04
1432 expand raw data types to support emsl workflows

aclum:


SamuelPurvine (approved):

kheal (approved):

turbomam:
Even for small, incremental PRs like this, I would prefer that we looked to see if there were any invalid data examples that use the elements that were modified by the PR. If yes, then no further action is required. If not, we should add an invalid data examples. 

For example, something based on `src/data/valid/DataObject-mass_spec.yaml` but with

`data_object_type: LC-MS/MS Raw Data, data dependent acquisition`

PR #1482, opened 2023-12-05
create ChromatographicSeparationProcess class

anastasiyaprymolenna:
Added a PhaseSeparationProcess class and example data


JamesTessmer:
> I do not support removing `required: true` from `has_solution_components`
> 
> `Solution` is the only slot to which `has_solution_components` is assigned.
> 
> How can you instantiate a `Solution` without saying what it's components are?

I agree, it should be required. It's only removed temporarily because if it is required it causes a technical issue with validation.

turbomam:
Wow, fast turnarround @anastasiyaprymolenna

turbomam:
https://github.com/microbiomedata/nmdc-schema/actions/runs/7145238321/job/19460511341?pr=1482#step:7:622

turbomam:
> `ValueError: File "nmdc.yaml", line 1668, col 9 Class "ChromatographicSeparationProcess" - unknown slot: "ordered_mobile_phase"`

turbomam (approved):
Thanks @anastasiyaprymolenna and team

PR #1539, opened 2023-12-12
1525 add slots to material processing classes that were on omics processing

mslarae13:
nucl_acid_ext exists on OmicsProcessing (DataGeneration) and Biosample

This PR is adding MIXS slot nucl_acid_ext as a narrow_mapping for extraction_method and extraction_target as it's providing BOTH
This PR also adds http://purl.obolibrary.org/obo/NCIT_C177560 as a narrow_mapping for both extraction_method and extraction_target as this is also a more specific way of identifying BOTH.


mslarae13:
PR should have been on fork

PR #1582, opened 2023-12-16
1432 fix file type enum permissible value msn data

aclum:
Update to FIleTypeEnum permissible values.
Python doesn't like forward slashes in permissible values.


SamuelPurvine (commented):
looks fine to me

Issue #1621, opened 2024-01-06
`pure-export` fallback data source may not match primary data source


eecavanna:
The `make` target `local/mongo_as_unvalidated_nmdc_database.yaml` runs a command called `pure-export`, passing it several CLI options, including:
- `--client-base-url https://api.microbiomedata.org`
- `--mongo-host localhost`
- `--mongo-port 27777`

A user could have localhost port `27777` mapped to a different Mongo database than the one used by the web server at `https://api.microbiomedata.org`. In that case, when the `pure-export` command (i.e. whatever runs when that command is issued) uses a Mongo client, the Mongo database it is accessing would differ from the one it would be accessing when using HTTP.

Here's a message the command showed when I ran it:
```yaml
Attempting to get collection stats from functional_annotation_agg
estimated_document_count = 11816893
Attempting to get 200000 documents from nmdcschema/functional_annotation_agg in pages of 200000.
warning: 500 Server Error: Internal Server Error for url: https://api.microbiomedata.org/nmdcschema/functional_annotation_agg?max_page_size=200000
warning: FastAPI request to nmdcschema/functional_annotation_agg appears to have failed. Trying as a PyMongo query.
```

I want to emphasize the final line: `warning: FastAPI request to nmdcschema/functional_annotation_agg appears to have failed. Trying as a PyMongo query.`

This implies to me that the underlying code first tried to fetch data via HTTP (which involved one Mongo database) and then was going to try via a Mongo client (which would involve a **different** Mongo database).

turbomam:
Yeah, pure-export needs a lot of work! I already have a few related issues that I will add soon. @brynnz22 and I have had some  conversations about the `local/mongo_as_unvalidated_nmdc_database.yaml` target.

Maybe there should be subcommands for
- just getting are report about the intersection of the Database slots and the MongoDB collection names, plus the 
- downloading with a pure API approach
- downloading with a pure MongoDB approach

Or do you think that doing away with the hybrid/fallback solution is a bad idea @eecavanna ?

Do you have suggestions for breaking improvements to pure-export into small commits and then prioritizing them? I would love to work on this with you.

We could also include @mbthornton-lbl because he wrote an API-only dumper with fewer configuration options that only returns records (from multiple collections) if they have some path back to a named Study. We worked on that together to launch as a command line tool () and to take some of its configuration from `local/.env`. I would like to either integrate that with `pure-export` or at least to make the launching and configuration more similar.

[...]

PR #1624, opened 2024-01-09
Implement and integrate adapters for data migrations

eecavanna:
### Summary of changes

- Re-architected migration framework so that migrators use adapters to communicate with the underlying data store
- Implemented an adapter that can communicate with a database implemented as a Python dictionary
- Implemented an adapter that can communicate with a MongoDB database
- Updated existing migrators, migrator creation script, and documentation accordingly
- "Regression" ☹️ : Migrators no longer contain a programmatically-consumable data structure that indicates which collections the migrator interacts with

---

An adapter is a Python class in which some predetermined operations (e.g. rename collection, find document) are implemented in a way that is specific to a given type of data store (e.g. Python dictionary, MongoDB database). Adapters allow the same migrators to be used with different data stores.


eecavanna:
FYI - @brynnz22 and @turbomam: this is the (WIP) PR I want to discuss during tomorrow's meeting.

turbomam:
Nice work, @eecavanna

Let's think about how this code implements solutions that might be generalize-able to all LinkML users, and not just to NMDC users interacting with MongoDB.

I'd like for us to 
- feed this work back to the LinkML community if possible,
- always be on the lookout for existing LinkML solutions, so we don't reinvent the wheel

I suspect that [there isn't an existing LinkML solution corresponding to this code](https://linkml.io/linkml/faq/tools.html#can-i-use-linkml-in-conjunction-with-mongodb), but encourage you to become more involved in the LinkML community for the benefit of both NMDC and LinkML.

see also
- [LinkML Developers Meeting rolling agenda](https://docs.google.com/document/d/1OOgBRPYVwcD0hbKc79EOKP750vVCLEKAtbztSMQdC7A/edit#heading=h.6sqkx1xhumse)
- [LinkML Community Meeting rolling agenda](https://docs.google.com/document/d/1MStpDyh9LOZYJTjLtnpOsNYc3HaeU-bz0CHAI9xOjfQ/edit#heading=h.6sqkx1xhumse)

@pkalita-lbl @cmungall do you have anything to add to this?

cmungall:
These are great ideas @turbomam

I suggest though that Eric focuses on the functionality needed for NMDC, we can then look at aligning this with plans for a more generic common virtual database store abstraction in LinkML, see https://github.com/orgs/linkml/discussions/1818

turbomam:
Thanks @cmungall. That's a good path forward.

@eecavanna do you expect that this PR would interfere with my ability to dump database contents with `pure-export` and then validate them in the nmdc-schema environment? Or will this introduce new capabilities that would need to be added back into `pure-export` via an import?

see also
- https://github.com/microbiomedata/nmdc-schema/issues/1621

eecavanna:
Hi @turbomam, thanks for attending part of today's meeting. In response to your questions:

> do you expect that this PR would interfere with my ability to dump database contents with `pure-export` and then validate them in the `nmdc-schema` environment?

No. The changes in this PR only come into play (in the `nmdc-schema` environment) when the `migration_recursion.py` file is processing the YAML-formatted data. As far as I know, that happens downstream from `pure-export`, which would have been used to generate that YAML-formatted data.

Regarding validation, I don't expect these changes to interfere. These changes only impact the data/data structure transformation process.

> Or will this introduce new capabilities that would need to be added back into `pure-export` via an import?

I don't know (I haven't thought about this). I haven't had any other consumers of these changes in mind (other than the `Migrator` classes) as I have been designing and implementing them.

eecavanna:
A note to myself (@eecavanna):

The following are still missing from this branch:
- [ ] A place to register the `Migration` methods that utilize the adapter methods (the `self._agenda` dictionary was not designed for this—although maybe the design could be updated to accommodate it)
- [ ] Determine whether the existing document-level transformation functions will be updated to use an adapter (now or later or never)

eecavanna:
Still to do (note to self, @eecavanna):
- [ ] Use terminology that is agnostic to the underlying data store (e.g. ~`collection_name`~ -> `database_slot_name`)

I may postpone that task for a future PR (TBD).

ssarrafan:
Based on recent comment I'll move to the new sprint. Please remove from sprint if not working on this.

turbomam:
> Use terminology that is agnostic to the underlying data store (e.g. ~`collection_name`~ -> `database_slot_name`).
> * I'm a bit torn on this.

OK

nmdc_schema/migrators/README.md
```diff
@@ -4,22 +4,26 @@
...
+databases between schemas.
 
 In this document, I'll refer to those Python classes as "migrators."
 
 ## Contents
 
 This directory contains the following things:
 
+- `adapters/adapter_base.py` - definition of the `AdapterBase` class
+- `adapters/dictionary_adapter.py` - definition of the `DictionaryAdapter` class
```
eecavanna, line 17: Add a line about the Mongo adapter

eecavanna:
Looks like some migrators were merged in today that do not fit into the current framework. I'll update those soon.

turbomam (approved):

eecavanna:
Thanks for reviewing this! Next, I'm going to test the MongoDB migrators locally (hopefully this afternoon); after which point I'll be ready for this to be merged in. I'll post a follow-up comment after I test them locally.

brynnz22 (approved):
This all makes sense to me!

eecavanna:
Hi @turbomam, I noticed a potential dependency-related issue today. 

The `MongoAdapter` imports `pymongo`. 

`pymongo` is currently a **dev-dependency** of this project; not a production dependency. 

So, installing the `nmdc-schema` package from PyPI (once published with this branch) won't automatically install `pymongo`. 

I consider that inconsistency ("mismatch") to be a bug.

Our options include:
1. Remove the `MongoAdapter` from this repository. 
2. Make `pymongo` a production dependency instead of a dev-dependency. 

I think option 1 will facilitate maintenance of the `nmdc-schema` package, as a whole. Downside: more difficult to keep the adapters' capabilities in sync over time (requires inter-repo coordination).

I think option 2 will facilitate maintenance of the adapters over time (both are in same repo). Downside: introduces a dependency that may not benefit most schema package consumers.

What do you think?

turbomam:
I would vote for 2. because I'm comfortable doing all of my work out of the nmdc-schema repo. But maybe too comfortable. 

I mainly want an environment in which I can use the MongDB adapter without having to use dagster. Can I do that in a checkout of the nmdc-runtime repo, or iwth an import of the repo's code?

@mbthornton-lbl and I have been talking about potential other repos for data access. I think he would like to focus on methods that don't require any authentication, but I don't know how rigid he feels about that.

I would also like to include @sierra-moxon in this discussion.

eecavanna:
> Next, I'm going to test the MongoDB migrators locally (hopefully this afternoon); after which point I'll be ready for this to be merged in. I'll post a follow-up comment after I test them locally.

I finished testing the MongoDB migrators locally (i.e. I ran the `unittest`s I have since committed to this branch).

Outstanding tasks (these are blocking the merge):
- [ ] Designate `pymongo` as a production dependency (that's option 2 above)
- [ ] Confirm that I can import `MongoAdapter` from outside this project (i.e. I would be using `from nmdc_schema.migrators.adapters import MongoAdapter` in a Jupyter notebook in `nmdc-schema`)

eecavanna:
> @mbthornton-lbl and I have been talking about potential other repos for data access.

I think he's focusing on things that access data via HTTP APIs; as opposed to focusing on things that access the underlying database directly. The adapter falls into the latter category.

eecavanna:
> I mainly want an environment in which I can use the MongDB adapter without having to use dagster. Can I do that in a checkout of the nmdc-runtime repo, or iwth an import of the repo's code?

I'm interpreting these questions as (in my own words):
1. If the adapter were to live in the `nmdc-runtime` repo, how could I use it in my `nmdc-schema` workspace?
2. If the adapter were to live in the `nmdc-runtime` repo, would I have to run Dagster in order to use the adapter?

Here are my responses:
1. There's an `nmdc-runtime` package on [PyPI](https://pypi.org/project/nmdc-runtime/). I think you would be able to import that and get access to the adapter through that. However, `nmdc-runtime` already depends on `nmdc-schema`, so the result may be a circular dependency. I don't know how Python/pip/Poetry handles that.
2. No, Dagster wouldn't be involved.

---

On a side note, we are talking about using the adapters in more contexts than I have had in mind while implementing them. The only context I had in mind was the migration context. They might prove to be useful in other contexts also, but I'd prefer to explore that separately from this PR so that it doesn't delay the merging of this PR, since the implementation of migrators is currently blocked by the absence of the adapters (feature) in the `berkeley-schema-fy24` repo.

eecavanna:
All that's left for this branch before I'll be ready for it to be merged in is:

> Designate pymongo as a production dependency (that's option 2 above)

eecavanna:
I'm ready for this to be merged in (I won't use it until next week, though).

PR #1637, opened 2024-01-10
add `src/docs/element_names.md`

turbomam:


turbomam:
is there an optimal element name length (in tokens)?

maybe 2 as a rough guideline?

PR #1638, opened 2024-01-10
DO NOT MERGE: Simulate merging `berkeley-schema-fy24/main` into `nmdc-schema/main`

eecavanna:
Please do not merge.

Here's a PR for a merge in the opposite direction: https://github.com/microbiomedata/berkeley-schema-fy24/pull/65


turbomam:
@cmungall has encouraged us to merge this is ASAP

@eecavanna @brynnz22 @mslarae13 @aclum what are the prerequisites to doing that? I think at least one of you has a plan, and I wasn't able to articulate that to Chris very well.

eecavanna:
I'm OOO today. 

Not all of the migrators have been implemented yet; and at least one of them depends upon functionality that the migration framework doesn't provide yet (that is: the ability to mint IDs). So, if this merge were to happen and our systems were updated to that new version of `nmdc-schema`, we would not be ready to migrate our existing data accordingly (using our standard migration process).

In the meantime, we can continue to merge in the other direction on a regular basis, so that the eventual merge in this direction is simpler than if the differences were to accumulate.

mslarae13:
This PR says "DO NOT MERGE" but looks like it's been merged??
@turbomam @eecavanna

eecavanna:
Thanks, @mslarae13. It was done by accident (on the plus side, it confirmed for us how smoothly an eventual merge in that direction would be). We are working to effectively "undo" it as part of this newly-created issue: https://github.com/microbiomedata/nmdc-schema/issues/2077

eecavanna:
FYI: This accidental merge has effectively been "undone."

mslarae13:
So, should we click the "Revert" button on that merge so this isn't incorrectly labeled as "merged" or purple

eecavanna:
I'd rather nobody click that button. I don't think there is any way to make this PR say anything but "Merged." I think clicking the "Revert" button would just initialize a PR branch containing the _opposite_ of these changes (e.g. deletions instead of insertions and vice versa), so that merging _that_ PR branch would (effectively) "undo" the changes by making the opposite changes—this PR would still say "Merged," though.

I'll rename this PR, though, so it reflects the (new) reality in which the branch that was previously named `main` (which was the branch into which this PR was merged) has been renamed to something else.

Issue #1633, opened 2024-01-10
`get-study-related-records` aka `src/scripts/nmdc_database_tools.py` output path not working correctly


turbomam:
Fix:
- make output work as expected - default to project_root/local (excluded in .gitignore)
- remove dependencies on .env
- make api_base_url option explicit with a default to NAPA

By default, `get-study-related-records` tries to write an output and a logging file to `~/local/`. My intention was that it would go to the `local/` sub-directory in the `nmdc-schema` repo. In that case the files will be gitignored, i.e. not committed back to GitHub. Side note: it should be easy to expose them somewhere public.

Even better than that, I would prefer that `get-study-related-records` would take  `--output-file` and `--log-file` parameters. Then it would be easier/more deterministic to use the output in a subsequent validation target. The current behavior builds filenames based on the `--study_id`, which is admittedly nice in the case where many files from this process might be accumulated in the same directory.

turbomam:
```python
@click.option("--output_dir", default="../../local/", help="Output directory.")
```

turbomam:
Now running like

mbthornton-lbl:
related #1634

Issue #1635, opened 2024-01-10
All click options should use hyphens not underscores


turbomam:
Click knows that parameters expressed with hyphens will create variables with underscores

So for `src/scripts/nmdc_database_tools.py` we would have a section like this:

```Python
@click.group()
@click.option("--api-base", default="NAPA", help="API base to use.")
@click.option("--env-file", default="../../local/.env", help="dotenv file for username, password, etc.")
@click.option("--output-dir", default="../../local/", help="Output directory.")
@click.pass_context
def cli(ctx, api_base, env_file, output_dir):

turbomam:
Although we have other issues with suggestions for eliminating `--env-file`, using a more literal `--api-base` and specifying `--output-file` instead of `--output-dir` !

turbomam:
see also
- #1633 
- #1634

mbthornton-lbl:
covered by #1641

Issue #1634, opened 2024-01-10
`get-study-related-records` should not require an `.env` file at all even for the API base


turbomam:
just get the API base from the Click interface and provide a reasonable default

This will also eliminate the abstraction of API endpoints with values like "NAPA"

mbthornton-lbl:
fixed in #1641

PR #1641, opened 2024-01-11
1633 get study related records aka srcscripts nmdc_database_tools.py outputd path not working correctly

mbthornton-lbl:
This PR fixes:
#1633 output-path argument not working correctly
#1634 should not require a .env file
#1635 click options should use hyphens not underscores

Additionally, this PR provides
- ability to provide study-id AND/OR output-dir.  If output-dir is provided as `{local_dir}/nmdc-sty-1-abcdef` with no study-id provided, the tool will infer a study ID `nmdc:sty-1-abcdef` for the last part of the output-dir arg.
- updated `project.Makefile` to run using a local output-dir target

### DIFF
--- a/project.Makefile
+++ b/project.Makefile
@@ -582,10 +582,8 @@ migrator:
 	$(RUN) create-migrator
 
 .PHONY: extract-study
-extract-study:
+extract-study: local/nmdc-sty-11-aygzgv51
 	$(RUN) get-study-related-records \
-		--api_base NAPA \
-		--env_file local/.env \
+		--api-base-url https://api-napa.microbiomedata.org \
 		extract-study \
-		--study_id nmdc:sty-11-aygzgv51 \
-		--yaml_out
+		--output-dir $<
--- a/project.Makefile
+++ b/project.Makefile
@@ -581,9 +581,10 @@ migration-doctests:
 migrator:
 	$(RUN) create-migrator
 
-.PHONY: extract-study
-extract-study: local/nmdc-sty-11-aygzgv51
+local/nmdc-sty-11-aygzgv51.yaml:
 	$(RUN) get-study-related-records \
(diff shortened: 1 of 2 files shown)


project.Makefile
```diff
@@ -582,10 +582,8 @@ migrator:
 	$(RUN) create-migrator
 
 .PHONY: extract-study
-extract-study:
+extract-study: local/nmdc-sty-11-aygzgv51
```
turbomam, line 585: please rewrite the Python script so that both the study id and an output path (including filename) are required. Once we've done that, we can write linked makefile targets like this:

```make
local/nmdc-sty-11-aygzgv51.yaml:
	$(RUN) get-study-related-records \
		--api-base-url https://api-napa.microbiomedata.org \
		extract-study \
		--study-id nmdc:sty-11-aygzgv51 \
		--quick-test \
		--output-file $@

local/nmdc-sty-11-aygzgv51-validation.log: nmdc_schema/nmdc_materialized_patterns.yaml local/nmdc-sty-11-aygzgv51.yaml
	$(RUN) linkml-validate --schema $^ > $@
```

turbomam: we can add Makefile inference of the study id from the target name/output file path later

https://www.gnu.org/software/make/manual/html_node/Text-Functions.html

you can get a rough idea of that here:

- https://github.com/microbiomedata/nmdc-schema/blob/20010c597e1b5f00dcc194bf22cd7e4eb1710b29/project.Makefile#L542-L550

turbomam: We can consult with @cmungall if necessary. His Make skills are greater than mine, but ironically he often suggests something simpler.

turbomam (changes requested):

turbomam (approved):

PR #1670, opened 2024-01-16
support for local Fuseki

turbomam:


eecavanna:
This afternoon, @turbomam and I walked through the process of configuring `docker-compose.yml` to run an off-the-shelf Fuseski image as a separate container.

turbomam:
> run [this off-the-shelf Fuseki image](https://hub.docker.com/r/secoresearch/fuseki/) as a separate container. I think that is the direction he will be taking this pursuit

Thanks for the tutorial @eecavanna and for writing this up. I do like the idea of using an off-the-shelf solution, but I'm still not clear on where the database directory is on the filesystem after spinning up the Fuseki server.

I also feel like secoresearch/fuseki is a little elaborate for our needs. It doesn't take too long to spin up and I haven't really checked the RAM, CPU or disk usage (but I assume they're modest). It just looks more complicated than I'm used to.

I guess I need to open another issue for the difficulties running the Dockerfile on Ubuntu Linux, even in the absence of the Fuseki server. It builds but fails to launch the mkdocks server.

turbomam:
I think I might prefer to use this containerization: https://jena.apache.org/documentation/fuseki2/fuseki-docker.html

turbomam:
maybe this one?

https://hub.docker.com/r/dskow/jena-fuseki

turbomam:
This standalone command launches fuseki in a configuration that is almost good enough. I would just want to add externalk storage of the TDB2 backed directory (and then pre-loading data into the TDB2) 

```shell
docker run --name fuseki -p 3031:3030 dskow/jena-fuseki
```

the password is `pw`, not anything that appears in a log or a shiro config file

turbomam:
> Fuseki's data is stored in the Docker volume /var/fuseki_home within the container. Note that unless you use docker restart or one of the mechanisms below, data is lost between each run of the jena-fuseki image.




trying

_fuseki_home or home_fuseki ?!_

```
docker run --name fuseki-data -v /var/fuseki_home busybox
```
----

```
docker run --rm \
   --volumes-from fuseki-data ubuntu bash \
   -c "find /var/fuseki_home | xargs -i chown 1000:1000 {}"
```

----

optionally `-d`

```
docker run  --name fuseki \
  -p 3031:3030 \
   --volumes-from fuseki-data dskow/jena-fuseki
```

despita all of that, the data appears to be in the fuseki container's /tm/databases directory

turbomam:
## https://hub.docker.com/r/stain/jena-fuseki

```
docker run --name fuseki -p 3030:3030 stain/jena-fuseki
```

works and the reported password is honored

turbomam:
closed in favor of https://github.com/microbiomedata/nmdc-schema/pull/1686

PR #1674, opened 2024-01-17
1566 adding to InstrumentModelEnum and InstrumentVendorEnum for sequencing applications

aclum:
This PR expands on the two enumerations to support sequencing based vendors and models.


aclum:
wrong branch

Issue #1669, opened 2023-12-27
'sample linkage' may not always be a GUID


mslarae13:
Current description of `slot:sample link`

"A unique identifier to assign parent-child, subsample, or sibling samples. This is relevant when a sample or other material was used to generate the new sample."

Because we no longer required a `source material identifier` or `sample GUID` (GUID is assigned by NMDC) the user may not have a GUID to provide sample links. We should update the description to say "The sample name of samples to be linked to this sample. This should match a sample name provided to NMDC & should identify the sample relationship that is being provided. This is relevant when a sample or other material was used to generate the new sample."

Should update examples. 
core: soil core 1
parent: sample A
pooling: sample 1, sample 2, sample 3

mslarae13:
Blocked:
- How do we structure sample_link?
- What is the use of sample_link?

PR #1684, opened 2024-01-18
values in the `sample_link` slot may not always follow a formal syntax, like CURIe or GUID

bmeluch:
From 
- https://github.com/microbiomedata/nmdc-schema/issues/1669

I updated the slot definition as described in the original issue, then updated the valid example files that already contained `sample_link` and created an invalid example file.

### DIFF
--- a/src/data/invalid/Biosample-invalid-linkage.yaml
+++ b/src/data/invalid/Biosample-invalid-linkage.yaml
@@ -17,3 +17,4 @@ sample_link:
   - 'this sample is pooled from three other samples'
   - 'but I am not even listing which samples'
   - 'because of how bad and invalid this example is'
+  - 23456
--- a/src/data/invalid/Biosample-invalid-linkage.yaml
+++ b/src/data/invalid/Biosample-invalid-linkage.yaml
@@ -1,0 +1,19 @@
+id: nmdc:bsm-99-dtTMNb
+part_of:
+  - nmdc:sty-00-abc123
+env_broad_scale:
+  has_raw_value: ENVO:00002030
+  term:
+    id: ENVO:00002030
+env_local_scale:
+  has_raw_value: ENVO:00002169
+  term:
+    id: ENVO:00002169
+env_medium:
+  has_raw_value: ENVO:00005792
+  term:
+    id: ENVO:00005792
+sample_link: 
+  - 'this sample is pooled from three other samples'
+  - 'but I am not even listing which samples'
+  - 'because of how bad and invalid this example is'
--- a/src/data/valid/Biosample-exhasutive.yaml
+++ b/src/data/valid/Biosample-exhasutive.yaml
@@ -455,8 +455,7 @@ exp_duct:
 exp_pipe:
   has_raw_value: xxx
 
-experimental_factor_other: unconstrained text, but presumably expects 'term label
-  [term id]'
+experimental_factor_other: unconstrained text, but presumably expects 'term label [term id]'
 ext_door:
   has_raw_value: xxx
 ext_wall_orient: north
--- a/src/data/valid/Biosample-exhasutive.yaml
+++ b/src/data/valid/Biosample-exhasutive.yaml
@@ -876,8 +876,7 @@ samp_well_name:
   has_raw_value: xxx
 
 sample_link:
-  - IGSN:DSJ0284
-  - any:curie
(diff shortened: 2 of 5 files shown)


turbomam (commented):
Thanks @bmeluch 

I expect that I can merge this in very soon, but I would like to touch base with you and any other stakeholders first.

Your examples illustrate three different kinds of relationships (as the prefix portion of the values), but is that an exhaustive list? Is there some documentation or code snippet that shows how the different kinds of relationships will be expressed in data? Where will we find the part of the sample linkage value that comes after the colon? Like, **which** slot of some other Biosample I guess?

bmeluch:
Hi @turbomam, 
Good question - I would have to double check with @mslarae13 about the complete list of relationships (before colon). I believe the part after the colon has to match `samp_name` (s) of the linked biosamples.

src/data/valid/Biosample-exhasutive.yaml
```diff
@@ -876,8 +875,7 @@ samp_well_name:
   has_raw_value: xxx
 
 sample_link:
-  - IGSN:DSJ0284
-  - any:curie
+  - 'core: soil_core_O'
```
mslarae13, line 878: I would expect core to be source_mat_id. so i think this one should be removed. Maybe replace core with plant:

src/data/valid/Biosample-exhaustive-issue-796-bye-yq-for-7-4-10.yaml
```diff
@@ -380,9 +379,9 @@ technical_reps: "2"
 analysis_type:
   - metabolomics
   - metagenomics
+
 sample_link:
-  - IGSN:DSJ0284
-  - any:curie
+  - 'parent: example_samp_45'
```
mslarae13, line 384: parent is source_mat_id ... 1:1 parent to child. 
sample link is other sample types or horizontal relationships.. maybe make this one sibling:

src/schema/portal/sample_id.yaml
```diff
@@ -46,16 +46,19 @@ slots:
...
     comments:
       - 'This field allows multiple entries separated by ; (Examples: Soil collected
       from the field will link with the soil used in an incubation. The soil a plant
       was grown in links to the plant sample. An original culture sample was transferred
       to a new vial and generated a new sample)'
     examples:
-      - value: IGSN:DSJ0284
+      - value: 'core: soil core 1'
+      - value: 'parent: sample A'
+      - value: 'pooling: sample 1, sample 2, sample 3'
```
mslarae13, line 61: pooling is good!

mslarae13 (changes requested):

mslarae13:
@bmeluch ignore my comments until we chat. I need to review the NMDC schema and submission portal. source_mat_id needs some work. We can discuss this this week but I think we'll make a squad to do this.

turbomam:
> We can discuss this this week but I think we'll make a squad to do this.

Could I be part of this conversation?

src/data/invalid/Biosample-invalid-linkage.yaml
```diff
@@ -0,0 +1,20 @@
...
+    id: ENVO:00002030
+env_local_scale:
+  has_raw_value: ENVO:00002169
+  term:
+    id: ENVO:00002169
+env_medium:
+  has_raw_value: ENVO:00005792
+  term:
+    id: ENVO:00005792
+sample_link:
```
turbomam, line 16: is this a list of three free stating strings and an integer?

turbomam (commented):
I asked about the vocabulary of relationship types before, but now I think we need to discuss the overall grammar, and what agent (person or software) is responsible for interpreting it.

It wouldn't be too hard to do that with structured objects in the nmdc-schema, but it  will be harder in the flattened submission-schema. Are you planning on allowing different relationships to be semi-colon delimited in the DH cells? Will there be any entry-time validation?

turbomam:
We defined `sample_link` in `src/schema/portal/sample_id.yaml`, so we can do whatever we want with it, right? Let's start making a distinction between a more structured modeling in nmdc-schema and move this flattened approach to the submission-schema, where it will be totally relevant and appropriate.

bmeluch:
This is now way out of date, and we need more discussion on `sample_link` - closing this but will leave the issue open and start a new PR when the discussion comes back around.

PR #1685, opened 2024-01-19
Berkeley main replica

turbomam:
merged a local copy of the `nmdc-schema` into a local duplicate of `berkeley-schema-fy24`

now comparing against `berkeley-schema-fy24`'s `main`

will need to review the example data files that have been moved into the `problem` sub-directory

@eecavanna who else should we invite as reviewers?


turbomam:
closing in favor of
- https://github.com/microbiomedata/berkeley-schema-fy24/pull/73

Issue #1687, opened 2024-01-19
make the docker solution `nmdc-schema`'s preferred dev environment


turbomam:
- clear out environment variables like JENA_PATH
- ensure that all containers run in at least MacOS and Linux (and preferably Windows too)

turbomam:
The Docker environment still requires some tweaking, but we can create specific issues for that

PR #1689, opened 2024-01-20
Fix issue where `app` container cannot run Java (in `amd64` host environment)

eecavanna:
### Summary of changes

- Configured the `app` container to have an arm64 build of the JDK in arm64 environments, and an amd64 build in all other environments

Related: https://github.com/microbiomedata/nmdc-schema/issues/1687

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,12 +9,13 @@ RUN wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
     chmod +x /usr/bin/yq
 
 # Download and install a Java Development Kit >= 11.
+# Reference: https://jdk.java.net/21/
 # Note: This is a dependency of Apache Jena.
-RUN wget -P /downloads/tmp "https://download.java.net/java/GA/jdk21.0.1/415e3f918a1f4062a0074a2794853d0d/12/GPL/openjdk-21.0.1_linux-aarch64_bin.tar.gz"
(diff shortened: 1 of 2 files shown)


eecavanna:
I think I made a mistake somewhere. In a container based upon this updated `Dockerfile`, I am actually **not** able to run Java inside the `app` container.

Here's what happens when I try:

```shell
root@af8333ee5a4b:/nmdc-schema# uname -a
Linux af8333ee5a4b 6.5.11-linuxkit #1 SMP PREEMPT Wed Dec  6 17:08:31 UTC 2023 aarch64 GNU/Linux
root@af8333ee5a4b:/nmdc-schema# /downloads/openjdk/jdk-21.0.2/bin/java --version
qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
```

Environment: MacBook Pro M1

So, I think this change would break the container on ARM64 machines after all; in which case, we may need for there to be separate containers (or the container can include both versions of Java and some way of choosing the right one, depending upon the architecture).

I'll revisit this next week, @turbomam.

eecavanna:
In the latest commit, I updated the `Dockerfile` so that the system downloads an arm64 build of the JDK if the system architecture is arm64; otherwise, it downloads an amd64 build of the JDK. I also specified the name I want the downloaded file to have locally, so that it has the same name in both scenarios (allowing the subsequent commands to be simpler—and to not change since the previous commit).

eecavanna:
I'm ready for this to be reviewed and merged (e.g. next week). 

I'm particularly interested in whether the `Dockerfile` in the latest commit works in an amd64 environment (which I'm not prepared to test, myself).

turbomam (approved):

PR #1696, opened 2024-01-22
move lists of yq commands from project.Makefile target into a separate files

turbomam:


turbomam:
sorry, you can ignore commits to `project/`

PR #1732, opened 2024-01-25
Try to ensure badge appears up-to-date

eecavanna:
### Summary of changes

- I added a query string to the image URL in an attempt to cause a cache miss. I think some web browsers will still cache the image, since the query string doesn't change; but I want to confirm over time.


turbomam (approved):

turbomam:
@pkalita-lbl I would like to merge this. It doesn't effect the schema.

pkalita-lbl:
Sure, go for it

PR #1733, opened 2024-01-25
wipe all traces of "has unit" etc

turbomam:
thanks for you help brainstorming and debugging @sujaypatil96 

@pkalita-lbl I am tempted to merge this and make a 10.0.1 release, but you had written a [slack message](https://nmdc-group.slack.com/archives/CFVF3G3H7/p1706128772003919) suggesting that we refrain from doing that kind of thing before the next scheduled nmdc ecosystem release.

This PR removes `has unit` and some other slots whose names begin with 'has' and a whitespace. They were never used anywhere.


turbomam:
In addition to removing unused, old-style elements from the schema, this PR also improves the expereince of searching for "has unit" or "has_unt" in the online docs.

pkalita-lbl (approved):
This looks fine but let's just pause on merging it until after the 2024.1 release is on production. In the unlikely even that a critical change to the schema is needed for the release I'd like to not introduce any extra uncertainty.

sujaypatil96 (approved):

PR #1738, opened 2024-01-27
referential integrity checking of napa squads mongodb via sparql in fuseki

turbomam:
@Michal-Babins you may be interested in this too. We should all meet and discuss at some point

I will try to attach the referential integrity report, within the scope of the data that was retrieved (see the local/some_napa_collections.yaml` target in the `project.Makefile`)


turbomam:
[objects-that-are-never-subjects.tsv.txt](https://github.com/microbiomedata/nmdc-schema/files/14070467/objects-that-are-never-subjects.tsv.txt)

aclum (approved):

aclum:
I think a lot of the data object issues are b/c we didn't load the data_object_set when we were testing, I've fixed the OmicsProcessing part_of issue.

PR #1768, opened 2024-02-12
1767 map jgi template to nmdc slots add alias

mslarae13:
adding exact mappings to EMSL and JGI slots.


mslarae13:
exact_mappings aren't URI/CURIE... changed to aliases

turbomam (approved):

PR #1770, opened 2024-02-13
1747 improve logging for database tools `extract-study` warn for missing data objects

mbthornton-lbl:
This PR provides minor updates to the `extract-study` tool:

- make study_id an argument rather than option
- remove filename normalization swapping ':' for '-' in study_id.  Colons appear to work fine in filenames
- update logging ERROR for orphan data objects or study-id not found

### DIFF
--- a/src/scripts/nmdc_database_tools.py
+++ b/src/scripts/nmdc_database_tools.py
@@ -21,17 +21,6 @@ SCRIPTS_DIR = Path(__file__).parent
 PROJECT_ROOT = SCRIPTS_DIR.parent.parent
 LOCAL_DIR = PROJECT_ROOT / "local"
 
-# Studies with non-compliant IDs that have been re-IDed
-# Name:, study ID, legacy study ID
-STUDIES = {
-    "Stegen": ("nmdc:sty-11-aygzgv51", "gold:Gs0114663"),
-    "SPRUCE": ("nmdc:sty-11-33fbta56", "gold:Gs0110138"),
-    "EMP": (None, "gold:Gs0154244"),
-    "Luquillo": (None, "gold:Gs0128850"),
-    "CrestedButte": ("nmdc:sty-11-dcqce727", "gold:Gs0135149"),
-}
-# Update study IDs after re-IDing
-DEFAULT_STUDY_ID = STUDIES["Stegen"][0]
 
 # TODO: Move NmdcApi class to the nacent nmdc-common/client package
 # TODO: Add unit tests for NmdcApi class
--- a/src/scripts/nmdc_database_tools.py
+++ b/src/scripts/nmdc_database_tools.py
@@ -132,18 +132,25 @@ def cli(ctx, api_base_url):
 
 
 @cli.command()
-@click.argument("study_id", required=True,)
-@click.option("--output-file", help="Path to output file, relative to project "
-                                   "root.")
-@click.option("--quick-test", is_flag=True, default=False, help=("Quick test "
-                                                                 "mode - "
-                                                                 "biosamples "
-                                                                 "and omics "
-                                                                 "only "))
-@click.option("--search-orphaned-data-objects", is_flag=True, default=False,
-              help=("Search for orphaned data objects by description."))
-@click.option("--search-legacy-identifiers", is_flag=True, default=False,
-              help=("Search for legacy IDs for Study and OmicsProcessing."))
+@click.option("--study-id", required=True, )
+@click.option(
+    "--output-file",
+    help="Path to output file, relative to project root."
+)
+@click.option(
+    "--quick-test",
+    is_flag=True,
+    default=False,
+    help=("Quick test mode, Biosamples and OmicsProcessing records only")
+)
+@click.option(
+    "--search-orphaned-data-objects", is_flag=True, default=False,
+    help=("Search for orphaned data objects by description.")
+)
+@click.option(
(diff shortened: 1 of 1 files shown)


src/scripts/nmdc_database_tools.py
```diff
@@ -141,7 +143,7 @@ def cli(ctx, api_base_url):
 
 
 @cli.command()
-@click.option("--study-id", required=True,  help="Study ID to extract.")
+@click.argument("study_id", required=True,)
```
turbomam, line 146: please use hyphenated options

src/scripts/nmdc_database_tools.py
```diff
@@ -229,7 +231,9 @@ def extract_study(ctx, study_id, output_file, quick_test, search_orphaned_data_o
     search_study_ids = [study_id]
     if search_legacy_identifiers:
         # add legacy study IDs to search
+        logger.info("Searching using current and legacy identifiers")
         search_study_ids.extend(study.get("gold_study_identifiers", []))
+        logger.info(f"Using study IDs: {search_study_ids}")
```
turbomam, line 236: why are we making a `search_study_ids` list with one item?

turbomam (commented):
your code for building directories is different from the `os` join methods I'm used to . What are the advantages and disadvantages?

```python
SCRIPTS_DIR = Path(__file__).parent
PROJECT_ROOT = SCRIPTS_DIR.parent.parent
LOCAL_DIR = PROJECT_ROOT / "local"
```

turbomam:
It makes me really nervous that we are codifying nicknames for studies in this script. It's not your fault that we have nicknames, but I am going to push for them to be added to the Study instances in MongoDB. So hopefully we won;t need this after that. You would just modify your code to look by `id` or by nickname.

```python
# Studies with non-compliant IDs that have been re-IDed
# Name:, study ID, legacy study ID
STUDIES = {
    "Stegen": ("nmdc:sty-11-aygzgv51", "gold:Gs0114663"),
    "SPRUCE": ("nmdc:sty-11-33fbta56", "gold:Gs0110138"),
    "EMP": (None, "gold:Gs0154244"),
    "Luquillo": (None, "gold:Gs0128850"),
    "CrestedButte": ("nmdc:sty-11-dcqce727", "gold:Gs0135149"),
}
# Update study IDs after re-IDing
DEFAULT_STUDY_ID = STUDIES["Stegen"][0]
```

src/scripts/nmdc_database_tools.py
turbomam: actually, why are we stitching from a `click.option` to a `click.argument`?

turbomam:
I would prefer this formatting 

```python
@click.option("--quick-test", 
              is_flag=True, 
              default=False, 
              help=("Quick test mode - biosamples and omics only "))
```

over

```python
@click.option("--quick-test", is_flag=True, default=False, help=("Quick test "
                                                                 "mode - "
                                                                 "biosamples "
                                                                 "and omics "
                                                                 "only "))
```

mbthornton-lbl:
> your code for building directories is different from the `os` join methods I'm used to . What are the advantages and disadvantages?
> 
> ```python
> SCRIPTS_DIR = Path(__file__).parent
> PROJECT_ROOT = SCRIPTS_DIR.parent.parent
> LOCAL_DIR = PROJECT_ROOT / "local"
> ```

This is using `pathlib` which I find to be simpler and more succinct than using `os`

src/scripts/nmdc_database_tools.py
mbthornton-lbl: If the `search-legacy-identifiers` option is selected, there could be >1 `study_id` (NMDC study ID and gold dtudy identifier(s))that we want to use as a search term for finding biosamples / omics processing.

turbomam: so does `--study-id` need to be configure to take multiple values and return a list? I think that would be good future proofing.

mbthornton-lbl: I might lean the other way, actually. It's there right now only to support the ability to search `part_of` and `was_informed_by` using old (non-conformant) identifiers.  In normal cases we should not have any of these, so one could argue that once we get past the Re-ID-ing work (soon, I hope) we won't need either functionality anymore.  It's here as kind of a `look-for-relationships-that-should-not-exist` function for doing QC on our re-ID-ed studies

mbthornton-lbl: Similarly for `--search-orphaned-data-objects`.  Normally, these things should not exist, so once we are done with this data cleaning exercise we can remove these options and their logic and leave `extract_study` cleaned up and simplified.

turbomam:
can you say more about "Colons appear to work fine in filenames". What systems have you checked that on?

I have the expectation that this will work on recent MacOSes and Linuxes. It would be great it it ran on Windows too.

I guess I'm saying "In the absence of cross-platform testing, I want to be more defensive about file names"

mbthornton-lbl:
The main additional functionality was to make the log file more useful by elevating missing data objects to ERROR, and some other less adverse case to WARNING

My rationale for using colons in the filenames:
- It is consistent with the way data directories are name in our data store: `/global/cfs/cdirs/m3408/results`
- Same as corresponding object in the DB
- Colons in directory names *usually* don't do any harm ;-)
- removes the need for a normalization / de-normalization kludge

turbomam (approved):

Issue #1795, opened 2024-02-22
Create declarative mapping file for NCBI BioSample ingest


sujaypatil96:
A big data source that we will be ingesting multiple studies from in the coming weeks/months is NCBI, and for this we need a generic NCBI study/BioProject to NMDC ingest pipeline/importer/translator (similar to the GOLD study ingest pipeline).

The source will be a SQLite version of the entire content of NCBI BioSample. We will be using this to create NMDC Biosample records, and similar to other ingest pipelines we will be using the NMDC schema python dataclasses to create instances.

One of the first tasks that we will need to do as part of this process is to create a declarative mapping between the columns/fields in the NCBI BioSample SQLite source table/view, and NMDC schema slots.

After discussion with @turbomam, we decided to implement the mapping file as a TSV file with the following columns:

| NMDC schema class | NMDC schema slot (use dot syntax for nested ranges) | NCBI BioSample attribute name | static_value (optional) |
| ------------- | ------------- | ------------- | ------------- |

We will be adding a TSV file to this repository and updating it regularly with more mappings.

sujaypatil96:
To start working on this issue, contrary to ingest pipelines for other sources like NEON, GOLD, etc. we agreed to start from the NMDC schema class slots (for classes Biosample and OmicsProcessing) and map them to the right NCBI BioSample attribute names from the postgres database dump.

turbomam:
We have some good discussions about this in Slack threads and other issues

A recent theme has been the use of a standard mapping format, like SSSOM. How would you feel about that, @sujaypatil96 

I think our mapping **mechanism** is going to move towards an ad-hoc LLM query although I am open to using any off the self tools that are popular in the BBOP LinkML Monarch and OBO communities

sujaypatil96:
We had a discussion with @caufieldjh on 3/15 on the possibility of using existing LLM tools (like ontogpt) to drive a semi automated mapping.

But one of the main realizations that we had was to stop using a [dump](https://github.com/microbiomedata/nmdc-schema/blob/main/assets/ncbi_mappings/ncbi_pg_db_fields.txt) of all the columns/fields from the NCBI BioSample Postgres database, but instead use the XML dump of NCBI BioSample Attributes accessible here: https://www.ncbi.nlm.nih.gov/biosample/docs/attributes/?format=xml

mslarae13:
Should add this to the 'push to NCBI squad' proposal

sujaypatil96:
[...]

PR #1797, opened 2024-02-27
NCBI BioSample attributes/NMDC schema slots mapping file - exact term matching

sujaypatil96:
Add TSV file described in issue #1795 to start fleshing out a mapping file between NCBI BioSample attributes and NMDC schema slots.

CC: @turbomam


sujaypatil96:
I have not pushed up any code/updates to the mapping file yet, but I do have changes/updates on my local machine that I wanted to discuss with you first before pushing it up. We can talk about that tomorrow, 3/11.

turbomam:
Looking forward to reviewing your work!

turbomam:
The exact mapping to NCBI term names and synonyms are great, but the fuzzy matches aren't actionable. As discussed, I recommend using a LLM.

sujaypatil96:
@turbomam I have pushed up the code for exact term matching between NMDC schema slots and NCBI postgres db view field/column names.

sujaypatil96:
I have re-titled this PR to indicate that it has mappings from exact term matching. This PR is good to be merged in after your review. We can make another PR for the LLM based matching strategy.

turbomam (approved):

PR #1799, opened 2024-02-27
Update EMSL sample type enum

mslarae13:
added to emsl SampleTypeEnum


pkalita-lbl:
If I'm understanding what you're working on correctly, nothing needs to happen with `submission-schema`. Once a version of `nmdc-schema` with your changes in it is released, it can be brought into `nmdc-server` and then you should see the new options available in the interface.

turbomam (approved):

PR #1802, opened 2024-02-28
Invoke callback functions after manipulating collections

eecavanna:
This branch introduces callback support into the migration framework. It also introduces additional unit tests.

### Summary of changes

- Consumers can now specify callback functions when instantiating an adapter. The adapter will invoke those callback functions—if defined—when corresponding events happen. That will allow the consumers to "react" to those events. Supported events are currently: creating a collection, renaming a collection, and deleting a collection.
- The adapters now also check whether a collection exists, before operating upon it.
- I added unit tests targeting the `DictionaryAdapter` (previously, there were none)
- I added unit tests targeting the callback functionality, specifically, of both adapters.
- I updated the `migration_recursion.py` script to use callback functions (as an example).
  > I used `lambda`s to keep the example concise. I could have defined functions and passed references to them instead; like this:
  > ```py
  > def my_func(name: str):
  >     pass
  >
  > dictionary_adapter = DictionaryAdapter(
  >     on_collection_created=my_func,
  >     # ...
  > )
  > ```

### DIFF
--- a/nmdc_schema/migrators/adapters/dictionary_adapter.py
+++ b/nmdc_schema/migrators/adapters/dictionary_adapter.py
@@ -9,15 +9,18 @@ class DictionaryAdapter(AdapterBase):
     a Python dictionary.
     """
 
-    def __init__(self, database: dict) -> None:
+    def __init__(self, database: dict, **kwargs) -> None:
         r"""
-        Initializes the reference to the database this adapter instance will be used to manipulate.
+        Invokes the initialization method of the parent class, passing to it any specified keyword arguments.
+        Also initializes the reference to the database this adapter instance will be used to manipulate.
         """
+        super().__init__(**kwargs)
         self._db = database
 
     def create_collection(self, collection_name: str) -> None:
         r"""
         Creates an empty collection having the specified name, if no collection by that name exists.
+        Also invokes `self.on_collection_created`, if defined, passing to it the name of the collection.
 
         >>> database = {
         ...   "thing_set": [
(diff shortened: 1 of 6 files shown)


eecavanna:
FYI @turbomam, here's what I had in mind regarding the callback functions we discussed today. I haven't tested this code yet (so am leaving the PR in "draft" state).

CC: @brynnz22

nmdc_schema/migrators/adapters/dictionary_adapter.py
```diff
@@ -218,7 +242,10 @@ def delete_documents_having_value_in_field(self, collection_name: str, field_nam
         # Filter out the documents having the specified value in the specified field.
         # Reference: https://docs.python.org/3/library/functions.html#filter
         documents_initial = self._db[collection_name]
-        document_generator = filter(lambda d: field_name not in d or d.get(field_name) != value, documents_initial)
+        document_generator = filter(
+            lambda d: field_name not in d or d.get(field_name) != value,
+            documents_initial,
+        )
```
eecavanna, line 248: Note: This differs in format only — the code is the same as before, just reformatted.

eecavanna:
Hi @turbomam, I am ready for this to be reviewed and—if approved—merged into `main`.

eecavanna:
This PR branch seems to introduce a testing process that differs from the rest of the repo. Consider consolidating them. @eecavanna

eecavanna:
@turbomam will re-run the usual "shebang" of make targets on this branch before merging it in.

turbomam (approved):

PR #1808, opened 2024-02-28
Move docs outside of repo file tree before running baler link checker

eecavanna:
This is my first (hopefully, only) attempt at fixing the [issue](https://github.com/microbiomedata/nmdc-schema/actions/runs/8086888068/job/22097642276#step:7:94) where the baler link checker (GitHub Actions workflow job step) deletes the `docs` folder before trying to check it for broken links.

### DIFF
--- a/.github/workflows/check-links.yaml
+++ b/.github/workflows/check-links.yaml
@@ -24,8 +24,8 @@ env:
   # Files to check. (Put patterns on separate lines, no leading dash.)
   # Reference: https://github.com/marketplace/actions/baler-bad-link-reporter#files
   files: |
-    docs/*.md
-    docs/**/*.md
+    /docs/*.md
+    /docs/**/*.md
 
   # Label(s) assigned to issues created by this workflow. (Use commas to separate labels.)
   # Reference: https://github.com/marketplace/actions/baler-bad-link-reporter#labels
@@ -59,6 +59,8 @@ jobs:
       issues: write
     steps:
       - name: Check out commit
+        # TODO: There's a newer version (v4) of this `checkout` action available.
+        #       Reference: https://github.com/actions/checkout
         uses: actions/checkout@v2
       - name: Set up Python
         uses: actions/setup-python@v3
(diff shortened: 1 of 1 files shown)


eecavanna:
I am ready for this to be merged in.

sujaypatil96:
Is there a report being generated about bad links by the baler link checker, and if so, then where can we find it?

eecavanna:
> Is there a report being generated about bad links by the baler link checker, and if so, then where can we find it?

No report yet. Once this is merged into `main` and then someone triggers a run of the workflow, the baler link checker will create an Issue reporting the broken links it finds.

eecavanna:
> I'm not following the need to move the docs folder outside of the repo file tree

The reason I'm moving it out of the file tree is that, when it's in the repo file tree, the baler action deletes it. Moving it out the repo file tree before the baler action runs prevents that deletion from happening.

turbomam:
OK, that representation of the report is pretty good. I don't feel like the `docs/` directory needs to be checked, although we do need to check through the `src/schema/*.yaml` files from which `docs/` is derived. 

I'm comfortable merging this in. OK with you @sujaypatil96 ?

eecavanna:
> I am a little confused about the deletion of the docs/ directory. Would that deletion get checked back into the repo? I don' think we're doing any committing of changes in any of the other GH Actions.

No, nothing is being committed or checked in in this GHA workflow.

eecavanna:
> I don't feel like the `docs/` directory needs to be checked, although we do need to check through the `src/schema/*.yaml` files from which docs/ is derived.

This baler tool may not be a good fit. I think it only checks Markdown files.

turbomam:
> No, nothing is being committed or checked in in this GHA workflow.

Thanks for the quick reply. If nothing is being committed, then why does it matter that the directory is deleted in the action?

eecavanna:
I'll message you on Slack.

eecavanna:
Before the baler action performs the link checking, it restores the repo file tree to the state it is in in the repository, which means it deletes the built docs from the `docs` folder. This PR moves the build docs outside of the file tree so they don't get deleted at that stage. This PR also tells baler it can find the docs in the place where they were moved to. This is all to allow the link checking to happen. None of these things will impact anything outside of this GHA workflow, except that a GitHub Issue will be created it baler finds any broken links. Does that make sense?

eecavanna:
Maybe me using the term "repo file tree" is creating confusion. When I say "repo file tree", I mean the file tree (i.e. directories and files) that gets created with Git is used to check out a commit.

For example, when I run:
```
git clone https://example.com/my-repo.git
```
Git will create a `my-repo` folder. That folder and everything in it is what I'm calling the "repo file tree".

Making changes in the "repo file tree" (not including the `.git` folder) doesn't affect the repo unless someone performs a `git commit` at some point. There is no `git commit` happening in this GHA workflow.

eecavanna:
In case you'd feel more comfortable with this GHA workflow doing a `cp` (copy) instead of a `mv` (move) of the built docs, that would also work. The baler action will delete them from the repo file tree if they exist.

turbomam (approved):

PR #1819, opened 2024-03-05
Use Lychee directly instead of going through Baler

eecavanna:
This workflow has been tested in https://github.com/eecavanna/nmdc-schema-temp/commit/e3fc8235bdb7cfcad1fce98f85fb66f715b8aeb6 (it works). It basically does the same thing as the Baler-based would have done. Baler uses this (Lychee) under the hood.

This workflow takes about 90 seconds to run.

image

Here's an example of a GitHub Issue created by it: https://github.com/eecavanna/nmdc-schema-temp/issues/2. We can go through that report and come up with some Lychee configuration that ignores certain types of errors. I'd prefer to have this be merged in, run it in this repo (its trigger is still a manual one), then create an issue about customizing Lychee's configuration to be tolerant of whatever you consider to be a false positive.

### DIFF
--- a/.github/workflows/check-links.yaml
+++ b/.github/workflows/check-links.yaml
@@ -1,67 +1,30 @@
 ###############################################################################
-# Introduction, specific to nmdc-schema:
-# --------------------------------------
-# This workflow file is based upon the Baler 2.0.4 sample workflow file:
-# https://github.com/caltechlibrary/baler/blob/main/sample-workflow.yml
-# You can compare them in a text editor to see how they differ.
-#
-# In short, this workflow builds the schema documentation, then checks it for
-# broken links. If it finds any, it creates a GitHub Issue listing them.
+# Introduction:
+# -------------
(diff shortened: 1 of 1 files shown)


sujaypatil96 (approved):
LGTM 😁 Looks like we're switching from using Baler to Lychee in this PR, but everything seems to be working nicely on your fork @eecavanna and it's also creating a GitHub issue with the report from the Lychee run, so I'm very optimistic and excited for this PR to be merged in.

eecavanna:
Thanks, @sujaypatil96. I am excited, too. Once it is merged in, any of us will be able to run it on demand (via the "Actions" tab here on GitHub).

Once it is merged in and has been run, we can create an ticket about tweaking `lychee`'s configuration/parameters so that it doesn't report so many (or any) false positives. 

Once the configuration has been tweaked like that, we could make it so that the GHA workflow is triggered by more events, such as PRs being created or merged. That way, the link checking would happen in response to code changes and not just when people click a button on the GitHub GUI.

PR #1837, opened 2024-03-08
Implement preliminary migrator for PR1791

eecavanna:
I implemented this migrator as a reference for @aclum and @chienchi. Background information is here: https://github.com/microbiomedata/metaMAGs/issues/22#issuecomment-1984926129

I am putting this in a (draft) pull request—as opposed to just leaving it in a PR-less branch—so that people can comment on it here on GitHub.

Based on the `nmdc-schema` Release history, I think PR1791 was introduced in Release `10.1.0`; which would make the `_from_schema_version` the previous one, which was `10.0.0`. Something to confirm before merging.


eecavanna:
The updated migrator, itself, looks good to me—both the code and the file name. I approve of both of those things.

I'll defer to others on whether it fully compensates for the associated schema changes.

I think @turbomam also has experience evaluating whether a migrator does so.

aclum (approved):

PR #1924, opened 2024-04-12
Update identifiers.md

aclum:
Updating documentation on legacy identifiers including mapping between old and new identifiers.

I need this reviewed & merged so we can post an announcement on the data portal about re-iding.


SamuelPurvine:
@aclum is there any use in documenting to where the EMSL alternative identifier is pointing?

aclum:
@SamuelPurvine the emsl prefix is not currently resolvable and the my_emsl prefix is documented in the schema here:https://github.com/microbiomedata/nmdc-schema/blob/004d29a132846a7b1d3b75f86ea7237e93bf992e/src/schema/nmdc.yaml#L97C3-L97C10 
Having UI documentation for how nmdc expands prefixes I believe should be a linkml request since I don't see how you could do that with the current documentation. Please correct me if I'm wrong @turbomam @sujaypatil96 @cmungall

mbthornton-lbl (approved):

PR #1926, opened 2024-04-15
wrap `s.range` from class diagram jinja in `gen.name()`

sujaypatil96:
Sync with https://github.com/linkml/linkml/pull/2065


sujaypatil96:
Merge after tests pass

PR #1927, opened 2024-04-16
Display single/multi-valued (cardinality) on slot edges

sujaypatil96:
Mermaid UML class diagrams displaying cardinality on edges.


sierra-moxon (approved):
🎉

PR #1935, opened 2024-04-19
Update pypi-publish.yaml for trusted publishing

### LINKED ISSUE #1930: Use PyPI trusted publisher settings in release publishing workflow
We should set up the `nmdc-schema` PyPI project's [trusted publisher settings](https://docs.pypi.org/trusted-publishers/) so that we're not publishing with anyone's personal API token anymore.

pkalita-lbl:
Fixes #1930 

The necessary settings have already been updated in PyPI. These changes to the PyPI publishing workflow finish the job. For more information see https://docs.pypi.org/trusted-publishers/.

### DIFF
--- a/.github/workflows/pypi-publish.yaml
+++ b/.github/workflows/pypi-publish.yaml
@@ -8,6 +8,10 @@ jobs:
   build-n-publish:
     name: Build and publish Python 🐍 distributions 📦 to PyPI
     runs-on: ubuntu-latest
+    environment: release
+    permissions:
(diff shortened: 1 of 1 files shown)


eecavanna:
In order to be able to publish via GHA in the `berkeley-schema-fy24` repository, I think we will need to register `berkeley-schema-fy24` with PyPI as a "trusted publisher" (in addition to registering `nmdc-schema` as one, which I think has already happened).

Issue #1940, opened 2024-04-24
Create declarative mapping file for NMDC-to-NCBI push


sujaypatil96:
Similar to the "NCBI Postgres database column to NMDC slot names" mapping file that we have in `nmdc-schema`: https://github.com/microbiomedata/nmdc-schema/blob/main/assets/ncbi_mappings/ncbi_pg_db_field_mappings_filled.tsv already, we need another declarative mapping file that maps "NMDC slot names to all NCBI BioSample harmonized names".

The NCBI BioSample Attributes XML file can be found here: https://www.ncbi.nlm.nih.gov/biosample/docs/attributes/

The declarative mapping file, which will also be a TSV file similar to _ncbi_pg_db_field_mappings_filled.tsv_ will map NMDC schema slots (from classes Biosample, Extraction, LibraryPreparation, OmicsProcessing, DataObject) to values in the harmonized_name/synonym/title column from the above table. Since the range of some of the measurement slots in NMDC schema follow an object pattern (as opposed to scalar values in NCBI and MIxS) we need to compose the slots using the dot notation on slots.

Example:

| nmdc_schema_class | nmdc_schema_slot | nmdc_schema_slot_range | ncbi_biosample_attribute_name | static_value | ignore |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| Biosample | depth | QuantityValue | depth | |

Generic handling for nested data types in NMDC schema (to compose flat NCBI BioSample field values):
* QuantityValue
  * {slot_name}.has_numeric_value {slot_name}.has_unit, or
  * ( {slot_name}.has_maximum_numeric_value - {slot_name}.has_minimum_numeric_value ) {slot_name}.has_unit, or
  * <s>{slot_name}.has_raw_value</s>
* TextValue
  * {slot_name}.has_raw_value
* TimestampValue
  * {slot_name}.has_raw_value
* ControlledTermValue
  * {slot_name}.term.name [{slot_name}.term.id] , or
  * {slot_name}.term.id, or
  * {slot_name}.term.name, or
  * <s>{slot_name}.has_raw_value</s>
* ControlledIdentifiedTermValue
  * {slot_name}.term.name [{slot_name}.term.id], or
  * <s>{slot_name}.has_raw_value</s>
* GeolocationValue
  * {slot_name}.latitude {slot_name}.longitude, or
  * <s>{slot_name}.has_raw_value</s>

[...]

PR #1945, opened 2024-04-25
Ensure that `src/scripts/ncbi_postgres_nmdc_exact_term_matching.py` reads credentials from `local/.env`

sujaypatil96:
This PR is a continuation of the work in #1931 

Things that are being addressed in this PR:
* renaming `ncbi_nmdc_exact_term_matching.py` --> `ncbi_postgres_nmdc_exact_term_matching.py`
  * reason: there will be another TSV mapping file coming in from #1940 and the script that produces that mapping file should be called ncbi_nmdc_exact_term_matching.py because that maps fields/slots from the NCBI Attributes XML (not the NCBI postgres database) to NMDC slots
* make sure `src/scripts/ncbi_postgres_nmdc_exact_term_matching.py` follows best practices and reads NCBI postgres database connection credentials from `local/.env` file


assets/misc/.env.template
```diff
@@ -13,3 +13,6 @@ DEST_MONGO_PASS=
 DEST_MONGO_HOST=
 DEST_MONGO_PORT=
 DEST_MONGO_AUTH_SUFFIX=authSource=admin&readPreference=primary&directConnection=true&ssl=false
+
+NCBI_POSTGRES_USER=
```
turbomam, line 17: I'm not crazy about these constant names. They imply that they provide access to a Postgres server maintained by NCBI. Really, it's  Postgres database containing NCBI Biosample data, maintained by the BBOP group/NMDC

I'll try to think of something more helpful

If you want to change them now, how about just POSTGRES_* ?

turbomam (changes requested):

assets/misc/.env.template
sujaypatil96: Everything we are doing is in the context of NMDC, with that context in mind, I think it's possible that just POSTGRES_* might lead some people (users, developers) to be confused between the Postgres database that `nmdc-server` (data portal) read from, and the Postgres database containing NCBI BioSample data?

How about something like NCBI_DB_AS_POSTGRES_*?

turbomam: and no mention of biosample?

sujaypatil96: Oh right, sorry, missed that. How about NCBI_BIOSAMPLE_DB_AS_POSTGRES_*?

turbomam: thanks for brainstorming! how would you feel about BIOSAMPLES_RDB_PW

It's the password for a relational database of biosamples

turbomam: I asked ChatGPT for suggestions :-)

turbomam (approved):

PR #1950, opened 2024-05-02
Range for slots in slots table on class documentation pages account ofor `any_of`/`exactly_one_of`

sujaypatil96:
In its current state, classes which have slots that get modified to have ranges as unions (using `any_of`) like https://microbiomedata.github.io/nmdc-schema/Extraction/ do not show the modified ranges correctly in the slots table.

Merging this PR in will start rendering the range, for say the `has_input` slot on the above class documentation page correctly. Similar pattern will be applied to all applicable classes.

### DIFF
--- a/src/doc-templates/class.md.jinja2
+++ b/src/doc-templates/class.md.jinja2
@@ -8,6 +8,19 @@
     {%- endif -%}
 {%- endif -%}
 
+{% macro compute_range(slot) -%}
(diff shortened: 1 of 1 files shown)


turbomam (approved):

turbomam:
thanks @sujaypatil96 

after this we will have to make additional changes so that the `ProcessedSample` class page includes the usage in which an `Extraction` is allowed to put a `ProcessedSample` in it's `has_input`

PR #1976, opened 2024-05-15
add target for distributing prefix map as well

### LINKED ISSUE #1925: generate prefixmap for nmdc
We can generate a prefix map for all the NMDC schema prefixes currently supported with their expansions using this linkml command: 

`poetry run gen-prefix-map src/schema/nmdc.yaml` 

per @aclum's request in PR #1924

sierra-moxon:
fixes #1925

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,6 +113,7 @@ gen-project: $(PYMODEL) src/schema/mixs.yaml
 		--include owl \
 		--include python \
 		--include rdf \
+		--include prefixmap \
 		-d $(DEST) $(SOURCE_SCHEMA_PATH) && mv $(DEST)/*.py $(PYMODEL)
 		cp project/jsonschema/nmdc.schema.json  $(PYMODEL)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -161,6 +162,8 @@ gendoc: $(DOCDIR)
 	# added copying of images and renaming of TEMP.md
 	cp $(SRC)/docs/*md $(DOCDIR) ; \
 	cp -r $(SRC)/docs/images $(DOCDIR) ; \
+	$(RUN) gen-prefix-map -o project/prefixmap/nmdc-prefix-map.json $(SOURCE_SCHEMA_PATH) ; \
```
sujaypatil96, line 165: Just a thought for an improvement -- should we have `gen-prefix-map` also be able to return a Markdown file with the same results like the JSON would have, but instead of key-value pairs we would have a table of the prefixes?

That way we can have mkdocs convert the Markdown file into HTML and have it be shown as part of our web doc pages.

Definitely not a priority, but a nice to-have I would say.

sujaypatil96: Maybe add an option `--format` have it be restricted to `[json|md]`

sierra-moxon: @sujaypatil96 - I can see how that would be a nice doc; I do think the purpose of these is primarily programmatic.

turbomam (approved):

eecavanna (approved):

turbomam:
I'm merging this. @sujaypatil96 and @sierra-moxon do you think your mini-discussion should be turned into another issue?

turbomam:
oops, I didn't see that this PR was to merge the new content into `main-old`

PR #1984, opened 2024-05-16
Update eluent introduction description

mslarae13:
no migration needed


mslarae13:
@turbomam All i changed was moving the examples that were in the description into the examples for eluent_introduction_category. ... why is there so

PR #1985, opened 2024-05-17
Restore dump_single_modality version of pure-export

turbomam:
This adds to discreet modes to `pure-export`: one that exclusively dumps from an API and one that exclusively dumps from a MongoDB. The previous state was an awkward hybrid.

Most but not all of the command line options remain the same

see the two `local/mongo_as_unvalidated_nmdc_database.yaml` targets in `project.Makefile`. (I used the same name for each and just commented out the second)


nmdc_schema/dump_single_modality.py
```diff
@@ -0,0 +1,330 @@
...
+
+    ctx.obj['SCHEMA_HANDLER'] = SchemaHandler(schema_source)
+    ctx.obj['SELECTED_COLLECTIONS'] = selected_collections
+    ctx.obj['OUTPUT_YAML'] = output_yaml
+    ctx.obj['MAX_DOCS_PER_COLL'] = max_docs_per_coll
+
+
+# Sub-command for API access
+@cli.command()
+@click.option('--client-base-url', default="https://api.microbiomedata.org", show_default=True,
```
eecavanna, line 194: I see the default value here is the production Runtime API URL.

project.Makefile
```diff
@@ -197,9 +206,30 @@ local/mongo_as_unvalidated_nmdc_database.yaml:
 		--selected-collections read_based_taxonomy_analysis_activity_set \
 		--selected-collections read_qc_analysis_activity_set \
 		--selected-collections study_set \
-		--selected-collections data_object_set \
-		--selected-collections field_research_site_set \
-		--skip-collection-check
+		--selected-collections workflow_chain_set \
+		--selected-collections workflow_execution_set \
+		api-access \
+		--client-base-url "https://api-napa.microbiomedata.org" \
```
eecavanna, line 212: I see the value here is the Napa Runtime API URL.

nmdc_schema/dump_single_modality.py
```diff
@@ -0,0 +1,330 @@
...
+            if slot_obj.multivalued or include_scalars:
+                acceptable_slots.append(i)
+            else:
+                logger.warning(
+                    f"Skipping {class_name} slot {i} because it is not multivalued and include_scalars is False.")
+        acceptable_slots.sort()
+        return acceptable_slots
+
+
+def set_arithmetic(set1, set2, set1_name='set 1 only', set2_name='set 2 only', intersection_name='intersection'):
```
eecavanna, line 47: I recommend using a verb as the function name.

Example:
```py
def determine_exclusive_and_common_elements():
    pass
```

nmdc_schema/dump_single_modality.py
```diff
@@ -0,0 +1,330 @@
...
+    intersection = set1.intersection(set2)
+
+    return {
+        set1_name: list(set1_only),
+        set2_name: list(set2_only),
+        intersection_name: list(intersection),
+    }
+
+
+def get_collection_stats(selected_collections, db):
```
eecavanna, line 79: I recommend adding a docstring that summarizes what the function does.

nmdc_schema/dump_single_modality.py
```diff
@@ -0,0 +1,330 @@
...
+    ctx.obj['OUTPUT_YAML'] = output_yaml
+    ctx.obj['MAX_DOCS_PER_COLL'] = max_docs_per_coll
+
+
+# Sub-command for API access
+@cli.command()
+@click.option('--client-base-url', default="https://api.microbiomedata.org", show_default=True,
+              help='HTTP(S) path to the FastAPI server.')
+@click.option('--endpoint-prefix', default="nmdcschema", show_default=True,
+              help='FastAPI path component between the URL and the endpoint name')
```
eecavanna, line 197: I see the fact that the API is built on the FastAPI framework to be an implementation detail of the API. Mentioning it here couples this tool to that implementation detail more than I want them to be coupled. I recommend changing it to just "API":
```diff
- FastAPI
+ API
```

project.Makefile
turbomam: do you have a request or suggestion?

eecavanna (approved):

project.Makefile
eecavanna: I was surprised to see the Napa environment's URL being hard-coded anywhere (other than in re-ID-ing-specific code), since it is a "temporary" environment.

eecavanna: No particular suggestion. I'm OK with it being the Napa environment—I just was surprised to see it (given that the _production_ environment was used as a default elsewhere in the diff) and wanted to make sure it was intentional.

turbomam: I guess I should switch the usage example to the production endpoint. I'll do all of this Monday. THanks

turbomam:
thanks @eecavanna. I applied those changes. I think there's a bug in there, and the output is coming out malformed. As a result, `migration-recursion` is crashing.

I going to work on fixing my output, but I would also like to hear any feedback from you and @brynnz22 on what assumptions are made by `migration-recursion` regarding the structure or its inputs. I plan on taking  a LLM approach to figuration that out eventually and will let you know if I make progress.

turbomam:
My code (in API mode) was writing a list of documents, not a dictionary of lists of documents

aclum (approved):

nmdc_schema/dump_single_modality.py
```diff
@@ -0,0 +1,342 @@
...
+# Sub-command for API access
+@cli.command()
+@click.option('--client-base-url', default="https://api.microbiomedata.org", show_default=True,
+              help='HTTP(S) path to the API server.')
+@click.option('--endpoint-prefix', default="nmdcschema", show_default=True,
+              help='API path component between the URL and the endpoint name')
+@click.option('--page-size', default=200_000, show_default=True,
+              help='Size of each page in pagination.')
+@click.pass_context
+def api_access(ctx, client_base_url, endpoint_prefix, page_size):
```
mbthornton-lbl, line 203: Is there a clearer verb-noun type of name for these functions?  `dump_from_api` / `dump_from_db` or something like that

nmdc_schema/dump_single_modality.py
```diff
@@ -0,0 +1,342 @@
...
+@click.option('--mongo-db-name', default="nmdc", show_default=True, help='MongoDB database name')
+@click.option('--mongo-host', default="localhost", show_default=True,
+              help='MongoDB host name/address.')
+@click.option('--mongo-port', default=27777, type=int, show_default=True, help='MongoDB port')
+@click.option('--admin-db', default=None, help='MongoDB authentication source. Leave blank to not specify.')
+@click.option('--auth-mechanism', default=None, show_default=True,
+              help='Authentication mechanism for MongoDB connection. Leave blank to not specify.')
+@click.option('--direct-connection/--no-direct-connection', default=True,
+              help='Whether to use a direct connection to MongoDB. Defaults to direct connection.')
+def pymongo_access(ctx, env_file, mongo_db_name, mongo_host, mongo_port,
```
mbthornton-lbl, line 268: Another function naming comment - I think the salient point about this function is that it is directly from the DB, not mediated by the API. How it's implemented, via `pymongo` or some other means is something the user shouldn't have to be concerned with

mbthornton-lbl (changes requested):

nmdc_schema/dump_single_modality.py
turbomam: Thanks. See below.

nmdc_schema/dump_single_modality.py
turbomam: sorry, I see you had proposed different `cli.command` names. I went with `dump_from_api` and `dump_from_database`

turbomam: I'm going to merge now but am open to more revision if necessary.

PR #2012, opened 2024-05-28
Identify redundant EMP500 records and generate queries:run delete commands

aclum:
adding code to find workflow records that have the same was_informed_by values, find their data dependencies, and make queries: run deletion requests for duplicates. See https://github.com/microbiomedata/nmdc_automation/issues/162. This cleans up extra records for EMP500

### DIFF
--- a/nmdc_schema/identify_workflow_duplicates_emp500.py
+++ b/nmdc_schema/identify_workflow_duplicates_emp500.py
@@ -6,7 +6,8 @@ from pymongo.errors import ConnectionFailure
 from dotenv import load_dotenv
 import urllib
 
-# this code deletes multiple workflow records from EMP500 which have different base IDs with a .1 version.
+# this is generating a .json data file that can be used to delete these records by submitting it to the appropriate endpoint (ie queries:run)
+# for EMP500 (nmdc:sty-11-547rwq94) which have different base IDs with a .1 version.
 
 envfile_path = "../../.env.client"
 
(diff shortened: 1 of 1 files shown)


nmdc_schema/identify_workflow_duplicates_emp500.py
```diff
@@ -0,0 +1,214 @@
+import json
+import pymongo
+import os
+from pymongo import MongoClient
+from pymongo.errors import ConnectionFailure
+from dotenv import load_dotenv
+import urllib
+
+# this code deletes multiple workflow records from EMP500 which have different base IDs with a .1 version.
```
mbthornton-lbl, line 9: This should be more accurate - this is generating a .json data file that can be used to delete these records by submitting it to the appropriate endpoint

nmdc_schema/identify_workflow_duplicates_emp500.py
```diff
@@ -0,0 +1,214 @@
...
+    )
+    for rbt_dup in rbt_duplicate_cursor:
+        rbt_delete.append(rbt_dup["id"])
+        data_objects_to_delete(rbt_dup["has_output"])
+
+
+for doc in test_list:
+    omics_id = doc["_id"]
+    records = pick_records(omics_id)
+    if len(list(records)) == 1:
```
mbthornton-lbl, line 152: what should happen if len(list(records)) != 1?

mbthornton-lbl (changes requested):

nmdc_schema/identify_workflow_duplicates_emp500.py
aclum: updated

nmdc_schema/identify_workflow_duplicates_emp500.py
aclum: added else print statement

mbthornton-lbl (approved):

Issue #2010, opened 2024-05-28
Migrator: Write migrator for `NomAnalysisActivity` documents


aclum:
We need a migrator which will search for NomAnalysisActivity which do not have a version appended, for each of those records it should update ID to append a .1 to the existing value in slot ID and move the existing value of ID to alternative_identifiers.

Example before:
```json
{
  "_id": {
    "$oid": "649b0095daa6c19f56b2777c"
  },
  "type": "nmdc:NomAnalysisActivity",
  "has_input": [
    "nmdc:dobj-13-hg0x7944"
  ],
  "has_output": [
    "nmdc:dobj-13-48nyp930"
  ],
  "id": "nmdc:wfnom-13-7yf9qj85",
  "ended_at_time": "2021-01-21T23:27:57Z",
  "execution_resource": "EMSL-RZR",
  "git_url": "https://github.com/microbiomedata/enviroMS",
  "started_at_time": "2021-01-21T23:27:57Z",
  "used": "12T_FTICR_B",
  "was_informed_by": "nmdc:omprc-11-3x68c186"
}
```

example after:
```json
{
  "_id": {
    "$oid": "649b0095daa6c19f56b2777c"
  },
  "type": "nmdc:NomAnalysisActivity",
  "has_input": [
    "nmdc:dobj-13-hg0x7944"
  ],
  "has_output": [
    "nmdc:dobj-13-48nyp930"
  ],
  "id": "nmdc:wfnom-13-7yf9qj85.1",
  "ended_at_time": "2021-01-21T23:27:57Z",
  "execution_resource": "EMSL-RZR",
  "git_url": "https://github.com/microbiomedata/enviroMS",
  "started_at_time": "2021-01-21T23:27:57Z",
  "used": "12T_FTICR_B",
  "was_informed_by": "nmdc:omprc-11-3x68c186",
  "alternative_identifiers": ["nmdc:wfnom-13-7yf9qj85"]
}
```

Example migrators can be found https://github.com/microbiomedata/nmdc-schema/tree/main/nmdc_schema/migrators

Target completion for this is 6/17. This migrator is needed for the 6/24 release or the records will be invalid b/c that release will have more stringent pattern matches on IDs.  cc @ssarrafan

eecavanna:
[...]

PR #2059, opened 2024-06-11
Add migrator for `nom_analysis_activity` ID

JamesTessmer:
This migrator will be used to append version numbers to IDs that currently have none.

Add to resolve Issue https://github.com/microbiomedata/nmdc-schema/issues/2010


nmdc_schema/migrators/migrator_from_10_3_0_to_10_4_0.py
```diff
@@ -0,0 +1,33 @@
...
+
+        >>> m = Migrator()
+        >>> m.update_nom_analysis_activity_id({'id': '1'}) # test: adds version number to id
+        {'id': '1.1'}
+        """
+        self.logger.info(f"Processing NomAnalysisActivity: {nom_analysis_activity['id']}")
+
+        
+        # If this data object's `data_object_type` consists of the invalid value, replace it with the valid value.
+        if "." not in nom_analysis_activity['id']:
```
eecavanna, line 30: I am wondering whether this is a sufficient check. In other words, I am wondering whether it's possible that a "non-versioned" ID still contains a period for an unrelated reason.

nmdc_schema/migrators/migrator_from_10_3_0_to_10_4_0.py
```diff
@@ -0,0 +1,33 @@
...
+        )
+
+    def update_nom_analysis_activity_id(self, nom_analysis_activity: dict) -> dict:
+        r"""
+        Updates a nom_analysis_activity's id to contain a version number (.1),
+        if it has no version number.
+
+        >>> m = Migrator()
+        >>> m.update_nom_analysis_activity_id({'id': '1'}) # test: adds version number to id
+        {'id': '1.1'}
```
eecavanna, line 24: Will you add an example for where the document's ID already ends with `.1`?

```suggestion
        {'id': '1.1'}
        >>> m.update_nom_analysis_activity_id({'id': '1.1'}) # test: no change
        {'id': '1.1'}
```

nmdc_schema/migrators/migrator_from_10_3_0_to_10_4_0.py
```diff
@@ -0,0 +1,33 @@
...
+        """
+        self.logger.info(f"Processing NomAnalysisActivity: {nom_analysis_activity['id']}")
+
+        
+        # If this data object's `data_object_type` consists of the invalid value, replace it with the valid value.
+        if "." not in nom_analysis_activity['id']:
+            nom_analysis_activity['id'] = str(nom_analysis_activity['id']) + ".1"
+
+        return nom_analysis_activity
```
eecavanna, line 33: I think one of the requirements was to put the original `id` into a field named `alternative_identifiers`.

eecavanna (commented):
Hi @JamesTessmer, thanks for implementing this. I left some feedback comments.

nmdc_schema/migrators/migrator_from_10_3_0_to_10_4_0.py
eecavanna: Here's a regex pattern you can use to determine whether the ID "ends with a dot followed by one or more numeric digits":

```py
pattern = re.compile(r"\.[0-9]+$")  # ends with dot and number(s)
```

nmdc_schema/migrators/migrator_from_10_3_0_to_10_4_0.py
eecavanna: I'd also like to see a counterexample where the document's ID contains a dot that is not at the end. That would serve as documentation about the migrator's behavior for such a corner case.

For example:

```py
>>> m.update_nom_analysis_activity_id({'id': 'nmdc:foo-12.3-potato'}) # test: no change
{'id': 'nmdc:foo-12.3-potato'}
```

nmdc_schema/migrators/migrator_from_10_3_0_to_10_4_0.py
eecavanna: Here's how I did it in the example one: https://gist.github.com/eecavanna/fc1e20dc919f87746c8668e86ea68a4a#file-migrator_from_10_3_0_to_10_4_0-py-L42-L61

nmdc_schema/migrators/migrator_from_10_3_0_to_10_4_0.py
```diff
@@ -0,0 +1,33 @@
...
+
+        self.adapter.process_each_document(
+            "nom_analysis_activity_set", [self.update_nom_analysis_activity_id]
+        )
+
+    def update_nom_analysis_activity_id(self, nom_analysis_activity: dict) -> dict:
+        r"""
+        Updates a nom_analysis_activity's id to contain a version number (.1),
+        if it has no version number.
+
```
eecavanna, line 21: Great description!

nmdc_schema/migrators/migrator_from_10_3_0_to_10_4_0.py
```diff
@@ -0,0 +1,33 @@
...
+        >>> m = Migrator()
+        >>> m.update_nom_analysis_activity_id({'id': '1'}) # test: adds version number to id
+        {'id': '1.1'}
+        """
+        self.logger.info(f"Processing NomAnalysisActivity: {nom_analysis_activity['id']}")
+
+        
+        # If this data object's `data_object_type` consists of the invalid value, replace it with the valid value.
```
eecavanna, line 29: I think this comment is out-of-date.

JamesTessmer:
> Hi @JamesTessmer, thanks for implementing this. I left some feedback comments.

Thanks for all the feedback @eecavanna ! I'll get this all done today.

JamesTessmer:
@eecavanna I've updated everything with the examples you showed and your suggestions.
I'm marking this as ready for review

nmdc_schema/migrators/migrator_from_10_3_0_to_10_4_0.py
```diff
@@ -0,0 +1,82 @@
...
+            "nom_analysis_activity_set", [self.populate_alternative_identifiers,self.update_nom_analysis_activity_id]
+        )
+    
+    def is_valid_id(self, s: str) -> bool:
+        r"""
+        Helper function that checks whether a string is a valid ID (has a version number).
+        Reference: https://docs.python.org/3/library/re.html#re.Pattern.search
+
+        >>> m = Migrator()
+        >>> m.is_valid_id("invalid_id")
```
aclum, line 28: I think this will work but next time we should use the materialized patterns generated by running `make all` which generates `project/nmdc_materialized_patterns.yaml` (^(nmdc):wfnom-([0-9][a-z]{0,6}[0-9])-([A-Za-z0-9]{1,})(\.[0-9]{1,})$)

JamesTessmer: Ah, I didn't know about the patterns, thanks for pointing that out!
I can add that pattern to the regex instead.

JamesTessmer: Hmm, when I plug that in to the regex and test it says IDs without version numbers are still valid.

JamesTessmer: Ah, the issue is a period character is a wildcard so it's just seeing that. If I escape it with a backslash it does work as intended.

nmdc_schema/migrators/migrator_from_10_3_0_to_10_4_0.py
```diff
@@ -0,0 +1,82 @@
...
+
+from nmdc_schema.migrators.migrator_base import MigratorBase
+
+
+class Migrator(MigratorBase):
+    r"""Migrates a database between two schemas.
+    """
+
+    _from_version = "10.3.0"
+    _to_version = "10.4.0"
```
eecavanna, line 11: FYI @aclum: These lines, and the name of the migrator Python file, are the two places that would require changing if you wanted this migrator to indicate that it was associated with different schema versions.

eecavanna:
I will update this branch to reference different schema version numbers. New commit(s) coming soon.

eecavanna:
@JamesTessmer, do you know what the schema version is in which the **schema changes** associated with this migrator were introduced in? Alternatively, do you know what the PR is in which those schema changes were introduced into the repo? I'm trying to determine what the correct "to" and "from" versions are for this migrator.

CC: @turbomam

eecavanna:
@JamesTessmer, do you know the version number of the schema release in which the schema changes associated with this migrator were introduced? Alternatively, do you know which PR those schema changes were introduced into the repo in? I'm trying to determine what the correct "to" and "from" versions are for this migrator. I assumed the 10.3.0 and 10.4.0 (which I have since replaced) were placeholders.

CC: @turbomam

JamesTessmer:
I have no idea where the changes for these come from. In the issue (https://github.com/microbiomedata/nmdc-schema/issues/2010) this PR is for @aclum said the 'from' version was 10.3.0

eecavanna:
OK, thanks, @JamesTessmer. I'll work with @aclum and @turbomam tomorrow to identify the version numbers, then I'll put them into this branch.

turbomam (approved):

eecavanna:
I'll merge this in once the GHA checks pass.

PR #2083, opened 2024-06-20
Use `--include-range-class-descendants` flag in JSON Schema generator calls

pkalita-lbl:
These changes are related to the issues uncovered while trying to integrate `nmdc-schema@10.5.4` into `nmdc-runtime` (https://github.com/microbiomedata/nmdc-runtime/pull/560)

The immediate cause of the problem was that the `WorkflowExecutionActivity` class became abstract while still being used as the `range` of the `activity_set` slot. This isn't necessarily wrong from a modeling perspective, but it does cause some confusing in LinkML's JSON Schema generator. The generator is more equipped to handle this situation when the `--include-range-class-descendants` flag is used. This flag instructs the generator to do more work in flattening out the class hierarchy in defined in the LinkML schema when going to the JSON Schema representation (which has no concept of class hierarchy). Arguably this should be the default for the JSON Schema generator, but I need to work that out on the LinkML side.

These changes also add a new unit test that make sure that both JSON Schema artifacts generated from the schema are compatible with the `fastjsonschema` library that `nmdc-runtime` uses. This will give us an earlier failure if something similar crops up again.


turbomam (approved):
I haven't run this locally, but since it passes the GHA checks and if you think the test really tests what needs to be tested, then I approve!

PR #2087, opened 2024-06-25
Render `Mappings` section generally on all element documentation pages

sujaypatil96:
Analogous to https://github.com/linkml/linkml/pull/2172


sujaypatil96:
Converting this PR to draft because the GitHub Action workflow needs to be fixed in linkml-project-cookiecutter and then the changes need to be applied here: https://github.com/linkml/linkml-project-cookiecutter/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/.github/workflows/test_pages_build.yaml to keep both of them in sync

turbomam (commented):
doesn't change schema but does improve understand-ability of the schema through the documentation. I recommend merging into both schema now.

sujaypatil96:
This PR is ready to be approved and merged.

See these two pages for example:
* https://microbiomedata.github.io/nmdc-schema/objective/ (no mappings section)
* https://microbiomedata.github.io/nmdc-schema/pr-preview/pr-2087/objective/#mappings (mappings section present)

turbomam:
https://microbiomedata.github.io/nmdc-schema/pr-preview/pr-2087/objective/#mappings

404 - Not found

turbomam:
can't even load https://microbiomedata.github.io/nmdc-schema/pr-preview/pr-2087

sujaypatil96:
@turbomam the above links are working again.

What I did was to re-trigger the PR preview GitHub Action workflow/run on this PR again: https://github.com/microbiomedata/nmdc-schema/actions/runs/11350934707/job/31644169923

However, if this PR preview Action is running on other PRs then I believe it might overwrite this one, and it will preserve whatever is latest so it can avoid "overpopulating" the gh-pages branch.

sujaypatil96:
I'm not entirely sure. I can look through some documentation and try to see if I can understand what _self_ and _native_ signify exactly, but here is the schemaview method that is responsible for what is rendered in the _Mappings_ section: https://github.com/linkml/linkml-runtime/blob/main/linkml_runtime/utils/schemaview.py#L1171-L1199

turbomam:
Let's talk about this again soon. I wrote some code to dump all mappings in the schema, so that gave me a chance to review concepts like native, self and undefined.

My goal is to emphasize the most useful parts of the mappings and omit the things that are obvious, can be taken for granted, etc. I think the self and native mappings are usually not very enlightening, especially if they are the same, and/or if they are equivalent to the default prefix + the element name. 

I think i prefer the term "unqualified" or just "mappings" over "undefined", but we shouldn't go by my personal preference alone.

turbomam (approved):
approved. deal with these later
- https://github.com/microbiomedata/nmdc-schema/issues/2289 
- https://github.com/microbiomedata/nmdc-schema/pull/2087

PR #2091, opened 2024-06-28
Allow `nmdc_data.py` consumer to specify which variant of schema they want to access

eecavanna:
### Summary of changes

> Note: This summary was copied from [here](https://github.com/microbiomedata/berkeley-schema-fy24/pull/204).

In this branch, I updated `nmdc_data.py` so that the fuctions that provide access to the JSON Schema version of the schema, allow the caller to specify which _variant_ of the schema they want (they currently only support the default variant and the `nmdc_materialized_patterns` variant).

That will allow callers (such as the Runtime and the migration notebooks) to access the `nmdc_materialized_patterns.schema.json` JSON Schema (which they can use to perform data validation).

I also added doctests to all modified functions to demonstrate that they return values of the types I expect. The doctests can be run via `$ poetry run python -m doctest nmdc_schema/nmdc_data.py` (or, now, `$ make test-python`).

---

### Note for reviewers

- This PR contains the same changes as https://github.com/microbiomedata/berkeley-schema-fy24/pull/204 (in fact, the commits on this branch were `git cherry-pick`-ed directly from the branch in that PR). This PR is the `nmdc-schema` equivalent of that `berkeley-schema-fy24` PR.
- I tagged the two people that reviewed the `berkeley-schema-fy24` PR, as reviewers of this PR. I also tagged someone else that has worked on the `nmdc_data.py` file in the past.

---

### Note regarding schema freeze

- This branch does not contain any schema changes.
- This branch _does_ add functionality that is available to downstream things that `import` the `nmdc_data.py` script (e.g. the Runtime), but it doesn't break any existing functionality. It is "opt in" functionality.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,6 +141,7 @@ test-schema:
 
 test-python:
 	$(RUN) python -m unittest discover
+	$(RUN) python -m doctest nmdc_schema/nmdc_data.py
 
 lint:
 	$(RUN) linkml-lint $(SOURCE_SCHEMA_PATH) > local/lint.log
--- a/nmdc_schema/nmdc_data.py
+++ b/nmdc_schema/nmdc_data.py
@@ -22,7 +22,11 @@ class SchemaVariantIdentifier(str, Enum):
 
     >>> type(SchemaVariantIdentifier.nmdc_materialized_patterns) is SchemaVariantIdentifier
     True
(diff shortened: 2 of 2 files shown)


pkalita-lbl (approved):

turbomam (approved):

mslarae13:
Approved during berk-schema meeting check in. 
Only change, default should include/ require materialized patterns & you have the option to say no but we will split that out from this PR.

PR #2097, opened 2024-07-01
Use newer version of `urllib3` to avoid moderate security bug

turbomam:
GitHub's dependabot recently [reported an issue](https://github.com/microbiomedata/nmdc-schema/security/dependabot/11) in the `urllib3` Python package upon which this project depends. According to dependabot, the issue is present in `urllib3` version 2.2.1, but not 2.2.2. In this branch, I ran `poetry update urllib3` to update `urllib3` to a version newer than 2.2.1, and committed the resulting `poetry.lock` file.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -122,13 +122,13 @@ lxml = ["lxml"]
 
 [[package]]
 name = "bioregistry"
-version = "0.11.7"
+version = "0.11.10"
 description = "Integrated registry of biological databases and nomenclatures"
 optional = false
 python-versions = ">=3.8"
 files = [
-    {file = "bioregistry-0.11.7-py3-none-any.whl", hash = "sha256:3167ea4ec19c220df803823c38bce5a5b573cfa24b6ea85bd883c75a0b0e164e"},
-    {file = "bioregistry-0.11.7.tar.gz", hash = "sha256:64c1f64fd872aa9a623b9ec47ce76941dc163ad69dd3ef1633c9a44217eb546d"},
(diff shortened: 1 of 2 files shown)


eecavanna:
I'd rather not have a new direct dependency be introduced when the only reason it is being introduced is to specify that a specific version of it be installed (and it is otherwise a transitive dependency). With this approach, even if the direct dependencies switch to using something other than urllib3, `nmdc-scehma` will continue to "depend" on it.

eecavanna:
I think you can instead run:
```shell
poetry update urllib3
```
> Source: https://python-poetry.org/docs/cli/#update

And then commit the resulting `poetry.lock` file.

turbomam:
> I'd rather not have a new direct dependency be introduced when the only reason it is being introduced is to specify that a specific version of it be installed (and it is otherwise a transitive dependency). With this approach, even if the direct dependencies switch to using something other than `urllib3`, `nmdc-schema` will continue to "depend" on it.

make sense @eecavanna. can you suggest an alternative?

eecavanna:
> can you suggest an alternative?

I posted a shell command (suggestion) right above your most recent comment (I think we posted those at roughly the same time).

eecavanna (approved):

eecavanna:
@turbomam, here's what I suggest for a PR description:

> GitHub's "dependabot" recently [reported an issue](https://github.com/microbiomedata/nmdc-schema/security/dependabot/11) in the `urllib3` Python package upon which this project depends. According to "dependabot," the issue is present in `urllib3` version 2.2.1, but not 2.2.2. In this branch, I ran `poetry update urllib3` to update `urllib3` to a version newer than 2.2.1, and committed the resulting `poetry.lock` file.
>
> This branch does not contain any schema changes.

mslarae13 (commented):
@eecavanna approved, and I'll defer to him.

PR #2113, opened 2024-07-09
1000 model rna library preparation

aclum:
This PR is to model how to capture RNA-seq library preparation information which is needed input for the expression step of the Metatranscriptome analysis workflow.

This adds two slots which are applied to Class LibraryPreparation. 
Changes are additive, no migrators needed.


turbomam:
Thanks @aclum! I prefer the explicitness and atomicity of the `is_stranded` and `stranded_orientation` style. That option only adds one slot beyond the other option, right?

src/schema/nmdc.yaml
```diff
@@ -1813,6 +1815,14 @@ classes:
       - https://casrai.org/credit/
 
 enums:
+
+  StrandedOrentationEnum:
```
eecavanna, line 1819: I think there is a typo in the name of this enum:

```suggestion
  StrandedOrientationEnum:
```

src/schema/nmdc.yaml
```diff
@@ -2532,6 +2542,17 @@ slots:
...
   
+  is_stranded:
+    description: Is the (RNA) library stranded or non-stranded (unstranded). 
+    range: boolean
+    comments: 
+      - A value of true means the library is stranded, flase means non-stranded. 
+  
+  stranded_orientation:
+    description: Lists the strand orientiation for a stranded RNA library preparation.
+    range: StrandedOrentationEnum
```
eecavanna, line 2553: I think there is a typo in the name of the enum:

```suggestion
    range: StrandedOrientationEnum
```

eecavanna (changes requested):
> Changes are additive, no migrators needed.

I agree. This schema change adds a non-required slot named `is_stranded` to an existing class, adds a non-required slot named `stranded_orientation` (whose range is an enum) to that same class, and defines that enum. None of that has the potential to make currently-valid data become invalid, which is what I consider to be the condition for whether a migrator is necessary or not.

I'm requesting a couple changes related to the spelling of "orientation". I don't have any other concerns about these changes.

eecavanna (approved):
All of my concerns have been resolved. I'm comfortable with this being merged in.

aclum:
Discussed at the workflows meeting and there were no concerns.

src/schema/nmdc.yaml
```diff
@@ -1813,6 +1815,14 @@ classes:
       - https://casrai.org/credit/
 
 enums:
+
+  StrandedOrientationEnum:
+    description: This enumeration specifies information about stranded RNA library preparations.
+    permissible_values:
+      antisense orientation:
```
turbomam, line 1822: actually, could we please add descriptions to these permissible values?

turbomam (approved):
nice

Issue #1881, opened 2024-03-29
DataPortal can be searched for `Pathway`s and `OntologyGroup`s... does that justify leaving them in nmdc-schema?


turbomam:
Good question. Suggested by @mslarae13

mslarae13:
I don't remember making this suggestion :)

turbomam:
whops! sorry @mslarae13

turbomam:
@bmeluch you and discussed something like this earlier this week, right? I recommend

bmeluch:
Yes we talked about it since those classes do not appear anywhere in the example data. I guess we just have to make sure no one has been incubating any plans for them?

turbomam:
Id there's time should we discuss this in the upcoming _NMDC Schema & Metadata_ meeting?

bmeluch:
From 03 July Metadata/ Data Schema / Data Management Meeting: these classes were never used, we aren't modeling these at this level of granularity, put in a PR to remove them. Team will review and decide if it can be merged before or after Berkeley schema rollout.

bmeluch:
Closing this issue, this was addressed in Berkeley, see discussion in the linked PR

bmeluch:
It's not in a comment but in the issue history, https://github.com/microbiomedata/berkeley-schema-fy24/pull/222

PR #2114, opened 2024-07-09
Deprecate `EnvironmentalMaterialTerm`, `FunctionalAnnotationTerm` + children

bmeluch:
See https://github.com/microbiomedata/nmdc-schema/issues/1881

This PR is the first step in deprecating `EnvironmentalMaterialTerm`, `FunctionalAnnotationTerm` + children. It adds a `deprecated` string to each class definition. This PR will need to be reviewed to see if/when it can be merged in during the Berkeley schema rollout, or if we need to wait.


bmeluch:
this was supposed to be created in berkeley schema fork, whoops

PR #2118, opened 2024-07-10
Create DataObject_gen_by_omics.yaml

aclum:
This PR adds example valid data to test that OmicsProcessing records are valid identifiers to use on DataObject was_generated_by.


sierra-moxon (approved):
approving!  

These are general questions to get me up to speed with PRs like this: 
- What are the naming conventions on the files themselves?
   - snake, kebab, camel case: is it CamelCase for class names, slot_names are snake cased, and dashes separate the two?  Should the class and slot names reflect valid model elements?
- Should we add "invalid" files that correspond to the "valid" files (for every valid example?  or just in certain cases?)

aclum:
The file naming convention varies but here are the general guidelines, if a DataObject is the output of OmicsProcessing then files are based on the source organization so there can be a lot of variability. Example `52818.1.460070.TGGCACTA-TGGCACTA.fastq.gz` is a sequencing record from JGI that is named based on a  $instrument_un_id.$flowcell_lane.$basecall_id,$barcode_index.$file_type.$compression. If the DataObject is generated by nmdc sequencing based workflows the convention is $nmdc_identifier_$filedescriptor$file_format.$compression. An underscore is always used in place of a colon after 'nmdc' because the workflow manager can't handle colons in file names (ie nmdc_wfrqc-11-c5w5df71.1_filtered.fastq.gz  is a gzip compressed fastq file which contains qc filtered Illumina data that was generated from a workflow activity with an ID of nmdc:wfrqc-11-c5w5df71.1). File names after the nmdc identifer can be camel or snakecase depending, NDMC as much as possible synchronizes file names with the corresponding user facility that we share the workflow with so there isn't case consistency.  Some of the EMSL based workflow results dont use nmdc identifiers, instead re-use EMSL file names (ex Brodie_381_R3_CHCl3_03May19_Alder_Infuse_IATp05_144SA_1_01_43877.csv)

I do normally try to make invalid data for all PRs but didn't in this case b/c there are no schema changes, I was just testing what I thought should be true about the allowed range for was_generated_by.

src/data/valid/Database-gen_from_omics.yaml
```diff
@@ -0,0 +1,28 @@
...
+  file_size_bytes: 1382304
+  md5_checksum: "22afa3d49b73eaec2e9787a6b88fbdc3"
+  name: "1237.14.3718.ATCG-ACTG.fastq.gz"
+  type: "nmdc:DataObject"
+  was_generated_by: "nmdc:omprc-99-123456"
+omics_processing_set:  
+ - id: nmdc:omprc-99-123456
+   alternative_identifiers:
+    - gold:Gp0108335
+   name:  Permafrost 712P3D
```
eecavanna, line 14: In this comment, I am suggesting the removal of a whitespace character I assume was included unintentionally.

```suggestion
   name:  Permafrost 712P3D
   name: Permafrost 712P3D   
```

eecavanna (approved):
Looks good to me. I left a comment about a whitespace character I thought might have been included by accident. 

Also, I think @sierra-moxon was referring to the name of the example data file: `DataObject-gen_by_omics.yaml` (I'm guessing the general structure is `{PascalCaseClassName}-{snake_case_slot_name}.yaml`). Note: "Pascal case" is like "Camel case," but the first letter of the variable name is also uppercase (as in `PascalCase`, as opposed to `camelCase`).

turbomam (approved):

naglepuff (approved):

mslarae13 (approved):

PR #2120, opened 2024-07-12
Add tooltip annotation for gold_path_field

aclum:
This PR adds a tool tip for slot gold_path_field. Description change only, no changes to example data. No migration needed.

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -2968,10 +2968,11 @@ slots:
     abstract: true
     description: >-
       This is a grouping for any of the gold path fields
+    
     annotations:
-      tool_tip: 
-        GOLD Ecosystem Classification paths describe the surroundings from which an environmental 
-        sample or an organism is collected.
+      tool_tip: GOLD Ecosystem Classification paths describe the surroundings from which an environmental sample or an organism is collected.
+    comments:
+      - Tool tip pull from https://gold.jgi.doe.gov/ecosystem_classification
   ecosystem:
     is_a: gold_path_field
     description:
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -2968,6 +2968,10 @@ slots:
     abstract: true
     description: >-
       This is a grouping for any of the gold path fields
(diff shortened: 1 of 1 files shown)


turbomam:
@aclum can you say where you got that definition?

You wanted me to do something similar for the MIxS environmental triad as a group, right?

mslarae13 (approved):

src/schema/nmdc.yaml
```diff
@@ -2968,6 +2968,10 @@ slots:
     abstract: true
     description: >-
       This is a grouping for any of the gold path fields
+    annotations:
+      tool_tip:
```
turbomam, line 2972: single line

aclum:
@turbomam  that description came from https://gold.jgi.doe.gov/ecosystem_classification, should i add that as comment?

aclum:
@turbomam  okay to merge this in? This was tentatively approved by the refactor folks yesterday.

turbomam:
I was hoping we could put the comment on the annotation itself. But that would require the use of a wordier [collection forms
](https://linkml.io/linkml-model/latest/docs/specification/06mapping/) and a nested annotation!

```yaml
    annotations:
      tool_tip:
        tag: tool_tip
        value: GOLD Ecosystem Classification paths describe the surroundings from which an environmental sample or an organism is collected.
        annotations:
          source: https://gold.jgi.doe.gov/ecosystem_classification
```

I will approve now. We can use either what I suggested above or what you already have.

turbomam (approved):

aclum:
I don't have a preference, I just didn't have an example to work off of. I'll update to this format now.

aclum:
@turbomam  updated the modeling to your preferred syntax.

src/schema/nmdc.yaml
```diff
@@ -2968,6 +2968,12 @@ slots:
...
```
eecavanna, line 2972: I don't think I've come across this spelling of "tooltip" before. I don't meant to say it's incorrect, just that I think I've only ever seen it spelled as a single word.

PR #2201, opened 2024-10-08
move valid `src/data/invalid/DissolvingProcess-minimal-no-solvent.yaml` -> `src/data/problem/invalid/DissolvingProcess-minimal-no-solvent.yaml`

turbomam:
I could have sworn that `make squeaky-clean all test` completed yesterday, but now it wont', because of this valid "invalid" example file


turbomam:
there isn't any `solution` slot to be required any more

turbomam:
I have run make squeaky-clean all test with this file in the invalid directory a few times and sometimes it works and sometimes it doesn't!?

could it be related to the versions in my poetry environment?

turbomam:
```shell
linkml-validate --schema nmdc_schema/nmdc_materialized_patterns.yaml --target-class DissolvingProcess src/data/invalid/DissolvingProcess-minimal-no-solvent.yaml 
```

> [ERROR] [src/data/invalid/DissolvingProcess-minimal-no-solvent.yaml/0] None is not of type 'array' in /substances_used

turbomam:
image

turbomam:
If we want to show that lacking any `substances_used` is invalid, then we should remove the key and rename the file

bmeluch:
what probably happened is that this gave me a single error when checking with `linkml-validate` and I said, great, put it in /invalid/ and forgot to update the filename to reflect what it's actually testing. `substances_used` does not appear to be required according to the documentation. So what do we care about testing for `DissolvingProcess`? Do we need this file anymore?

turbomam:
Yes, I think upgrading from linkml 1.8.1 to 1.8.4 makes the example file with the empty key illegal

bmeluch:
In the interest of keeping example files out of `data/problem/` if possible, I edited the file to have an invalid value for `substances_used` and updated the filename.  So it should now be testing something we care about (?), and fail with one error. Let me know if you want it changed!

turbomam:
> I edited the file to have an invalid value for `substances_used` and updated the filename. So it should now be testing something we care about (?), and fail with one error. Let me know if you want it changed!

This works with the latest linkml (1.8.4), so I think we should merge it.

I had built up my own pet vocabulary of valid and invalid placeholder strings for the example files. At some point, we could  talk about whether we want to form a consensus on what those values should be.

bmeluch (approved):

PR #2213, opened 2024-10-16
2212 replace usages of api berkeley in makefiles with production api

turbomam:
catch up with otherwise completed Berkeley schema integration


turbomam:
oops, committed directly into `main`

Issue #766, opened 2023-02-08
We have no modeling to account for technical replicates


turbomam:
This will probably require a companion issue in a submission portal repo (sheets_and_freinds, nmdc-server)

Issues: 
- how would the technical replicates be entered in the Submission Portal?
- how would Biosamples whose TechnicalReps have property of interest be found in the Data Portal?

options:
- a designator in Biosample that indicates that the Biosample is a "technical replicate" and a relationship between real Biosamples and TechnicalReplicates.
- a remodeling of Quantity values into lists of dicts, with subslots for values, identifiers for the technical replicates, and comments

anastasiyaprymolenna:
technical replicates could be marked by the biosample ID being used as another slot in a new sample entry to indicate a technical replicate of the specific biosample without repeating the metadata associated with the biosample ID

aclum:
@turbomam Would you comment on how long this would take to implement? The bioscales team is hoping to publish in a week or two and we'd like to have the soil measurements, for which there are technical replicates for a subset of the samples available for download via the API.

turbomam:
I'm revisiting old issues. Is this still important? I'd like to talk at least once more before giving an estimate.

aclum:
Storing replicates for the bioscales biogeochemcial measurements for bioscales is still the only use case, given that this is a pretty large schema change there are other items I'd advocate for working on before this like sorting out measurement units, enumerations for instrument names, syncing an updated list of GOLD pathways but we should discuss as a larger group to see if others feel differently.

ssarrafan:
Old issue. Doesn't appear to be high priority. Closing this one. @aclum @turbomam

PR #2230, opened 2024-10-23
Add `Manifest` class

kheal:
This PR

1) Adds a `Manifest` class (inherits from `Information` class) to model collections of `DataObjects` records for workflows. `Manifest` class has slot `manifest_category` with range of `ManifestCategoryEnum`
2) Adds `in_manifest` slot to `DataObject` to connect `DataObject`s to a `Manifest` or to multiple `Manifests`s
3) Adds `manifest_set` to `Database` class
4) Expands existing examples the use of the `in_manifest` slot, `Manifest` class, and `manifest_set` slot

No migration necessary, this only adds slots and classes and does not refactor existing classes or slots.

Addresses https://github.com/microbiomedata/nmdc-schema/issues/766; should unblock https://github.com/microbiomedata/nmdc-server/issues/1365

### DIFF
--- a/src/data/valid/Database-mass_spectrometry_gc.yaml
+++ b/src/data/valid/Database-mass_spectrometry_gc.yaml
@@ -25,8 +25,6 @@ data_generation_set:
     analyte_category: metabolome
     associated_studies:
       - nmdc:sty-00-555xxx
-    in_manifest:
-      - nmdc:manif-99-krhtest                                   #defined below
 manifest_set:
   - id: nmdc:manif-99-krhtest
     type: nmdc:Manifest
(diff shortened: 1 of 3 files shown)


kheal:
@aclum, @SamuelPurvine, @corilo, @turbomam, @pdpiehowski  - I could benefit from an early review and/or input on the `CollectionTypeEnum` values.  I will plan to present at metadata meeting on 10/30.

SamuelPurvine (approved):

aclum:
the current values for ManifestCategoryEnum make sense to me. Tagging @mbthornton-lbl, if implemented nmdc_automation would have to check manifest_set to see if data objects should be combined together.

turbomam (commented):
Thanks, this is good. Let's see what we can do together today in terms of refining the descriptions. I see you already have one comment about that. I'm going to suggest some more edits that emphasize minimal, realism-based descriptions.  

I especially want to add comments about potential vulnerabilities in this modelling, which I think is mainly in the area of not enforcing some coupling with other schema elements, like any fractions manifest should correspond to data objects that were the output of one single fractionation process.

kheal:
@turbomam - Thanks for the assist with the descriptions and suggested comments, I've incorporated them all now and I think this is ready for merging!

PR #2239, opened 2024-10-31
OntologyClass metadata additions: alternative_names, alternative_identifiers, relations

sierra-moxon:
Adds metadata to `OntologyClass` so that its collection in Mongo and its table in Postgres can hold cross-references (curie form), synonyms, and term descriptions available from external ontologies.  It also adds a new class, `OntologyRelation,` to capture NMDC-search/display-relevant axioms/relationships (in particular, `subclassOf` and `part_of` relations). 

In particular,r in the example data, @turbomam @cmungall - I'd like some feedback on the difference between "is_a" and "BFO:0000050" as the value for "predicate" in an OntologyRelation (using "is_a" follows the obographs convention, but I picked `rdfs:subclassOf` as the predicate for "is_a" because it sort of enables consistent processing of this field by downstream software?). 

These changes support [generic ontology metadata loading into NMDC](https://github.com/sierra-moxon/ontology-loader).

Some related tickets:
https://github.com/microbiomedata/nmdc-server/issues/1388 - support more facets for GO in search
https://github.com/microbiomedata/pilot/issues/17 - support metadata search from ontologies
https://github.com/microbiomedata/pilot/issues/12 - support ENVO search facets
https://github.com/microbiomedata/nmdc-field-notes/issues/187 - field notes ENV triad term selection

We also want to provide users with more details about our existing terms.  For example, we could extend the widget on the data portal search to include ontology metadata like cross-references, definitions, etc, to enable more informed selections.  

Screen Shot 2024-10-31 at 12 12 00 PM

Storing these metadata in Mongo and, subsequently, in Postgres might make this data more accessible for downstream tools.


src/schema/basic_classes.yaml
```diff
@@ -8,11 +8,22 @@ license: https://creativecommons.org/publicdomain/zero/1.0/
 prefixes:
   linkml: https://w3id.org/linkml/
   EFO: http://www.ebi.ac.uk/efo/
+  BFO: http://purl.obolibrary.org/obo/BFO_
+  rdfs: http://www.w3.org/2000/01/rdf-schema#
+  biolink: https://w3id.org/biolink/vocab/
+  owl: http://www.w3.org/2002/07/owl#
+
+default_curi_maps:
```
turbomam, line 16: I stopped using default curie maps under @cmungall 's guidance a long time ago. I'd like the three of us to discuss this before merging, but that would push it back to the first week of December.

src/schema/basic_classes.yaml
```diff
@@ -25,14 +36,70 @@ classes:
       - description
       - alternative_identifiers
       - type
+
   OntologyClass:
+    aliases: ["OntologyClass"]
```
turbomam, line 41: If I ask a blunt question it doesn't mean I think your idea was crazy. Just want to succinctly start some conversations.

- Why are you asserting the class name an alias?
- why are you using [short list] notation? I don't think we have any precedent for that in nmdc-schema

src/schema/basic_classes.yaml
```diff
@@ -25,14 +36,70 @@ classes:
...
     class_uri: nmdc:OntologyClass
+    description: >-
+      This class is used to represent ontology terms.
```
turbomam, line 44: I would like to stay away form self-referential definitions. Would something like "An optimized, local representation of class defined in an external ontology" be OK?

src/schema/basic_classes.yaml
```diff
@@ -25,14 +36,70 @@ classes:
...
       id:
         pattern: '^[a-zA-Z0-9][a-zA-Z0-9_\.]+:[a-zA-Z0-9_][a-zA-Z0-9_\-\/\.,]*$'
+        notes:
+          - The identifiers for terms from external ontologies can't have their ids constrained to the nmdc namespace
+    exact_mappings:
+      - biolink:OntologyClass
+      - owl:Class
+      - schema:Class
+
+  OntologyRelation:
```
turbomam, line 68: this whole thing is awesome

src/schema/basic_slots.yaml
```diff
@@ -550,7 +550,7 @@ slots:
     multivalued: true
     description: >-
       A list of alternative identifiers for the entity.
-    pattern: '^[a-zA-Z0-9][a-zA-Z0-9_\.]+:[a-zA-Z0-9_][a-zA-Z0-9_\-\/\.,]*$'
+    pattern: '^[a-zA-Z0-9][a-zA-Z0-9_\.]+:[a-zA-Z0-9_][a-zA-Z0-9_\-\/\.,\(\)\=\#]*$'
```
turbomam, line 553: what are some examples of alternative identifiers using those additional characters?

turbomam (commented):
thanks, this very good but I would like to talk though some subtleties. apologies for terms comments.

src/schema/basic_classes.yaml
sierra-moxon: @turbomam - I toyed with the idea of changing the name of the Class ("Class" is redundant and inconsistent in the context of this schema, but after some research -- I checked into Biolink, and other LinkML schemas, and I see that its a common name for this kind of class).  I will remove the alias for now. 

short list notation is a convention I use in other models, happy to remove the list.

turbomam (approved):
big step forward for NMDC and maybe for LinkmML/RDF/MongoDB in general?

Issue #220, opened 2021-12-16
Run test-jsonschema_invalid as part of test suite


cmungall:
To ensure the schema is robust, we need to check both

 1. Valid schemas pass validation
 2. Invalid schemas fail validation

Currently we only run checks for 1 (`make test-jsonschema`) as part of github actions

We also need to check for 2. We previously had some code in the makefile for this, and a set of deliberately invalid schemas (e.g. mis-spelled keys), but this seems to have dropped out of tests

cmungall:
@sujaypatil96 will handle this (I can't assign yet, but have sent invite)

sujaypatil96:
@cmungall: As part of this issue do we need to come up with more invalid schemas? Or does it entail just updating the [check-pull-request.yaml](https://github.com/microbiomedata/nmdc-schema/blob/main/.github/workflows/check-pull-request.yaml) to include `make test-jsonschema_invalid` as part of one of the steps?

We should also think about separating the schemas in [`data`](https://github.com/microbiomedata/nmdc-schema/tree/main/test/data) to `valid` and `invalid` sub folders.

cmungall:
A general SOP here is to make a series of 'mutant' files each exhibiting a
different phenotype

 - undeclared slot
 - missing field
 - multi/single value mixup
 - invalid range

you don't need to understand the schema. just take the existing positive
example, and make deliberate errors. E.g. add a field "foo": "bar" to
biosample

wdduncan:
@sujaypatil96 This may be related to #194   
I have considered using unit test for validation rather than executing `make test-jsonschema`. Using unit tests may work better for negative tests. 
Or it may be better to use the [test-jsonschema_invalid](https://github.com/microbiomedata/nmdc-schema/blob/8a528782313d667ff2261bcd61197478a6bdab40/Makefile#L210) target in the makefile.  

However, these tests were never implemented in the [make test](https://github.com/microbiomedata/nmdc-schema/blob/8a528782313d667ff2261bcd61197478a6bdab40/Makefile#L34) target. It may be easiest to get this working.

Note: there is already a variable setup for an invalid tests: `SCHEMA_TEST_EXAMPLES_INVALID`

And some invalid files in the `test/data` directory:  
- `invalid_study_test.json`
- `functional_annotation_set_invalid.json`

PR #2245, opened 2024-11-01
add 'Annotation' grouping class, abstract, to remove the need for any…

sierra-moxon:
…_of constraints


sierra-moxon:
@kheal pointed out that this field is actually slotted for deprecation and we do not need to handle it at this time.  https://github.com/microbiomedata/nmdc-schema/pull/220

closing.

Issue #2238, opened 2024-10-31
make slot usage range match structured syntax patterns 


aclum:
The current range constraints where the range is a class are frequently looser than the constraints on identifier mentions via structured syntax patterns.  This is causing confusion as people read the linkml documentation without looking at the linkml source and makes the inter-class diagrams refscan is making impractical to use.

Example: Range for has_input on WorkflowExecution has a range of NamedThing, this can be tightened to DataObject which what is specified in the structured_pattern syntax 

'{id_nmdc_prefix}:(dobj)-{id_shoulder}-{id_blade}$' 

Please go through the ranges in the schema to make sure the level of granularity is the same, making grouping classes as appropriate.

ssarrafan:
@sierra-moxon @turbomam any update on this one?  I see no updates or comments for 2 weeks so I will remove from the sprint and add the backlog label unless you're working on it?

FYI @aclum @eecavanna

sierra-moxon:
I am just seeing this; apologies.  I will take it on, but I have not started it during this sprint - it should go to the next.

turbomam:
Agree with @sierra-moxon . Important issue, but needs to move to next sprint... is that this sprint now?

Requires automation, and seperately a long term plan to avoid saying the same thing two different ways. To be implemented in LinkML.

ssarrafan:
> Agree with @sierra-moxon . Important issue, but needs to move to next sprint... is that this sprint now?
> 
> Requires automation, and seperately a long term plan to avoid saying the same thing two different ways. To be implemented in LinkML.

Yes @turbomam the new sprint started yesterday. This is a short sprint due to the Thanksgiving holiday.

ssarrafan:
Appears to have related PRs. Moving to next sprint.  
@sierra-moxon let me know if this is not active.

sierra-moxon:
needs to move to the next sprint.

ssarrafan:
Moving one more sprint?  @sierra-moxon let me know if you're not actively working on this.

ssarrafan:
I'm going to remove this from the sprint board and add the backlog label. Doesn't look active anymore.  FYI @sierra-moxon @aclum

eecavanna:
> [...] makes the inter-class diagrams refscan is making impractical to use.

[...]

PR #2246, opened 2024-11-01
reduce dependence on any_of in range constraints - all planned proces…

sierra-moxon:
…ses can be used in a 'was_generated_by' biosample slot

### DIFF
--- a/src/schema/basic_classes.yaml
+++ b/src/schema/basic_classes.yaml
@@ -444,12 +444,9 @@ slots:
     range: PersonValue
     description: Principal Investigator who led the study and/or generated the dataset.
   was_generated_by:
-    range: WorkflowExecution
+    range: PlannedProcess
     mappings:
       - prov:wasGeneratedBy
-    any_of:
-      - range: WorkflowExecution
-      - range: DataGeneration
   associated_dois:
     description: A list of DOIs associated with a resource, such as a list of DOIS associated with a Study.
     aliases:


aclum:
Is this sufficient or do we need a grouping class?

sierra-moxon:
@aclum—I looked at the descendants of PlannedProcess and we probably should discuss on a metadata call.  From the definitions of the child classes of PlannedProcess, it seems that any of them could be the value of `was_generated_by` even if we don't specifically have data for them yet, but if we need to refine definitions to make this more restrictive, we certainly can.  For this PR I tried to keep it simple and just "go up a level" in the hierarchy for the range constraint.

sierra-moxon:
from modeling call:
- @kheal - would like to review so that we can see slot_usage is correctly documented for the usages of was_generated_by (if we were don't do this, it might be messier in viz)
-

src/schema/basic_classes.yaml
```diff
@@ -444,12 +444,9 @@ slots:
     range: PersonValue
     description: Principal Investigator who led the study and/or generated the dataset.
   was_generated_by:
-    range: WorkflowExecution
+    range: PlannedProcess
```
sierra-moxon, line 447: from DM call:
the grouping here is that these two create a "dataobject", e.g. a "DataOutputter" class.

kheal:
@sierra-moxon  - I think we should put a pin in trying to resolve the `any_of`s related to `WorkflowExecution` until https://github.com/microbiomedata/nmdc-schema/issues/2238 is addressed.  I suspect that when we add `ranges` to the `slot_usage`s, it will create several more `any_of` instances regarding the `WorkflowExecution` classes.

sierra-moxon:
Can we document the new ones we need to add?  It would be my preference if we could avoid adding to the any_of paradigm.

src/schema/basic_classes.yaml
turbomam: DataEmittingProcess?

kheal:
@sierra-moxon See #2392 and #2393 for exact grouping class(es) needed for the `was_generated_by` slot.

PR #2249, opened 2024-11-07
Remodeling of range for analyte_cateogry

### LINKED ISSUE #2154: `berkeley` breakdown AnalyteCategoryEnum into types allowable for NucleotideSequencing and MassSpectrometry
Not required before the rollout but at some point we should break down AnalyteCategoryEnum into enumerations specific to subcategories. At present it would be valid to have  analyte_category populated with nom for class NucleotideSequencing

cc @mslarae13

samobermiller:
This PR replaces AnalyteCategoryEnum with NucelotideSequencingEnum and MassSpectrometryEnum and adds titles to the corresponding permissible values
fixes #2154


kheal:
@aclum and @turbomam FYI, neither @samobermiller  or I could actually get the inheritance on the "sub Enums" to result in properly validating or invalidating examples (see this example src/data/invalid/MassSpectrometry-working.yaml in this branch validates but it should not). 

We tried a few different things to no avail. We only opened this PR to see if the documentation would render appropriately (it doesn't).  Just to give you a heads up about this branch and PR.  I don't think @samobermiller  or I plan to pick it back up anytime soon (we were using it as a training example for how to write and test example yaml data).

aclum:
This does not appear to be working properly based on the passing invalid data MassSpectrometry-working.yaml, it should fail b/c of an invalid analyte_category PV.

aclum:
@kheal  yes, I see that now. @sierra-moxon  or @turbomam  can you take over here?

aclum:
cc @pkalita-lbl  in case you have any linkml insights

pkalita-lbl:
The JSON Schema generator only handles explicitly defined permissible values at the moment: https://github.com/linkml/linkml/blob/5cc5a16e345d3ee5233a193eba4909a8999b43db/linkml/generators/jsonschemagen.py#L426-L457

So, in particular, it doesn't have any logic yet to deal with `inherits` or `minus` as used in this changes.

aclum:
I will wait until the 12/16 release to merge this in  but @turbomam  @sierra-moxon @kheal @mslarae13  I need at least one reviewer on this

src/schema/basic_slots.yaml
```diff
@@ -413,7 +413,7 @@ slots:
     description: >
       The type of analyte(s) that were measured in the data generation process and analyzed
         in the Workflow Chain
-    range: AnalyteCategoryEnum
+
```
sierra-moxon, line 416: 👍 - so for now, we fall back to "string" for the range of `analyte_category` if not otherwise defined in `slot_usage`.

sierra-moxon (approved):

kheal (approved):

PR #2262, opened 2024-11-12
Migrations: Implement adapter method that copies values between two fields in all documents in a collection

### LINKED ISSUE #2259: Migrations: Implement adapter method `copy_value_from_field_to_field_in_each_document`
Adapter method signature:

```py
def copy_value_from_field_to_field_in_each_document(
    self,
    collection_name: str,
    source_field_name: str,
    destination_field_name: str,
) -> None:
    r"""
    For each document in the collection that has the source field,
    copy the value of the source field into the destination field,
    creating the destination field if it doesn't already exist.
    If the document lacks the source field, do nothing.
    """
    pass
```

Mongo query:

```js
db.getCollection("things").updateMany({
    {"foo": {$exists: true}},   // for each document that has a `foo` field,
    [{"$set": {"bar": "$foo"}}] // populate its `bar` field with the value in that `foo` field
});
```

eecavanna:
In this branch, I implemented an adapter method named `copy_value_from_field_to_field_in_each_document`, which migrator authors can use to copy all values in one field of a document into another field of that same document (creating the latter field if it doesn't already exist), for every document in a given collection.

Previously, migrator authors could do this by iterating over documents in Python. This new adapter method delegates the iteration to MongoDB.

Fixes #2259

### DIFF
--- a/nmdc_schema/migrators/adapters/adapter_base.py
+++ b/nmdc_schema/migrators/adapters/adapter_base.py
@@ -126,3 +126,16 @@ class AdapterBase(ABC):
         to facilitate iterating over all documents in a collection without actually modifying them.
         """
         pass
+
+    @abstractmethod
+    def copy_value_from_field_to_field_in_each_document(
+        self,
+        collection_name: str,
+        source_field_name: str,
+        destination_field_name: str,
+    ) -> None:
+        r"""
+        For each document in the collection that has the source field, copy the value of the source field
+        into the destination field, creating the destination field if it doesn't already exist.
+        """
+        pass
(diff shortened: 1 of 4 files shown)


kheal:
@eecavanna Did you want to add implementing this method in to https://github.com/microbiomedata/nmdc-schema/blob/main/nmdc_schema/migrators/partials/migrator_from_11_0_3_to_11_1_0/migrator_from_11_0_3_to_11_1_0_part_2.py ?  If so, I think doing that in this PR makes sense.

kheal (approved):
This all looks good from my perspective, though I am not able to give the most thorough of review since these methods are a bit new to me. All the documentation and doctests make sense to me and match the descriptions.

eecavanna:
Thanks for reviewing this! That level of review is helpful for me (e.g. in case I made a mistake somewhere while implementing this in haste). Rather than merge this in, I'd prefer to merge the third PR I opened this afternoon, since it includes these changes. That PR is: #2265

eecavanna:
On second thought, I implemented that third PR in a way where this can be merged in independently. I'll go ahead and merge this in.

PR #2263, opened 2024-11-12
Migrations: Implement adapter method that deletes field from all documents in collection

### LINKED ISSUE #2261: Migrations: Implement adapter method `delete_field_from_each_document`
Adapter method signature:

```py
def delete_field_from_each_document(
    self,
    collection_name: str,
    field_name: str,
) -> None:
    r"""
    Deletes the specified field from each document in the collection.
    """
    pass
```

Mongo query ([docs](https://www.mongodb.com/docs/manual/reference/operator/update/unset/)):

```js
db.collection.updateMany({}, {$unset: {foo: 1}}); // delete `foo` field from each document
```

eecavanna:
In this branch, I implemented an adapter method named `remove_field_from_each_document `, which migrator authors can use to delete a given field from every document in a given collection.

Previously, migrator authors could do this by iterating over documents in Python. This new adapter method delegates the iteration to MongoDB.

Fixes #2261

### DIFF
--- a/nmdc_schema/migrators/adapters/adapter_base.py
+++ b/nmdc_schema/migrators/adapters/adapter_base.py
@@ -118,6 +118,17 @@ class AdapterBase(ABC):
         pass
 
     @abstractmethod
+    def remove_field_from_each_document(
+        self,
+        collection_name: str,
+        field_name: str,
+    ) -> None:
+        r"""
+        Removes the specified field from each document in the collection.
+        """
+        pass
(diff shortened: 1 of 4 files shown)


eecavanna:
Merging without additional review (with respect to the self-review I did just now) so that the reviewers can focus on the third PR (the one that impacts the outside world): https://github.com/microbiomedata/nmdc-schema/pull/2265

PR #2269, opened 2024-11-15
Fix invalid or confusing examples

pkalita-lbl:
Related to https://github.com/microbiomedata/submission-schema/issues/227.

These changes update the `examples` attribute of slots where we don't make any further modifications when building `submission-schema`. So I think it makes the most sense to make the changes here and inherit the updates downstream.

The one exception is that I believe we decided to leave the example on `host_sex` in place and update the `host_sex_enum` permissible values. I don't recall exactly where @turbomam got the new permissible values from, but that's what I had in my notes.

### DIFF
--- a/assets/yq-for-mixs_subset_modified.txt
+++ b/assets/yq-for-mixs_subset_modified.txt
@@ -183,7 +183,8 @@
 'del(.classes)'
 'del(.enums.[].name)' 
 'del(.enums.[].permissible_values.[].text)' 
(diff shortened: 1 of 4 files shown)


pkalita-lbl:
I just found out that I'm not supposed to modify `src/schema/mixs.yaml` directly since that file get generated. Putting this into draft until I can figure out how to make the changes to that file the right way.

turbomam (approved):

Issue #2274, opened 2024-11-25
in_manifest metamodelcore.py issues


aclum:
nmdc_automation had some issues with DataObject records in dev which had in_manifest populated. These records were modifed via changesheets which passed validation and also pass linkml-validate but have an error from linkml_runtime/utils/metamodelcore.py

Testing was done in dev using https://github.com/microbiomedata/issues/issues/948 (note @eecavanna  has subsequently modified this data in dev to unblock @mbthornton-lbl). 

Corresponding nmdc_automation issue is https://github.com/microbiomedata/nmdc_automation/issues/303

I will make some invalid data which *should* fail based on but does not currently.

aclum:
cc @bmeluch @samobermiller @kheal

aclum:
see example data (src/data/invalid/DataObject-in_manifest-invalid.yaml) in linked development branch

aclum:
I think this is because we are missing a pattern constraint on in_manifest on DataObject.

kheal:
@aclum  - I think you're right. I should have added that as part of #2230.

aclum:
@kheal  I've testing locally & will have a PR review request coming shortly.

aclum:
planning on merging this in after the metadata meeting today.

PR #2275, opened 2024-11-25
Add pattern constraint to `in_manifest` on `DataObject`

aclum:
This PR adds a pattern constraint to in_manifest. Without this it is possible to malform the id mention such that it causes voliations for refscan and for linkml_runtime/utils/metamodelcore.py in nmdc_automation.

Fixes https://github.com/microbiomedata/nmdc-schema/issues/2274

Related to https://github.com/microbiomedata/issues/issues/948

### DIFF
--- a/src/data/invalid/DataObject-in_manifest-invalid.yaml
+++ b/src/data/invalid/DataObject-in_manifest-invalid.yaml
@@ -1,4 +1,4 @@
-#this test should fail because the value of in_manifest should be a list, not a list of lists
+#this test should fail because the id mention for in_manifest is not formatted correctly
 id: nmdc:dobj-11-hfjbht29
 name: nmdc_dobj-11-agsd2f41_nmdc_dobj-11-8yvaz057_QC_metrics.tsv
 description: Overall statistics from MSGF+ search results filtered to ~5% FDR
--- a/src/data/invalid/DataObject-in_manifest-invalid.yaml
+++ b/src/data/invalid/DataObject-in_manifest-invalid.yaml
@@ -1,0 +1,11 @@
+#this test should fail because the value of in_manifest should be a list, not a list of lists
+id: nmdc:dobj-11-hfjbht29
+name: nmdc_dobj-11-agsd2f41_nmdc_dobj-11-8yvaz057_QC_metrics.tsv
+description: Overall statistics from MSGF+ search results filtered to ~5% FDR
+file_size_bytes: 173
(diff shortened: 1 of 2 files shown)


src/data/invalid/DataObject-in_manifest-invalid.yaml
```diff
@@ -0,0 +1,11 @@
+#this test should fail because the value of in_manifest should be a list, not a list of lists
```
eecavanna, line 1: The JSON equivalent of the `in_manifest` value below is:

```json
  "in_manifest": [
    "['nmdc:manif-11-7796sg87']"
  ]
```

It is already a list of strings (as opposed to a list of lists)—but the string, itself, does not have the structure I think schema authors expect.

aclum:
cc @mbthornton-lbl  FYI

src/schema/basic_classes.yaml
```diff
@@ -305,7 +305,11 @@ classes:
         structured_pattern:
           syntax: "^{id_nmdc_prefix}:(wfmag|wfmb|wfmgan|wfmgas|wfmsa|wfmp|wfmt|wfmtan|wfmtas|wfnom|wfrbt|wfrqc)-{id_shoulder}-{id_blade}{id_version}$|^{id_nmdc_prefix}:(omprc|dgms|dgns)-{id_shoulder}-{id_blade}$"
           interpolated: true
-
+      in_manifest:
```
kheal, line 308: Can we add the `structured_pattern` to the global definition instead?  That way the `range` and the `structured_pattern` are stated in the same spot.

kheal (approved):

bmeluch (approved):

src/schema/basic_classes.yaml
```diff
@@ -305,7 +305,7 @@ classes:
         structured_pattern:
           syntax: "^{id_nmdc_prefix}:(wfmag|wfmb|wfmgan|wfmgas|wfmsa|wfmp|wfmt|wfmtan|wfmtas|wfnom|wfrbt|wfrqc)-{id_shoulder}-{id_blade}{id_version}$|^{id_nmdc_prefix}:(omprc|dgms|dgns)-{id_shoulder}-{id_blade}$"
           interpolated: true
-
+
```
eecavanna, line 308: I think the addition of this space was unintentional. Committing this suggestion will remove it, simplifying the PR diff.

```suggestion

```

PR #2276, opened 2024-11-26
Add slot and enumeration to `MetaproteomicsAnalysis` class and implement migrator.

### LINKED ISSUE #2256: Add slot and enumeration to `MetaproteomicsAnalysis` to enable separate types of workflows
The upcoming changes associated with `MetaproteomicsAnalysis` will result in two workflows that are inherently different, one that uses a metagenome that is a result of a `DataGeneration` process and one that uses a _in-silico_ generated metagenome.

We want to increment these two types of workflows separately and for users to understand (from the `WorkflowExecution` record) which records are a result of which type of workflow.  We also want individually-incremented ids for these workflows so users (and the data portal) do not assume that one is the "most recent".  

Our solution is to add a required slot and associated enumeration to the `MetaproteomicsAnalysis` into class.  This will make it clear which of the two workflows were used and allow for incrementing "re-runs" of the each type of workflow separately.  All existing records would be migrated to use the enumeration that represents a `MetaproteomicsAnalysis` uses a metagenome that is a result of a `DataGeneration` process.

kheal:
This MR will close #2256.

It will
1) Add the **required** slot `metaproteomics_analysis_category` to the `MetaproteomicsAnalysis` class
2) Add the definition of the slot `metaproteomics_analysis_category` to the schema, with a range of `MetaproteomicsAnalysisCategoryEnum`
3) Add definition of the `MetaproteomicsAnalysisCategoryEnum` to the schema
4) Adjust valid test to include the `metaproteomics_analysis_category`
5) Add invalid test for record that is missing the `metaproteomics_analysis_category`
6) Add migrator that populates the records of all existing `MetaproteomicsAnalysis` records with `metaproteomics_analysis_category` value of `matched_metagenome`.  Note that this logic is used because all current data use a `matched_metagenome` and we're adding this enumeration in preparation for the second category of analysis.

I have tested the migrator locally with no issues.

### DIFF
--- a/nmdc_schema/migrators/migrator_from_11_1_0_to_11_2_0.py
+++ b/nmdc_schema/migrators/migrator_from_11_1_0_to_11_2_0.py
@@ -14,17 +14,17 @@ class Migrator(MigratorBase):
 
     def set_metap_analysis_category(self, workflow: dict) -> dict:
         r"""
-        If the workflow execution records is of the type "nmdc:MetaprotemicsAnalysis", 
+        If the workflow execution records is of the type "nmdc:MetaproteomicsAnalysis", 
         add field `metaproteomics_analysis_category` and assign it the value "matched_metagenome".
 
         >>> m = Migrator()
-        >>> m.set_metap_analysis_category({'id': 123, 'type': 'nmdc:MetaprotemicsAnalysis'})  # field doesn't exist yet
-        {'id': 123, 'type': 'nmdc:MetaprotemicsAnalysis', 'metaproteomics_analysis_category': 'matched_metagenome'}
+        >>> m.set_metap_analysis_category({'id': 123, 'type': 'nmdc:MetaproteomicsAnalysis'})  # field doesn't exist yet
+        {'id': 123, 'type': 'nmdc:MetaproteomicsAnalysis', 'metaproteomics_analysis_category': 'matched_metagenome'}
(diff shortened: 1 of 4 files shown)


nmdc_schema/migrators/migrator_from_11_1_0_to_11_2_0.py
eecavanna, line 1: Migrator looks great!

eecavanna (approved):
I approve of the migrator as is (looks good to me). And thank you for writing that PR description, which I found easy to read/follow.

eecavanna:
FYI: The deadline for merging this into `main` for inclusion in the [December NMDC Release](https://github.com/microbiomedata/issues/issues/947) is this Wednesday, December 4.

kheal:
I've incorporated edits from @SamuelPurvine regarding PV descriptions.

@turbomam - if you could please review this before Wednesday, December 4, it's important for the new Metaproteomics workflow to have this enumeration in place and we want to use that workflow to process new data for the renewal.

src/schema/workflow_execution_activity.yaml
```diff
@@ -553,3 +561,14 @@ slots:
     range: MetaboliteIdentification
     multivalued: true
     inlined_as_list: true
+
+enums:
+  MetaproteomicsAnalysisCategoryEnum:
+    description: The category of metaproteomics analysis being performed.
+    permissible_values:
```
turbomam, line 568: the PV description are more value-added than the `MetaproteomicsAnalysisCategoryEnum` description!

turbomam (approved):

PR #2328, opened 2025-01-28
Ensure dependencies are correctly classified

pkalita-lbl:
This is related to the error I mentioned [here](https://github.com/microbiomedata/submission-schema/pull/297#issuecomment-2619697404).

Currently the dependency specifications in this project are all out of sorts:

* Regular dependencies declared as dev dependencies
* Dev dependencies declared as regular dependencies
* Transitive dependencies used without being declared

These changes:

* Fix the classification of existing dependencies
* Add missing dependency declarations
* Add [deptry](https://deptry.com/) checking to prevent backsliding into this situation again


turbomam (approved):
thank you very much

PR #2332, opened 2025-01-28
Add slot and enumeration to `MetabolomicsAnalysis` and implements migrator

### LINKED ISSUE #2330: Add slot and enumeration to `MetabolomicsAnalysis` to enable separate types of workflows
With the addition of the lipidomics workflow (which is a type of metabolomics, and therefore does not use a new class to describe its records), we need a way to differentiate the types of workflows from the workflow records.  Following the pattern we've set on the `MetaproteomicsAnalysis` class, we propose to add a slot and associated enumeration to differentiate these types of analyses.

This will require a migrator to back populate existing records if the slot is required.

kheal:
Closes #2330 

It will
1. Add the required slot `metabolomics_analysis_category` to the `MetabolomicsAnalysis` class
2. Add the definition of the slot `metabolomics_analysis_category` to the schema, with a range of `MetabolomicsAnalysisCategoryEnum`
3. Add definition of the `MetabolomicsAnalysisCategoryEnum` to the schema
4. Adjust valid and invalid tests to include the `metabolomics_analysis_category`
5. Add invalid test for record that is missing the `metabolomics_analysis_category`
6. Add migrator that populates the records of all existing `MetabolomicsAnalysis` records with `metabolomics_analysis_category`.  Note that there are two types of values being populated, based on the prescence of the field `has_metabolite_identifications`, which indicates that the record is associated with gc_ms_metabolomics.  Otherwise the `MetabolomicsAnalysis` records are populated with lc_ms_lipidomics

I have tested this migrator locally, and it is working as expected.

### DIFF
--- a/nmdc_schema/migrators/migrator_from_11_3_0_to_11_4_0.py
+++ b/nmdc_schema/migrators/migrator_from_11_3_0_to_11_4_0.py
@@ -14,7 +14,7 @@ class Migrator(MigratorBase):
 
     def set_metab_analysis_category(self, workflow: dict) -> dict:
         r"""
-        If the workflow execution records is of the type "nmdc:MetabolomicsAnalysis" and it has a `has_metabolite_identifications` field,
+        If the workflow execution record is of the type "nmdc:MetabolomicsAnalysis" and it has a `has_metabolite_identifications` field,
         add field `metabolomics_analysis_category` and assign it the value "gc_ms_metabolomics". If the record does not 
         have a `has_metabolite_identifications` field, it is assigned the value "lc_ms_lipidomics".
 
--- a/nmdc_schema/migrators/migrator_from_11_3_0_to_11_4_0.py
+++ b/nmdc_schema/migrators/migrator_from_11_3_0_to_11_4_0.py
@@ -1,0 +1,35 @@
+from nmdc_schema.migrators.migrator_base import MigratorBase
+
+
+class Migrator(MigratorBase):
+    r"""Migrates a database between two schemas."""
+
+    _from_version = "11.3.0"
+    _to_version = "11.4.0"
+
+    def upgrade(self):
+        r"""Migrates the database from conforming to the original schema, to conforming to the new schema."""
+
+        self.adapter.process_each_document("workflow_execution_set", [self.set_metab_analysis_category])
+
+    def set_metab_analysis_category(self, workflow: dict) -> dict:
+        r"""
+        If the workflow execution records is of the type "nmdc:MetabolomicsAnalysis" and it has a `has_metabolite_identifications` field,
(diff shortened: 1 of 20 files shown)


aclum:
this is different than https://microbiomedata.github.io/nmdc-schema/EluentIntroductionCategoryEnum/?

kheal:
> this is different than https://microbiomedata.github.io/nmdc-schema/EluentIntroductionCategoryEnum/?

Good question. Yes.  We'll have LC metabolomics and LC lipidomics - both with same EluentIntroductionCategoryEnum, but analyzed differently.

aclum (approved):

nmdc_schema/migrators/migrator_from_11_3_0_to_11_4_0.py
```diff
@@ -0,0 +1,35 @@
...
+    _to_version = "11.4.0"
+
+    def upgrade(self):
+        r"""Migrates the database from conforming to the original schema, to conforming to the new schema."""
+
+        self.adapter.process_each_document("workflow_execution_set", [self.set_metab_analysis_category])
+
+    def set_metab_analysis_category(self, workflow: dict) -> dict:
+        r"""
+        If the workflow execution records is of the type "nmdc:MetabolomicsAnalysis" and it has a `has_metabolite_identifications` field,
```
eecavanna, line 17: I think "records" was a typo. I singular-ized it here.

```suggestion
        If the workflow execution record is of the type "nmdc:MetabolomicsAnalysis" and it has a `has_metabolite_identifications` field,
```

src/data/invalid/MetabolomicsAnalysis-invalid_execution_resource_not_in_enum.yaml
```diff
@@ -11,3 +11,5 @@ has_output:
   - nmdc:dobj-11-123d
   - nmdc:dobj-11-34dj4
 was_informed_by: nmdc:omprc-11-284u7d
+metabolomics_analysis_category: gc_ms_metabolomics
+
```
eecavanna, line 15: In this suggestion, I have removed the empty line (as I don't see one in the other YAML files in this PR).

```suggestion
```

eecavanna (approved):
LGTM! Thanks for including the thorough docstring (including doctests) in the migrator!

turbomam:
Thanks. I am going to approve this, but @aclum's analysis

> this is different than https://microbiomedata.github.io/nmdc-schema/EluentIntroductionCategoryEnum/?

is important. We should always be on the lookout for places where the schema lets us say to different things that might be redundant with, or incompatible with one another.

turbomam (approved):

turbomam:
I'm really impressed with how the migrators have been keeping up with the schema changes

sierra-moxon (approved):

PR #2354, opened 2025-03-04
mark sample_state_information to be deprecated

mslarae13:
discussions in the mass spec material processing squad 
this slot shouldn't be on `PortionOfSubstance` and there's no indication that it's relevant to NMDC


src/schema/core.yaml
```diff
@@ -1813,6 +1813,8 @@ slots:
     description: >-
       The chemical phase of a pure sample, or the state of a mixed sample
     range: SampleStateEnum
+    deprecated: true
+    last_updated_on: "2025-03-04"
```
turbomam, line 1817: is a migrator required because "the schema would have allowed this"

turbomam (commented):
otherwise a good PR

kheal:
@mslarae13  - you need to remove the slot from it's class as well, not just deprecate the slot.

src/schema/core.yaml
eecavanna: Thanks for asking!

The addition of these two attributes—by itself—doesn't (to me) warrant creating a migrator.

Once this slot gets removed from the classes that "have" it, I will want to create a migrator that can be used to update database documents accordingly. So, I think this PR will eventually warrant creating a migrator—but it does not yet (to me), in its current state (where these two lines are the only lines that have been changed).

src/data/valid/PortionOfSubstance-1.yaml
```diff
@@ -3,7 +3,6 @@
 #  type: nmdc:QuantityValue
 #  has_unit: percent
 #  has_numeric_value: 10
-#sample_state_information: liquid
```
mslarae13, line 6: @turbomam what's up with this example? Was it just never finished?

eecavanna:
Now that (i.e. as of commit https://github.com/microbiomedata/nmdc-schema/pull/2354/commits/2c9c250b4129dfd186d8b4891be83510f6427de7 on this branch) instances of the `PortionOfSubstance` schema class can no longer have a field named `sample_state_information`, creating an accompanying migrator will be necessary (in order to "keep data valid across schema versions").

In the [schema docs](https://microbiomedata.github.io/nmdc-schema/PortionOfSubstance/#usages), I can see that that the `PortionOfSubstance` class is in the range of the `substances_used` slot of the following schema classes:
- `Extraction`
- `StorageProcess`
- `DissolvingProcess`
- `ChemicalConversionProcess`
- `MobilePhaseSegment`

The migrator will locate each "instance" of each of those classes in the database, check whether it has a `substanced_used` field and—if it does—delete the `sample_state_information` field from each object in that `substances_used` field.

kheal:
@eecavanna  note that there will not be instances of `MobilePhaseSegment`, that class is also in-lined so you'll need to go one class up for that class.

eecavanna:
Thanks, @kheal! In search of instances of the `MobilePhaseSegment` class, I'm planning to have the migrator look at the `ordered_mobile_phases` fields of instances of the [ChromatographyConfiguration](https://microbiomedata.github.io/nmdc-schema/ChromatographyConfiguration/) class (all of which are represented by documents in the `configuration_set` collection) and the [ChromatographicSeparationProcess](https://microbiomedata.github.io/nmdc-schema/ChromatographicSeparationProcess/) class (all of which are represented by documents in the `material_processing_set` collection).

eecavanna:
Regarding the migrator, it _might_ be written already (it is on this branch). This is the most complex schema change that I have come up with the migrator requirements (mostly) myself for, and I have not yet validated those requirements with someone more familiar with the schema. That is the main thing I want to do before calling the migrator "done."

Given my current workload and the fact that the schema freeze date is today, I plan to not finish this today and, instead, finish it before the _next_ release (i.e. the April release).

mslarae13:
> add deprecated slot to deprecated YAML file

@turbomam I thought you didn't move it to the deprecated.yaml until after 1 releas?
@sierra-moxon can you confirm or deny?

eecavanna:
My understanding of the standard, 2-step deprecation process is:

When someone decides to deprecate a slot...
1. At schema release N:
   1. Add the "deprecated" attribute to the slot (to inform people that it's going away soon).
   2. Keep the slot on the class so it remains valid (to give people a chance to transition away from it).
2. At schema release N+1:
   1. Remove the slot from the class (so its usage is invalid).
   2. Move the slot to the "deprecated" (`*`) section of the schema (so people know it used to exist).

- `*` I think calling this section the "deprecated" section is an understatement. The things in that section have been deprecated in the past, but, by the time they are added to that section, they are "retired". I'd recommend calling it "retired". I think the name of this section is tangential to @turbomam's feedback above, though.

I think @turbomam meant that he's OK with condensing the "release N and release N+1" actions into a single release—given that we haven't seen people use this slot—but not with step "2.i" being done without also doing step "2.ii". I have not confirmed with @turbomam that that is what he meant, though.

nmdc_schema/migrators/partials/migrator_from_11_5_1_to_11_6_0/migrator_from_11_5_1_to_11_6_0_part_2.py
eecavanna, line 1: This migrator is not being introduced via this PR. It is only being _moved_.

This migrator was introduced via PR https://github.com/microbiomedata/nmdc-schema/pull/2413/. There, its location was `nmdc_schema/migrators/migrator_from_11_5_1_to_11_6_0.py`.

In this PR, it is being moved (and its `_from_version`/`_to_version` changed) to be a "partial" migrator, since there are now multiple migrators involved between the two schema versions.

nmdc_schema/migrators/partials/migrator_from_11_5_1_to_11_6_0/migrator_from_11_5_1_to_11_6_0_part_1.py
eecavanna, line 1: This is the only migrator being introduced to accompany this schema change.

nmdc_schema/migrators/partials/migrator_from_11_5_1_to_11_6_0/migrator_from_11_5_1_to_11_6_0_part_1.py
```diff
@@ -0,0 +1,251 @@
...
+                  they are represented by objects having a type value of `nmdc:MobilePhaseSegment`
+                  in the multivalued `ordered_mobile_phases` fields of instances of the following schema classes:
+                  - [x] A. `ChromatographyConfiguration`
+                    - Docs: https://microbiomedata.github.io/nmdc-schema/ChromatographyConfiguration/
+                    - All of these are documents in the `configuration_set` collection,
+                      having a type value of `nmdc:ChromatographyConfiguration`.
+                  - [x] B. `ChromatographicSeparationProcess`
+                    - Docs: https://microbiomedata.github.io/nmdc-schema/ChromatographicSeparationProcess/
+                    - All of these are documents in the `material_processing_set` collection,
+                      having a type value of `nmdc:ChromatographicSeparationProcess`.
```
eecavanna, line 48: Hi @kheal, this `Note:` comment block contains my "thoughts" as I attempted to convert the schema change into migrator requirements. If you have a chance, will you read this and let me know whether anything in it stands out to you as problematic? I can also walk through it with you in real time (e.g. via Zoom).

kheal (approved):
@eecavanna - I reviewed the migrator.  The logic, methods, and tests all look great to me. Thank you for the thorough documentation.

mslarae13:
@turbomam @eecavanna @sierra-moxon 
Anyone know why these checks are failing now that I've moved the slot to the deprecated.yaml?

eecavanna:
@mslarae13 the check failed because the computer that runs the check failed to install some Python packages. I think it's a transient failure unrelated to the changes on this branch. I've launched a re-run of the check.

turbomam (approved):

PR #2356, opened 2025-03-04
Draft: add mixing slot & range of MixingProcess

### LINKED ISSUE #2355: add mixing and allow MixingProcess to be inlined
MixingProcess currently can't be inlined.

I need to capture mixing as part of an extraction.
Can we allow MixingProcess to be a collection or an inlined range?

mslarae13:
Closes #2355

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -391,7 +391,7 @@ classes:
       - extraction_targets
       - input_mass
       - volume
-      - mixing
+      - mixing_steps
     slot_usage:
       has_input:
         required: true
@@ -1355,10 +1355,10 @@ slots:
     multivalued: true
     description: The portion of the sample that is taken for downstream activity.
 
-  mixing:
+  mixing_steps:
     range: MixingProcess
     inlined_as_list: true
-    inlined: true
-    description: Describes the mixing of a sample during a process.
+    description: Describes how mixing of a sample was done during a process.
     list_elements_ordered: true
+    multivalued: true
 
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -391,6 +391,7 @@ classes:
       - extraction_targets
       - input_mass
       - volume
+      - mixing
     slot_usage:
       has_input:
         required: true
@@ -1344,6 +1345,7 @@ slots:
   stationary_phase:
     range: StationaryPhaseEnum
     description: The material the stationary phase is comprised of used in chromatography.
+  
   chromatographic_category:
     range: ChromatographicCategoryEnum
     description: The type of chromatography used in a process.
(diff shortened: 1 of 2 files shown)


kheal (changes requested):
Would you please add an example of using this slot to one of the existing examples so we have it covered for CI/CD (maybe src/data/valid/Extraction-metabolomics.yaml)?

src/schema/nmdc.yaml
```diff
@@ -1352,3 +1354,11 @@ slots:
     range: SamplePortionEnum
     multivalued: true
     description: The portion of the sample that is taken for downstream activity.
+
+  mixing:
```
turbomam, line 1358: mixing_steps?

src/schema/nmdc.yaml
```diff
@@ -1352,3 +1354,11 @@ slots:
...
+    range: MixingProcess
+    inlined_as_list: true
+    inlined: true
```
turbomam, line 1361: inferred form inlined_as_list

src/schema/nmdc.yaml
```diff
@@ -1352,3 +1354,11 @@ slots:
...
+    description: Describes the mixing of a sample during a process.
+    list_elements_ordered: true
+
```
turbomam, line 1364: add multivalued

turbomam: we need to check whether all of the transformations and ETL-like processes in our ecosystem honor list_elements_ordered

turbomam (commented):
otherwise a good PR

src/schema/nmdc.yaml
mslarae13: made multivalued and updated slot

mslarae13:
@turbomam any thoughts on ID being required and inlined? Is that ok? Seems weird to have inlined and ID?

cc: @sierra-moxon

mslarae13:
note, I added probe sonication information to the description. I think we can do the same for vortexing as it's not really a piece of information I think we would ever filter on or would have affects on results. Just an FYI also captured in the protocol

kheal:
Group agrees that id and in-line is a no go.  We will try to think of a different solution for substeps of processes that do not have `has_input` `has_output` or `id`.  These would be in-lined data only.

mslarae13:
MixingProcess: upon further review, isn’t actually important for the standard reason of capture by NMDC

Steps in a process will be captured as collections and called out in metadata if
- It influence workflows & needs referenced in results or for running workflows
- If it’s a metadata field someone would need / want to filter or search by

Mixing as was modeled in this class doesn’t fit either of those, as such will not be captured

For now, convert this PR to draft, and after more MS MP metadata has been captured, we’ll return to this slot and class and determine if it’s needed, how to capture it, or if it should be deprecated.

ssarrafan:
Appears to be active. Will move to current sprint.

ssarrafan:
Is it active? @mslarae13 @kheal @turbomam 
I'm removing from this sprint.

mslarae13:
We decided NOT to do this & allow inlined

PR #2338, opened 2025-02-06
Deprecate `ProtocolExecution` class and related slots

bmeluch:
This PR adds `deprecation` annotations to the definitions for class `ProtocolExecution`, enum `ProtocolCategoryEnum`, and slots `protocol_execution_set` and `has_process_parts`. It also removes `ProtocolExecution` from the example data files and `protocol_execution_set` from the Makefile.

PR #2359, opened 2025-03-05
Cleanup: Move `ProtocolExecution` and related slots to `deprecated.yaml`

bmeluch:
Follow up to https://github.com/microbiomedata/nmdc-schema/pull/2338. 
Moves class and slot definitions to `deprecated.yaml`

### DIFF
--- a/src/schema/deprecated.yaml
+++ b/src/schema/deprecated.yaml
@@ -58,6 +58,35 @@ classes:
       - type
     deprecated: "not used. 2024-11 https://github.com/microbiomedata/nmdc-schema/issues/2250"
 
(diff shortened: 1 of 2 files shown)


mslarae13:
I believe we should not merge this until 1 release cycle. So, after the 24th, is probably best?
Yes, @sierra-moxon & @turbomam 

Also, reminds me to confirm that a schema release will be done before the production release on the 24th. When will that be? So I can make sure all the PRs I have in are merged in time.

turbomam (approved):
thanks, this seems like really good adherence of the NMDC deprecation protocol

PR #2360, opened 2025-03-05
add extra metadata to ontology_class in order to capture obsolete and…

sierra-moxon:
… root status of a term.

fixes: https://github.com/microbiomedata/issues/issues/1118

### DIFF
--- a/src/schema/basic_classes.yaml
+++ b/src/schema/basic_classes.yaml
@@ -54,7 +54,7 @@ classes:
         description: >-
           A boolean value indicating whether the ontology term is obsolete.
         comments: >-
-          If true (the ontology term is delcared obsolete via the ontology source itself), 
+          If true (the ontology term is declared obsolete via the ontology source itself), 
           the term is no longer considered a valid term to use in an annotation at NMDC,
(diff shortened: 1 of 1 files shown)


src/schema/basic_classes.yaml
```diff
@@ -49,6 +49,19 @@ classes:
         multivalued: true
         inlined_as_list: true
         inlined: true
+      is_obsolete:
```
turbomam, line 52: thanks for adding these useful attributes. I would prefer that all nmdc-schema classes use slots and not attributes, but you may have a good reason for this that we just haven't discussed yet.

turbomam (approved):

src/schema/basic_classes.yaml
sierra-moxon: sure, can do. 
I think the slot name "definition", for example, is pretty broad but its description is narrow and the mismatch there led me to make it an attribute instead of a reusable slot.  In addition, I'm not sure we want folks to apply `is_obsolete` or `is_root` to other classes?  But, for consistency, I've made them into slots.

Issue #2258, opened 2024-11-12
Find replacements for GENEPIO and OMIT mappings if possible and/or define prefixes


turbomam:
Thanks for providing these mappings, @aculm. 

`GENEPIO` and  `OMIT` aren't defined as prefixes yet. That's an error condition but it isn't super visible in the standard build process. We haven't used them before (although GENEPIO appears in an un-checked example in the mixs.yaml file). We should be conservative about adding new namespaces to the schema, especially if they aren't clearly authorities on the concepts being addressed.

Examples:

```yaml
  members_id:
    range: string
    multivalued: true
    description: Names of the contigs that make up a metagenome-assembled genome.
    close_mappings:
      - GENEPIO:0100596
```

and

```yaml
  num_5s:
    description: Number of 5s seqeuences detected, a subunit of ribosomes.
    range: integer
    minimum_value: 0
    broad_mappings:
      - OMIT:0013248
```

eecavanna:
When I ran `make squeaky-clean all test` today in order to generate release artifacts for `11.1.0`, I saw this at the end of the console output:

```shell
# ...
mkdir -p examples/output
poetry run linkml-run-examples \
        --schema nmdc_schema/nmdc_materialized_patterns.yaml \
        --input-directory src/data/valid \
        --counter-example-input-directory src/data/invalid \
        --output-directory examples/output > examples/output/README.md
Unrecognized prefix: GENEPIO
Unrecognized prefix: OMIT
```

The final two lines caught my eye. I think they are what @turbomam was referring to when he said this error currently "isn't super visible in the standard build process."

Issue #2186, opened 2024-09-24
post berkeley-schema-fy24 merge issue: review `FileTypeEnum` composition and correlation with other `DataObject` slots/relationships


turbomam:
- Can a `DataObject` that is the output of any process use any `FileTypeEnum` in it's `data_object_type` slot?
- do the different permissible values come from different axes of differentiation?
- should we use an `is_a` hierarchy within the PVs?
-

turbomam:
Claude finds these different axes of differentiation or concerns in `FileTypeEnum`:

1. Data Type / Analysis Method:
   - Metagenome data
   - Metabolomics data (FT ICR-MS, GC-MS)
   - Metaproteomics data
   - Assembly data
   - Annotation data (various types)
   - Read-based analysis
   - Taxonomic classification (GOTTCHA2, Kraken2, Centrifuge)

2. Processing Stage:
   - Raw data
   - Filtered data
   - Error-corrected data
   - Assembled data
   - Annotated data

3. File Format:
   - FASTQ
   - BAM
   - FASTA
   - GFF
   - JSON
   - TSV
   - PDF
   - HTML

4. Sequencing Read Type:
   - Raw Read 1 (forward)
   - Raw Read 2 (reverse)
   - Interleaved paired-end

5. Quality Control Stage:
   - QC Statistics
   - QC non-rRNA reads

6. Biological Entity Focus:
   - Protein-related
   - Peptide-related
   - RNA-related (rRNA, tRNA, etc.)
   - Gene-related

7. Output Type:
   - Report files
   - Statistical files
   - Plot files (heatmap, barplot, Krona plot)
   - Binning results

8. Annotation Type:
   - Structural annotation
   - Functional annotation
   - Various specific annotation types (e.g., TIGRFam, CRT, Genemark, etc.)

9. Compression Status:
   - Compressed files (e.g., zip files for bins)
   - Uncompressed files

10. Workflow Stage:
    - Intermediate files
    - Final output files
    - Workflow statistics

mslarae13:
@turbomam is this accomplished with the merged PR?

turbomam:
enum | permissible_value | description | file_name_pattern_annotation | see_also | example_values | broad_mappings | exact_mappings
-- | -- | -- | -- | -- | -- | -- | --
[...]

PR #2409, opened 2025-04-09
#2258 prefix audit/normalization

turbomam:
and comment etc tidying


turbomam:
I checked in the generated files in case anybody needs to check them

turbomam:
- removes or updates old comments, notes and todos
- turns tolerated but technically incorrect scalar comments into list comments
- ensures all modules have the right prefix definitions as needed by their 
    - mappings
    - meaning
    - pattern
    - id_prefixes
    - structured_pattern
    - structured_aliases
    - see_also
    - settings
    - examples

wrapped all prefix expansions in quotes, since the quotes are required for the bioregistry prefixes
    
also checked/removed verbose forms of dictionaries and auto-generated slots
- name
- tag
- from_schema
- prefix_prefix

src/schema/core.yaml
```diff
@@ -1017,7 +1023,8 @@ classes:
 
       gold_biosample_identifiers:
         description: Unique identifier for a biosample submitted to GOLD that matches the NMDC submitted biosample
-        comments: This is the ID provided by GOLD that starts with 'GB'
+        comments:
+          - This is the ID provided by GOLD that starts with 'GB'
```
sujaypatil96, line 1027:
```suggestion
          - This is the ID provided by GOLD that starts with 'Gb'
```

src/schema/external_identifiers.yaml
```diff
@@ -10,10 +10,19 @@ imports:
   - attribute_values # in cycle with basic_classes
 
 prefixes:
+  bioproject: 'https://identifiers.org/bioproject:'
```
sujaypatil96, line 13: Should we use https://bioregistry.io/registry/bioproject instead of identifiers.org since most of our other prefixes are using bioregistry?

```suggestion
  bioproject: 'https://bioregistry.io/bioproject:'
```

Especially since the biosample prefix right below is also using bioregistry?

src/schema/portal_sample_id.yaml
```diff
@@ -29,9 +29,8 @@ slots:
     title: analysis/data type
     examples:
       - value: metagenomics; metabolomics; metaproteomics
-    from_schema: https://example.com/nmdc_dh
-    see_also:
-      - MIxS:investigation_type
+    comments:
+      - MIxS:investigation_type was included as a see_also but that term doesnt' resolve any more
```
sujaypatil96, line 33:
```suggestion
      - MIxS:investigation_type was included as a `see_also` but that term doesn't resolve any more
```

nmdc_schema/nmdc_materialized_patterns.schema.json
```diff
@@ -5801,6 +5801,7 @@
                 "instrument_used": {
                     "description": "What instrument was used during DataGeneration or MaterialProcessing.",
                     "items": {
+                        "pattern": "^^(nmdc):inst-([0-9][a-z]{0,6}[0-9])-([A-Za-z0-9]{1,})$",
```
sujaypatil96, line 5804: There's a general pattern that we're seeing with the regex patterns being resolved to double carats (`^^`) and I think that's coming up because the `id_nmdc_prefix` setting pattern (for structured patterns) is written this way: `"^(nmdc)"` (see [here](https://github.com/microbiomedata/nmdc-schema/blob/main/src/schema/nmdc.yaml#L111))

The structured pattern also uses a carat, and the setting pattern also has a carat, so hence the double carat in the resolved pattern. We should make it so that the carats are used in one place only.

Happy to have this fix on a separate PR to keep the scope of this PR isolated.

sujaypatil96 (commented):
Thanks for this PR that tidies up/does housekeeping on the schema files @turbomam ❤️ it definitely makes things cleaner and easier to read! 

I've left a few trivial comments on this PR, which we can choose to adopt/drop, but I think some of them are useful!

nmdc_schema/nmdc_materialized_patterns.schema.json
turbomam: Thanks @sujaypatil96 . I have removed the carets from the `structured_pattern.syntaxes`, so it only appears in   `settings.id_nmdc_prefix`

Whether we stick with that or ultimately decide to flip it, we should think of places to document it or test it

src/schema/basic_slots.yaml
```diff
@@ -695,16 +703,12 @@ enums:
       Metagenome HQMQ Bins Compression File:
         description: Compressed file containing high qulaity and medium quality metagenome bins and associated files
         annotations:
-          file_name_pattern:
-            tag: file_name_pattern
-            value: "[mag_wf_activity_id]_hqmq_bin.zip"
+          file_name_pattern: "[mag_wf_activity_id]_hqmq_bin.zip"
```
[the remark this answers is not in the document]
turbomam: `file_name_pattern` annotations are exclusively used in `FileTypeEnum` but as annotations, we don't have good control over where they appears or what their value looks like. They don't really look like true regular expressions to me. Maybe we should follow the `structured_pattern.syntax`/`settings` style, even if these are never used a s validation patterns

https://github.com/microbiomedata/nmdc-schema/issues/2186#issuecomment-2803555028

nmdc_schema/nmdc_materialized_patterns.schema.json
turbomam: https://github.com/microbiomedata/nmdc-schema/issues/2410

src/schema/basic_slots.yaml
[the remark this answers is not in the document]
turbomam: otherwise, I'd like to save this for another day and merge this now if that's OK with you @sujaypatil96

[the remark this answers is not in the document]
sujaypatil96: Yup, makes sense! was curious about the motivation to use `annotations` instead of validatable metaslots like `structured_patterns` / `patterns`.

sujaypatil96 (approved):
Thanks for addressing the changes @turbomam 💕 This PR is ready to be merged! 🚀

PR #2413, opened 2025-04-11
Add metatranscriptome raw read permissible values to `FileTypeEnum`

### LINKED ISSUE #2412: Add `FileTypeEnum` permissible values for metatranscriptome raw reads
Currently we have three permissible values related to metagenome raw reads:

* `Metagenome Raw Reads`
* `Metagenome Raw Read 1`
* `Metagenome Raw Read 2`

This tasks is to add the analogous permissible values for metatranscriptome raw reads.

pkalita-lbl:
Fixes #2412 

These changes add three new permissible values to `FileTypeEnum`:

* `Metatranscriptome Raw Reads`
* `Metatranscriptome Raw Read 1`
* `Metatranscriptome Raw Read 2`

Additionally a migrator is added which does the following:

* Find records in `data_generation_set` with `analyte_category = metatranscriptome`.
* For each such `nmdc:DataGeneration` instance, find the the `nmdc:DataObject` instances associated via the `has_output` relationship
* Foe each such `nmdc:DataObject`, if the `data_object_type` is `Metagenome Raw Reads`, `Metagenome Raw Read 1`, or `Metagenome Raw Read 2`, change it to the equivalent `Metatranscriptome` value.

### DIFF
--- a/nmdc_schema/migrators/migrator_from_11_5_1_to_11_6_0.py
+++ b/nmdc_schema/migrators/migrator_from_11_5_1_to_11_6_0.py
@@ -11,7 +11,7 @@ class Migrator(MigratorBase):
     def __init__(self, adapter: AdapterBase=None, logger=None):
         super().__init__(adapter, logger)
 
-        self.data_objects_to_update = set()
+        self.ids_of_data_objects_to_update = set()
         self.data_object_type_update_map = {
             "Metagenome Raw Reads": "Metatranscriptome Raw Reads",
             "Metagenome Raw Read 1": "Metatranscriptome Raw Read 1",
@@ -21,7 +21,7 @@ class Migrator(MigratorBase):
     def upgrade(self) -> None:
         r"""Migrates the database from conforming to the original schema, to conforming to the new schema."""
         self.adapter.do_for_each_document("data_generation_set", self.find_metatranscriptome_data_objects)
-        self.logger.info(f"Found {len(self.data_objects_to_update)} data objects to update.")
+        self.logger.info(f"Found {len(self.ids_of_data_objects_to_update)} data objects to update.")
 
         self.adapter.process_each_document("data_object_set",[self.update_data_object_type])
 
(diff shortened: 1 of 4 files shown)


nmdc_schema/migrators/migrator_from_11_5_1_to_11_6_0.py
```diff
@@ -0,0 +1,84 @@
...
+
+        >>> m = Migrator()
+        >>> m.find_metatranscriptome_data_objects({'id': 'nmdc:dgns-11-12345678', 'analyte_category': 'metatranscriptome', 'has_output': ["nmdc:dobj-11-abcd1234"]})
+        >>> m.data_objects_to_update
+        {'nmdc:dobj-11-abcd1234'}
+        >>> m.find_metatranscriptome_data_objects({'id': 'nmdc:dgns-11-12345678', 'analyte_category': 'metagenome', 'has_output': ["nmdc:dobj-11-abcd1234"]})
+        >>> m.data_objects_to_update
+        set()
+        """
+        self.data_objects_to_update = set()
```
eecavanna, line 42: I consider this assignment to be optional, since the variable is initialized to an empty set in the `Migrator`'s `__init__` method. If you do remove this assignment, I'd expect the final doctest in this method to fail due to the set still containing the previous element.

nmdc_schema/migrators/migrator_from_11_5_1_to_11_6_0.py
```diff
@@ -0,0 +1,84 @@
...
+class Migrator(MigratorBase):
+    r"""Migrates a database between two schemas."""
+
+    _from_version = "11.5.1"
+    _to_version = "11.6.0"
+
+    def __init__(self, adapter: AdapterBase=None, logger=None):
+        super().__init__(adapter, logger)
+
+        self.data_objects_to_update = set()
```
eecavanna, line 14: I recommend renaming this to indicate that it contains `id` values, as opposed to data objects, themselves; for example, `ids_of_data_objects_to_update` or `target_data_object_ids`.

If not renaming it, I'd recommend adding a comment saying that it contains `id` values of the data objects that will be updated.

nmdc_schema/migrators/migrator_from_11_5_1_to_11_6_0.py
```diff
@@ -0,0 +1,84 @@
...
+        set()
+        """
+        self.data_objects_to_update = set()
+        if data_generation.get("analyte_category") == "metatranscriptome":
+            for output in data_generation.get("has_output", []):
+                self.data_objects_to_update.add(output)
+
+    def update_data_object_type(self, data_object: dict) -> dict:
+        r"""
+        Update the data_object_type of data_objects that are in the data_objects_to_update set. If
```
eecavanna, line 49:
```suggestion
        Update the data_object_type of data_objects whose ids are in the data_objects_to_update set. If
```

eecavanna:
LGTM! I reviewed the migrator on Thursday afternoon, at commit https://github.com/microbiomedata/nmdc-schema/pull/2413/commits/5179cd421c85c156d1218c974b285e41a670780f, and saw no deal breakers. I found the code easy to read and understand, with no prior knowledge of the associated schema change. I left a few comments about (a) an extra assignment statement and (b) renaming a variable to be more self-documenting.

aclum:
There are some data objects aren't connected via has_output on the data_generation_set record. So we either need to add the query below to find them or I or someone else needs to do a changesheets to link the records up correctly.

```

db.getCollection('data_generation_set').aggregate(
  [
    {
      $match: {
        analyte_category: 'metatranscriptome'
      }
    },
    {
      $match: { has_output: { $exists: false } }
    },
    {
      $lookup: {
        from: 'workflow_execution_set',
        localField: 'id',
        foreignField: 'was_informed_by',
        as: 'workflow_execution_set'
      }
    },
    {
      $match: {
        workflow_execution_set: { $size: 1 }
      }
    },
    {
      $lookup: {
        from: 'data_object_set',
        localField:
          'workflow_execution_set.has_output',
        foreignField: 'id',
        as: 'data_object_set'
      }
    },
    {
      $project: {
        'data_object_set.id': 1,
        'data_object_set.data_object_type': 1
      }
    },
    {
      $match: {
        'data_object_set.data_object_type': {
          $in: [
            'Metagenome Raw Reads',
            'Metagenome Raw Read 1',
            'Metagenome Raw Read 2'
          ]
        }
      }
    }
  ],
  { maxTimeMS: 60000, allowDiskUse: true }
);
```

pkalita-lbl:
@aclum I think the latest changes should take that case into account. Can you take a look?

nmdc_schema/migrators/migrator_from_11_5_1_to_11_6_0.py
```diff
@@ -0,0 +1,131 @@
...
+        {'nmdc:dobj-11-abcd1234'}
+        >>> m = Migrator()
+        >>> m.ids_of_metatranscriptome_data_generation_with_no_output = {'nmdc:dgns-11-12345678'}
+        >>> m.scan_workflow_execution_set({'id': 'nmdc:wfnom-11-12345678', 'was_informed_by': 'nmdc:dgns-11-98765432', 'has_output': ["nmdc:dobj-11-abcd1234"]})
+        >>> m.ids_of_data_objects_to_update
+        set()
+        """
+        if "was_informed_by" in workflow_execution:
+            data_generation_id = workflow_execution["was_informed_by"]
+            if data_generation_id in self.ids_of_metatranscriptome_data_generation_with_no_output:
```
aclum, line 90: adding filter for 'type':'nmdc:MetagenomeSequencing' will eliminate having to check a bunch of data objects that don't need changing.

aclum (changes requested):
See comment about adding a type filter.

aclum (approved):

PR #2420, opened 2025-04-15
eliminates lingering `any_of`s and tests against future additions

turbomam:
this will make it easier to work with the schema and it's data in SQL and OWL/RDF

May add some redundant tests


src/data/valid/MassSpectrometry-valid-syringe-without-config.yaml
kheal, line 1: pretty redundant with existing src/data/valid/MassSpectrometry-NOM.yaml, but otherwise fine.

kheal (approved):
Looks fine to me.  I didn't test the test locally to make sure it would fail if we introduced an `any_of`, but I'm assuming that has been tested.

PR #2422, opened 2025-04-16
2414 change cardinality of was informed by to 1 to n

mbthornton-lbl:
This PR provides:
- an update to basic_classes WorkflowExecution - adding `multivalue: true` to the slot_usage for was_informed_by
- Update to valid examples to be correct
- Update invalid examples so that they are not invalid due to was_informed_by
- Added a new invalid example WorkflowExecution-was-invormed-by-not-a-list


src/data/valid/MetatranscriptomeAnnotation-1.yaml
```diff
@@ -17,7 +17,8 @@ has_output:
 img_identifiers:
   - img.taxon:x1
 name: xxx
-was_informed_by: nmdc:omprc-11-djb2d2
+was_informed_by:
+    - nmdc:omprc-11-djb2d2
```
eecavanna, line 21: I noticed the number of spaces (4) used for indentation on the added line differs from the number (2) used for indentation elsewhere in the file. I recommend using the same number (2) as is already in use in the file.

eecavanna:
I don't understand why these files were deleted from the repo:

image

eecavanna:
I recommend including at least one example data file in which there is more than one value in the `was_informed_by` list.

eecavanna:
If I am understanding this change correctly (that the `was_informed_by` field must now contain a list), a migrator will be required in order to update the existing `was_informed_by` values in the database.

aclum:
@eecavanna  those are derived files and shouldn't be included or else we end up with a bunch of merge conflicts.

eecavanna:
Hi @aclum, it makes sense to me that they not be updated in this PR. The thing I'm confused about is why they are being _deleted from the repo_ (as opposed to _doing nothing_) in this PR.

aclum:
I think the were accidentally added either here or with a previous PR

mbthornton-lbl:
@eecavanna I am not sure how those files got deleted either :-/.  I merged main back into this branch to restore them.

kheal (approved):
Looks great, thanks for taking the time to update the invalid examples.  Maybe update the title of the PR so it will make a nice bullet in the release notes "Make `was_informed_by` multivalued".

I don't understand why the nmdc_schema/gold-to-mixs.sssom.tsv file was delted though, was that on purpose?

aclum:
gold-to-mixs.sssom.tsv is a derived schema file from the make process so doesn't need to be included @turbomam to confirm.

PR #2426, opened 2025-04-16
Move development-only dependencies back to dev group

### LINKED ISSUE #2425: remove recent bloat from dependencies, like `pandas`

pkalita-lbl:
Fixes #2425 

These changes move a handful of dependencies that were added to the main dependency group to the dev group. These dependencies are used in development-only code.

These changes also remove some `deptry` configuration which is unnecessary when running it only against the `nmdc_schema` package. This is consistent with the intended use of `deptry` to only scan files which are included in the final published package.

### DIFF
--- a/.github/workflows/test_pages_build.yaml
+++ b/.github/workflows/test_pages_build.yaml
@@ -18,15 +18,13 @@ jobs:
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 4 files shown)


turbomam (commented):
thanks

PR #2428, opened 2025-04-17
Ensure all CURIE prefixes make it into generated Python

pkalita-lbl:
Currently the `nmdc` module cannot be imported because of a missing prefix:

```
>>> from nmdc_schema.nmdc import Database
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/PAKalita/Work/tmp/.venv/lib/python3.12/site-packages/nmdc_schema/nmdc.py", line 7772, in <module>
    class DoiProviderEnum(EnumDefinitionImpl):
  File "/Users/PAKalita/Work/tmp/.venv/lib/python3.12/site-packages/nmdc_schema/nmdc.py", line 7776, in DoiProviderEnum
    meaning=ROR["04rc0xn13"])
            ^^^
NameError: name 'ROR' is not defined. Did you mean: 'RO'?
```

I should say that the prefix isn't missing, per se. It is defined in `basic_slots.yaml`: https://github.com/microbiomedata/nmdc-schema/blob/742b5a3c6d3b8294aad61a017f1d58d854f78a65/src/schema/basic_slots.yaml#L24

However, it doesn't make its way into the generated Python file: https://github.com/microbiomedata/nmdc-schema/blob/742b5a3c6d3b8294aad61a017f1d58d854f78a65/nmdc_schema/nmdc.py#L102-L114

I can't say I totally understand LinkML's logic for identifying all the necessary prefixes across all imported YAML files. In fact, this may be considered a bug in LinkML. Regardless, instead of waiting for LinkML to maybe fix that, a quick fix is to simply hoist all of our prefix specifications up to the top-level `nmdc.yaml` file.

These changes also add a basic smoke test to ensure that the `nmdc` Python module can be imported.


aclum (approved):

turbomam:
thanks @pkalita-lbl . I don't think I understand or trust the materialization of prefixes in generated artifacts either. I think all declared prefixes should be materialized, and any prefix used in the schema, including id_prefixes should require declaration. But that may not be a complete rule either.

I definitely don't understand the significance of the emitted prefixes declaration either.

PR #2435, opened 2025-04-21
linkml 1.9.1

turbomam:
build process passes


aclum (approved):
Approve but this should not get merged in until after 4/21/25

eecavanna (approved):

pkalita-lbl (approved):
Looks good to me. Were you thinking about making a 11.6.2 release with this change? Or can we have this roll out with 11.7.0 in the next monthly release cycle (that would be my preference)?

Issue #2440, opened 2025-05-05
Add permissible values for `FileTypeEnum` to accomodate LC-NOM analyses


kheal:
Five new permissible values:

- [x] LC FT-ICR MS Raw Data
- [x] LC FT-ICR MS Analysis Results
- [x] LC FT-ICR MS QC Plots
- [x] Direct Infusion FT-ICR MS QC Plots
- [x] Direct Infusion FT-ICR MS Analysis Results

Remove existing permissible value "FT ICR-MS Analysis Results" and migrate all of those records to "Direct Infusion FT-ICR MS Analysis Results"

PR #2442, opened 2025-05-05
Add new PVs for LC NOM workflow

kheal:
Addresses #2440 

Adds the following permissible values to FileTypeEnum:
- LC FT-ICR MS Raw Data
- LC FT-ICR MS Analysis Results
- LC FT-ICR MS QC Plots
- Direct Infusion FT-ICR MS QC Plots
- Direct Infusion FT-ICR MS Analysis Results. _This PV replaces "FT ICR-MS Analysis Results"_

I also clarified some definitions for existing permissible values.

This PR includes a migrator which migrate all data objects with the value in `data_object_type` with of "FT ICR-MS Analysis Results" to "Direct Infusion FT-ICR MS Analysis Results".

### DIFF
--- a/src/schema/basic_slots.yaml
+++ b/src/schema/basic_slots.yaml
@@ -995,13 +995,12 @@ enums:
 
       Direct Infusion FT ICR-MS Raw Data:
         description:
-          Direct infusion 21 Tesla Fourier Transform ion cyclotron resonance
-          mass spectrometry raw data acquired in broadband full scan mode
+          Direct infusion Fourier transform ion cyclotron resonance
+          mass spectrometry raw data
       
       LC FT-ICR MS Raw Data:
         description:
-          Fourier Transform ion cyclotron resonance mass spectrometry raw data acquired with liquid chromatography
-          in broadband full scan mode
+          Fourier transform ion cyclotron resonance mass spectrometry raw data acquired with liquid chromatography
 
       LC-DDA-MS/MS Raw Data:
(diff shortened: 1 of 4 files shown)


eecavanna (approved):
The migrator looks good to me, and it makes sense to me with respect to the accompanying schema change.

src/schema/basic_slots.yaml
```diff
@@ -686,8 +686,19 @@ enums:
...
+        description: FT-ICR MS based molecular formula assignment results table
+      
+      Direct Infusion FT-ICR MS QC Plots:
+        description:
+          Quality control plots for FT-ICR MS raw data acquired by direct infusion
+
+      LC FT-ICR MS Analysis Results:
+        description: LC FT-ICR MS-based molecular formula assignment results tables
+
+      LC FT-ICR MS QC Plots:
```
bmeluch, line 699: Is there a stats CSV included or is this only plots? would it be more accurate to call it "LC FT-ICR MS QC Stats"? do we care?
cc @alexandriai168

alexandriai168: We decided to only include plots in the QC Plots and instead zip the stats file with the Analysis Results.  So, i think we should keep it as "LC FT-ICR MS QC Plots"

bmeluch (approved):

turbomam:
please include me as an approver for nmdc-schema PRs

If the subject matter experts approve the PR and it's just maintenance of enumerations, then there's no need to wait for more than 24 hours for me to approve, but I would like to see them in my inbox

Issue #2444, opened 2025-05-07
Create Migrator to backfill data_category slots


hesspnnl:
In order to make data_category a required field, all data_objects in the database need to have the data_category field filled. 

Related to issue [2346](https://github.com/microbiomedata/nmdc-schema/issues/2346)

PR #2445, opened 2025-05-07
Make `data_object_type` slot required on `DataObject` and implement migrator

### LINKED ISSUE #2431: Updated modeling for data_object_type
Make this required for all data_object_set records. 

Depends on:
- [x] https://github.com/microbiomedata/issues/issues/1175

kheal:
This MR will close #2431 

It will make the slot `data_object_type` required on `DataObject`s.  I've updated all valid and invalid tests accordingly and added an invalid test that is missing this field.

Note that _in theory_ this could make data that were compliant now not compliant; though our NMDC mongo data are fully populated with regards to this field (see https://github.com/microbiomedata/issues/issues/1175).  Therefore, the migrator implemented here simply raises an error if it encounters data that do not meet this expectation.

### DIFF
--- a/nmdc_schema/migrators/migrator_from_11_7_0_to_11_8_0.py
+++ b/nmdc_schema/migrators/migrator_from_11_7_0_to_11_8_0.py
@@ -1,0 +1,28 @@
+from nmdc_schema.migrators.migrator_base import MigratorBase
+
+class Migrator(MigratorBase):
+    r"""Migrates a database between two schemas."""
+
+    _from_version = "11.7.0"
+    _to_version = "11.8.0"
+
+    def upgrade(self) -> None:
+        r"""Migrates the database from conforming to the original schema, to conforming to the new schema."""
+        self.adapter.process_each_document("data_object_set", [self.validate_data_object_type])
+
+    def validate_data_object_type(self, data_object: dict) -> dict:
+        r"""
+        If the data object does not have data_category field, add a value based on the data_object_type.
+        If the data object has any other value for data_object_type; ignore it.
+
+        >>> m = Migrator()
+        >>> m.validate_data_object_type({"id":123, "type":"nmdc:DataObject", "data_object_type":"Type"})
+        {'id': 123, 'type': 'nmdc:DataObject', 'data_object_type': 'Type'}
+        >>> m.validate_data_object_type({"id":123, "type":"nmdc:DataObject"})
+        Traceback (most recent call last):
+        ...
+        ValueError: data_object_type is required and is not present in the data object 123
+        """
+        if not data_object.get("data_object_type"):
+            raise ValueError(f"data_object_type is required and is not present in the data object {data_object.get('id')}")
+        return data_object
--- a/nmdc_schema/migrators/migrator_from_11_7_0_to_11_8_0.py
+++ b/nmdc_schema/migrators/migrator_from_11_7_0_to_11_8_0.py
@@ -12,8 +12,7 @@ class Migrator(MigratorBase):
 
     def validate_data_object_type(self, data_object: dict) -> None:
         r"""
-        Raises an exception if the document either lacks a `data_object_type`
-        field or it has a `data_object_type` field whose value is falsey.
+        Raises an exception if the document lacks a `data_object_type` field.
 
         >>> m = Migrator()
         >>> m.validate_data_object_type({"id": 123, "type": "nmdc:DataObject", "data_object_type": "Type"})
--- a/nmdc_schema/migrators/migrator_from_11_7_0_to_11_8_0.py
+++ b/nmdc_schema/migrators/migrator_from_11_7_0_to_11_8_0.py
@@ -8,16 +8,15 @@ class Migrator(MigratorBase):
 
     def upgrade(self) -> None:
         r"""Migrates the database from conforming to the original schema, to conforming to the new schema."""
-        self.adapter.process_each_document("data_object_set", [self.validate_data_object_type])
+        self.adapter.do_for_each_document("data_object_set", self.validate_data_object_type)
 
-    def validate_data_object_type(self, data_object: dict) -> dict:
+    def validate_data_object_type(self, data_object: dict) -> None:
         r"""
         Raises an exception if the document either lacks a `data_object_type`
         field or it has a `data_object_type` field whose value is falsey.
 
         >>> m = Migrator()
         >>> m.validate_data_object_type({"id": 123, "type": "nmdc:DataObject", "data_object_type": "Type"})
-        {'id': 123, 'type': 'nmdc:DataObject', 'data_object_type': 'Type'}
         >>> m.validate_data_object_type({"id": 123, "type": "nmdc:DataObject"})
         Traceback (most recent call last):
         ...
@@ -25,4 +24,3 @@ class Migrator(MigratorBase):
         """
         if not data_object.get("data_object_type"):
             raise ValueError(f"data_object_type is required and is not present in the data object {data_object.get('id')}")
-        return data_object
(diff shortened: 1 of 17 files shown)


kheal:
todo - write a migrator that throws an error if there is a data object record with this field not popualted.

kheal:
@hesspnnl - do you want to give this migrator a go?  If so, you can check out this branch and add it.

hesspnnl:
> @hesspnnl - do you want to give this migrator a go? If so, you can check out this branch and add it.

I wonder if this should be combined with https://github.com/microbiomedata/nmdc-schema/issues/2444

nmdc_schema/migrators/migrator_from_11_7_0_to_11_8_0.py
```diff
@@ -0,0 +1,28 @@
...
+
+class Migrator(MigratorBase):
+    r"""Migrates a database between two schemas."""
+
+    _from_version = "11.7.0"
+    _to_version = "11.8.0"
+
+    def upgrade(self) -> None:
+        r"""Migrates the database from conforming to the original schema, to conforming to the new schema."""
+        self.adapter.process_each_document("data_object_set", [self.validate_data_object_type])
```
eecavanna, line 11: I recommend using the `do_for_each_document` adapter method instead of this `process_each_document` adapter method. The reason is that the former does something for each document, without modifying the document, itself; whereas the latter modifies the document (in this case, overwriting it with a copy of itself). In practice, the former will be faster during the real migration.

Switching to the former would involve making the following changes: replace `process_each_document` with `do_for_each_document` and then remove the surrounding `[`...`]` from `[self.validate_data_object_type]`.

nmdc_schema/migrators/migrator_from_11_7_0_to_11_8_0.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def validate_data_object_type(self, data_object: dict) -> dict:
+        r"""
+        If the data object does not have data_category field, add a value based on the data_object_type.
+        If the data object has any other value for data_object_type; ignore it.
+
+        >>> m = Migrator()
+        >>> m.validate_data_object_type({"id":123, "type":"nmdc:DataObject", "data_object_type":"Type"})
```
eecavanna, line 19:
```suggestion
        >>> m.validate_data_object_type({"id": 123, "type": "nmdc:DataObject", "data_object_type": "Type"})
```

nmdc_schema/migrators/migrator_from_11_7_0_to_11_8_0.py
```diff
@@ -0,0 +1,28 @@
...
+        {'id': 123, 'type': 'nmdc:DataObject', 'data_object_type': 'Type'}
+        >>> m.validate_data_object_type({"id":123, "type":"nmdc:DataObject"})
```
eecavanna, line 21:
```suggestion
        >>> m.validate_data_object_type({"id": 123, "type": "nmdc:DataObject"})
```

nmdc_schema/migrators/migrator_from_11_7_0_to_11_8_0.py
```diff
@@ -0,0 +1,28 @@
...
+    _to_version = "11.8.0"
+
+    def upgrade(self) -> None:
+        r"""Migrates the database from conforming to the original schema, to conforming to the new schema."""
+        self.adapter.process_each_document("data_object_set", [self.validate_data_object_type])
+
+    def validate_data_object_type(self, data_object: dict) -> dict:
+        r"""
+        If the data object does not have data_category field, add a value based on the data_object_type.
+        If the data object has any other value for data_object_type; ignore it.
```
eecavanna, line 16:
```suggestion
        Raises an exception if the document lacks a `data_objet_type` field or the value of the field is falsey.
```

nmdc_schema/migrators/migrator_from_11_7_0_to_11_8_0.py
```diff
@@ -0,0 +1,28 @@
...
+        >>> m.validate_data_object_type({"id":123, "type":"nmdc:DataObject", "data_object_type":"Type"})
+        {'id': 123, 'type': 'nmdc:DataObject', 'data_object_type': 'Type'}
+        >>> m.validate_data_object_type({"id":123, "type":"nmdc:DataObject"})
+        Traceback (most recent call last):
+        ...
+        ValueError: data_object_type is required and is not present in the data object 123
+        """
+        if not data_object.get("data_object_type"):
+            raise ValueError(f"data_object_type is required and is not present in the data object {data_object.get('id')}")
+        return data_object
```
eecavanna, line 28: If you switch from using the `process_each_document` adapter method to using the `do_for_each_document` method, then returning the document from this function will not be necessary anymore (if removed, I'd also recommend updating this function's return value type hint from `-> dict` to `-> None`).

kheal: Excellent feedback, thanks @eecavanna.  I haven't used the `do_for_each_document` method before.

hesspnnl: That's perfect. I'll implement that thank you!

eecavanna (approved):
The migrator looks good to me, and I like the approach of raising an exception when the original data does not meet our expectations.

nmdc_schema/migrators/migrator_from_11_7_0_to_11_8_0.py
```diff
@@ -0,0 +1,26 @@
...
+        field or it has a `data_object_type` field whose value is falsey.
+
+        >>> m = Migrator()
+        >>> m.validate_data_object_type({"id": 123, "type": "nmdc:DataObject", "data_object_type": "Type"})
+        >>> m.validate_data_object_type({"id": 123, "type": "nmdc:DataObject"})
+        Traceback (most recent call last):
+        ...
+        ValueError: data_object_type is required and is not present in the data object 123
+        """
+        if not data_object.get("data_object_type"):
```
eecavanna, line 25: I want to point out that this condition will be satisfied by not only the `data_object_type` field not existing in the dictionary, but also the field _existing_ and having a falsey value.

_If_ you want to use a condition that is only satisfied be the field not existing (I'm thinking that might be the case, based on the message being passed to `ValueError()`), I'd recommend changing the condition to:

```py
if "data_object_type" not in data_object:
```

An example dictionary for which this condition would evaluate to something different from the original condition is:

```py
{
  "id": "nmdc:dobj-00-000001",
  "data_object_type": "",  # <-- field exists, but value is falsey
}
```

hesspnnl: Sounds good. Thanks for the suggestion!

nmdc_schema/migrators/migrator_from_11_7_0_to_11_8_0.py
```diff
@@ -0,0 +1,26 @@
...
+    _to_version = "11.8.0"
+
+    def upgrade(self) -> None:
+        r"""Migrates the database from conforming to the original schema, to conforming to the new schema."""
+        self.adapter.do_for_each_document("data_object_set", self.validate_data_object_type)
+
+    def validate_data_object_type(self, data_object: dict) -> None:
+        r"""
+        Raises an exception if the document either lacks a `data_object_type`
+        field or it has a `data_object_type` field whose value is falsey.
```
eecavanna, line 16: Now that the condition has been updated, I recommend updating the docstring to match it. The current docstring reflects the previous behavior.

```suggestion
        Raises an exception if the document lacks a `data_object_type` field.
```

turbomam (approved):

PR #2452, opened 2025-05-09
added titles to PVs in eluentintroductioncategoryenum

### LINKED ISSUE #2451: Add titles to PVs in EluentIntroductionCategoryEnum
Add titles to PVs in EluentIntroductionCategoryEnum so that they can be used when displaying material processing information on UI

samobermiller:
closes #2451 

Add titles to PVs in EluentIntroductionCategoryEnum so that they can be used when displaying material processing information on UI

### DIFF
--- a/src/schema/nmdc.yaml
+++ b/src/schema/nmdc.yaml
@@ -855,14 +855,18 @@ enums:
       liquid_chromatography:
         aliases:
           - LC
+        title: liquid chromatography
         description: The processed sample is introduced into the mass spectrometer through a liquid chromatography process.
       gas_chromatography:
         aliases:
           - GC
+        title: gas chromatography
         description: The processed sample is introduced into the mass spectrometer through a gas chromatography process.
       direct_infusion_syringe:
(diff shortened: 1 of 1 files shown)


kheal (approved):

turbomam:
please add me as an approver to nmdc-schema PRs

samobermiller:
> please add me as an approver to nmdc-schema PRs

@turbomam My mistake! I'll make sure to do that in the future. Is there follow up work necessary on this PR?

turbomam:
You PR is good because it adds clarifying  metadata to the permissible values. So the followup work is to continue that practice and encourage others to do the same!

We should write a enumerations and permissible values expectations guide at some point, and bring older content up to that standard when possible.

Titles (and all other metaslots) are of the most value when they add something that isn't already in another metaslot. A really high value addition is the CURIe for an ontology class from one of our preferred ontologies in the `meaning` metaslot. But that also needs docuemtnation, especially regarding what our preferred ontologies are.

PR #2461, opened 2025-05-22
Make `eluent_introduction_category` & `has_mass_spectrometry_configuration` slots on `MassSpectrometry` required and implement migrator 

### LINKED ISSUE #2455: Make modifications to MassSpectromery's eluent_introduction_category
- Make MassSpectromery's slots eluent_introduction_category, has_mass_spectrometry_configuration required
- Add a rule that if MassSpectromery's eluent_introduction_category is liquid_chromatography, then has_chromatography_configuration is required
- create a migrator to validate the slots have been filled. 

Related to https://github.com/microbiomedata/issues/issues/1021

hesspnnl:
- Make MassSpectromery's slots eluent_introduction_category, has_mass_spectrometry_configuration required
- create a migrator to that will raise an error if the slots do not exist. Data for these slots has already been back filled. 
- Tested the migrators locally against mongo db data

This MR will close #2455

### DIFF
--- a/nmdc_schema/migrators/partials/migrator_from_11_7_0_to_11_8_0/migrator_from_11_7_0_to_11_8_0_part_3.py
+++ b/nmdc_schema/migrators/partials/migrator_from_11_7_0_to_11_8_0/migrator_from_11_7_0_to_11_8_0_part_3.py
@@ -13,7 +13,7 @@ class Migrator(MigratorBase):
 
     def validate_mass_spec_slots(self, data_generation_record: dict) -> None:
         r"""
-        If the data object does not have eluent_introduction_category or has_mass_spectrometry_configuration AND those keys do not have a value, raise a ValueError.
+        If the data object does not have eluent_introduction_category or has_mass_spectrometry_configuration OR those keys do not have a value, raise a ValueError.
 
         >>> m = Migrator()
         >>> m.validate_mass_spec_slots({"id": 123, "type": "nmdc:MassSpectrometry", "eluent_introduction_category": "gas_chromatography", "has_mass_spectrometry_configuration": "nmdc:mscon-123"})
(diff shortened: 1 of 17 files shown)


src/data/invalid/MassSpectrometry-invalid-lc-without-config.yaml
kheal, line 1: We can get rid of this test since we're not including the rule in this PR.

kheal (approved):

nmdc_schema/migrators/partials/migrator_from_11_7_0_to_11_8_0/migrator_from_11_7_0_to_11_8_0_part_3.py
```diff
@@ -0,0 +1,38 @@
...
+        >>> m.validate_mass_spec_slots({"id": 123, "type": "nmdc:MassSpectrometry", "has_mass_spectrometry_configuration": "nmdc:mscon-123"})
+        Traceback (most recent call last):
+        ... 
+        ValueError: `eluent_introduction_category` is required and is not present in the data object 123
+        """
+        # get the eluent_introduction_category and has_mass_spectrometry_configuration fields
+        if data_generation_record.get("type") == "nmdc:MassSpectrometry":
+            eluent_introduction_category = data_generation_record.get("eluent_introduction_category")
+            
+            if eluent_introduction_category is None:
```
eecavanna, line 33: This checks whether **either** (a) the field does not exist, or (b) the field _does_ exist and the value in the field is `None`. As a reminder, the `.get()` function will return `None` in both of those situations.

If you want to _only_ check whether the field exists, you can use:

```py
            if "eluent_introduction_category" not in data_generation_record:
```

I have the same comment about the `has_mass_spec_config` counterpart below.

This may be a non-issue (I'm not sure whether the docstring or the `ValueError` messages are referring to the field, the value in the field, or both).

eecavanna (approved):
The migrator looks good to me. I left one [PR comment](https://github.com/microbiomedata/nmdc-schema/pull/2461/files#r2105536313) about checking whether a key exists _versus_ checking whether it exists and the value of its item is not `None`.

nmdc_schema/migrators/partials/migrator_from_11_7_0_to_11_8_0/migrator_from_11_7_0_to_11_8_0_part_3.py
hesspnnl: I updated the docstring for more clarity

turbomam (commented):
Wow, you have been prolific recently

I have the same questions here that I had about invalid examples files in the following PR. You don't need to address that in both places.

- #2465

turbomam (approved):

PR #2463, opened 2025-05-23
Add "Direct Infusion FT-ICR MS Analysis Results" to migrator_from_11_7_to_11_8_part_2 mapping

hesspnnl:
From schema release 11.7 there is a new data_object_type. Add `Direct Infusion FT-ICR MS Analysis Results` to the migrator mapping.


kheal (approved):

kheal:
Added @turbomam for review as an FYI, but this is to fix a migrator that will need to be run for the June release.

turbomam (approved):

PR #2485, opened 2025-06-16
UCUM constraint implemented as LinkML annotation and Python script, with intentionally failing UCUM test

turbomam:
one strength of this specification-by-annotation is the ability to share one (merged) schema file, not a schema file and a separate TSV specification

PR #2439, opened 2025-04-29
DONOTMERGE: Create and apply `QuantityValueCelsius` as `range` for select temperature related slots and implement migrator

turbomam:
This PR creates and applies `QuantityValueCelsius` with `has_unit` of 'Cel' to standardize temperature measurements in the schema. 

- Creates `QuantityValueCelsius` subclass in attribute_values.yaml, with an annotation indicating unit (`ucum_unit`) for `submission-schema` propagation
- Changes `range` for `temperature` slot from `QuantityValue` to `QuantityValueCelsius` in core.yaml
- Changes `range` for these biosample slots from `QuantityValue` to `QuantityValueCelsius` by updating mixs.yaml:
  - air_temp
  - annual_temp
  - avg_temp
  - host_body_temp
  - samp_store_temp
  - season_temp
  - surf_temp
  - temp_out
- Updates existing valid tests that use these slots so they have `range` `QuantityValueCelsius` and have `has_unit` 'Cel'
- Creates invalid test with example `QuantityValueCelsius` where `has_unit` is 'F' 
- Creates invalid test with biosample that has temp range of `QuantityValue`
- Creates migrator transforming various iterations of 'Celsius' that exist in mongo into 'Cel'
   - Migrator has been tested locally against mongo database

** Slots were selected because they have 'temp' in the name, actually capture temperature and have a current range of `QuantityValue`. Confirmed that migrator sufficient for existing mongo data using script 'nmdc_schema_temp_checker.py' in assets/misc (output on google drive). 
** Mixs.yaml was updated by adding temp related biosample slots to yq-for-mixs_subset_modified.txt and setting their range to `QuantityValueCelsius`, then running commands `make squeaky-clean mixs-yaml-clean` and `make src/schema/mixs.yaml all test`

PR #2488, opened 2025-06-17
Create QuantityValueWithUnit and implement temperature as example

samobermiller:


sierra-moxon:
Is it possible to add mappings to the enum PV "Cel"?
https://units-of-measurement.org/Cel has some to consider in the skos:exact_match field and I think it would be helpful to both define this as a UCUM-based PV as you have done already but also consider extending the definition of the PV to include the skos:definition from that site too.

kheal:
> Is it possible to add mappings to the enum PV "Cel"? https://units-of-measurement.org/Cel has some to consider in the skos:exact_match field and I think it would be helpful to both define this as a UCUM-based PV as you have done already but also consider extending the definition of the PV to include the skos:definition from that site too.

@sierra-moxon I _think_ this sounds like a great idea, but I'm not following completely, could you add some "suggested edits" to the files in this PR to demonstrate what you mean?

src/schema/nmdc.yaml
```diff
@@ -981,6 +981,14 @@ enums:
       aqueous_layer:
       non_polar_layer:
 
+  UnitEnum:
+    permissible_values:
+      Cel:
+        aliases:
+          - Celsius
```
sierra-moxon, line 988:
```suggestion
          - Celsius
          - C
          - degreeCelsius
          - degree Celsius
```
I am not sure LinkML will handle the special character for `°` correctly, so I have not included it here; however, I think it would be another good alias to capture. I will test with `°C` separately from this PR.

src/schema/nmdc.yaml
```diff
@@ -981,6 +981,14 @@ enums:
...
+        description: Cel, the UCUM representation of degrees Celsius
```
sierra-moxon, line 989:
```suggestion
        description:  >-- 
           The Unified Code for Units of Measure (UCUM) representation of degrees Celsius; a SI unit of temperature equal to one unit Kelvin.
```

src/schema/nmdc.yaml
```diff
@@ -981,6 +981,14 @@ enums:
...
+
```
sierra-moxon, line 990:
```suggestion
exact_mappings:
   - UO:0000027
   - wikidata:Q25267
   - qud:DEG_C
   - OM:degreeCelsius
```

Note: the prefix in UCUM for the third mapping here is referenced with prefix `QUDT` at the source, but the NMDC schema prefixes this identifier with `qud`.  Bioregistry uses a third variant, `qudt`.  I've kept it inline with the NMDC schema in this PR, but will adjust in another PR. 

`NERC_P06:UPAA` is another exact_mapping we should add to this list, but it requires:
- adding `NERC_P06` to the nmdc.yaml "imports" section with a URL expansion like this:
```
NERC_P06: http://vocab.nerc.ac.uk/collection/P06/current/
```
- and, adding this as a prefix to bioregistry here: https://github.com/biopragmatics/bioregistry/issues/new?labels=New%2CPrefix&template=new-prefix.yml&title=Add+prefix+%5BX%5D
I will do that in another PR.

I did not include `OBOE:Celsius` because when I went to download and view the ontology from BioPortal I got all sorts of strange warnings about the content of the file having malware (likely a red herring), but also not browsable in BioPortal, not in the obofoundry, not viewable in protege or even text edit...so someone with experience with that ontology is welcome to add it at a later date.

sierra-moxon:
> > Is it possible to add mappings to the enum PV "Cel"? https://units-of-measurement.org/Cel has some to consider in the skos:exact_match field and I think it would be helpful to both define this as a UCUM-based PV as you have done already but also consider extending the definition of the PV to include the skos:definition from that site too.
> 
> @sierra-moxon I _think_ this sounds like a great idea, but I'm not following completely, could you add some "suggested edits" to the files in this PR to demonstrate what you mean?

Oh sure, sorry - I thought this kind of metadata was the key point in making an enumeration in the model itself (that and not adding "required" separately in slot_usage, and retaining a small check in linkml-validate that the unit itself was at least part of a larger enumerated list of permissible values).

samobermiller:
> > > Is it possible to add mappings to the enum PV "Cel"? https://units-of-measurement.org/Cel has some to consider in the skos:exact_match field and I think it would be helpful to both define this as a UCUM-based PV as you have done already but also consider extending the definition of the PV to include the skos:definition from that site too.
> > 
> > 
> > @sierra-moxon I _think_ this sounds like a great idea, but I'm not following completely, could you add some "suggested edits" to the files in this PR to demonstrate what you mean?
> 
> Oh sure, sorry - I thought this kind of metadata was the key point in making an enumeration in the model itself (that and not adding "required" separately in slot_usage, and retaining a small check in linkml-validate that the unit itself was at least part of a larger enumerated list of permissible values).

@sierra-moxon Thank you for reviewing and outlining these steps! I think there's also some overlap with Mark's PR #2485 which depending on his review we could add to this PR or merge in separately

assets/misc/nmdc_schema_temp_checker.py
```diff
@@ -0,0 +1,87 @@
+from linkml_runtime import SchemaView
```
turbomam, line 1: please put scripts in `nmdc_schema/` or `src/scripts/`

assets/misc/nmdc_schema_temp_checker.py
```diff
@@ -0,0 +1,87 @@
...
+from src.scripts.make_typecode_to_class_map import get_collection_names_for_class_name
+from nmdc_api_utilities.collection_search import CollectionSearch
+import pandas as pd
+
+schema_url = "https://raw.githubusercontent.com/microbiomedata/nmdc-schema/refs/heads/main/nmdc_schema/nmdc_materialized_patterns.yaml"
```
turbomam, line 6: this is fine for now, but it's better to get the schema from the repo package fiels, not over HTTP

src/schema/attribute_values.yaml
```diff
@@ -44,6 +44,16 @@ classes:
     mappings:
       - schema:QuantityValue
 
+  QuantityValueWithUnit:
+    class_uri: nmdc:QuantityValueWithUnit
+    is_a: QuantityValue
+    description: A QuantityValue whose unit value must be in UnitEnum, which contains UCUM representations of various metrics.
```
turbomam, line 50:
```suggestion
    description: A QuantityValue in which the has_unit field must be populated, and its value must come from the UnitEnum. Slots taking this class as their range are responsible for providing an annotation declaring which specific permissible values from the enumeration are acceptable from a business logic/science domain perspective.
    see_also
      - https://github.com/microbiomedata/nmdc-schema/pull/2485
```

turbomam (changes requested):
Great! This PR covers all of the bases except for the units annotations on slots and corresponding python code for testing patterns in both the schema and the data. I suggested adding a `see_also` for that.

assets/misc/nmdc_schema_temp_checker.py
turbomam: There are some illustrations of that in https://github.com/microbiomedata/nmdc-schema/blob/main/tests/test_nmdc_data.py. There are other ways to do it too.

assets/misc/nmdc_schema_temp_checker.py
samobermiller: @turbomam this isn't a script that's regularly run, its just how I assessed slot useage for 'temp' using SchemaView. I was asked to upload it to the assets folder as a use case example of SchemaView. Is it better to put in the nmdc_schema or src folders even if its not being run to directly generate schema related things?

samobermiller:
@eecavanna I've added you as a reviewer for this migrator but just a note its exactly the same as the one you reviewed in #2439 just with a different class name

assets/misc/nmdc_schema_temp_checker.py
turbomam: I like the spirit of this, but if it is being checked in as an illustration of best practice, it will need much more documentation, and tighter integration with pyproject.toml. For example, it imports a package (`nmdc_api_utilities`) that isn't in `tool.poetry.dependencies` or `tool.poetry.group.dev.dependencies`

Then this script should be given a `tool.poetry.scripts` entry and or a Makefile target

assets/misc/nmdc_schema_temp_checker.py
turbomam: I would like to hear more about

  • Downgrading rdflib (6.3.2 -> 6.2.0)
  • Installing matplotlib (3.10.3)
  • Installing nmdc-api-utilities (0.3.9)

nmdc_schema/migrators/partials/migrator_from_11_7_0_to_11_8_0/migrator_from_11_7_0_to_11_8_0_part_5.py
```diff
@@ -0,0 +1,58 @@
...
+
+    _from_version = "11.8.0.part_4"
+    _to_version = "11.8.0.part_5"
+
+    def upgrade(self) -> None:
+        r"""Migrates the database from conforming to the original schema, to conforming to the new schema."""
+        self.adapter.process_each_document("biosample_set", [self.standardize_unit_celsius])
+        self.adapter.process_each_document("material_processing_set", [self.standardize_unit_celsius])
+        self.adapter.process_each_document("configuration_set", [self.standardize_unit_celsius])
+        self.adapter.process_each_document("storage_process_set", [self.standardize_unit_celsius])
```
eecavanna, line 15: Note to self (@eecavanna): If we do incorporate this into schema version 11.8.0, update the migration notebook to account for the fact that these specific collections are involved.

PR #2491, opened 2025-06-18
Update 11.7->11.8 migrator data_category_type mapping to include "rRNA Filtered Sequencing Reads"

hesspnnl:
Bug identified where a record has the category "rRNA Filtered Sequencing Reads" but this was not included in the mapping yaml file. Add this value to the yaml file.


kheal:
@hesspnnl I'm pretty sure "rRNA Filtered Sequencing Reads" should be `instrument_data` but not positive, @aclum can you weigh in?

eecavanna (approved):

PR #2541, opened 2025-07-09
`has_unit` assertions for `src/data/valid/Biosample-possibly-exhaustive.yaml`

turbomam:
Adds `has_unit` to all `QuantityValue` slots in `src/data/valid/Biosample-possibly-exhaustive.yaml`

Some reports are included. We can remove them from the branch or find a better directory. The scripts that create them don't consistently use `click` yet and have been flagged with various Copilot nitpicks below. 

This PR does not add numerical values to all of the `QuantityValue` slots. see
- https://github.com/microbiomedata/nmdc-schema/discussions/2540

There are some internal inconsistencies in how our NMDC version of MIxS ~ 6.0 expresses preferred units in `annotations`, `examples` and `description`. Since we should really move to GSC's MIxS 6.2 or 7.0, I haven't taken action on that. see
- https://github.com/microbiomedata/nmdc-schema/issues/2542

Otherwise all units are in the spirit of the MIxS preferred units guidelines and all pass validation at the UOM dynamic website (or are skipped for legitimate reasons). see
- https://github.com/microbiomedata/nmdc-schema/blob/2539-add-units-etc-to-quantityvalues-in-srcdatavalidbiosample-possibly-exhaustiveyaml/src/scripts/check_units.py

Along with many of you working on the UCUM tasks, this code has identified some slots for which Quantity Value really doesn't seem like the right range. That can be tracked down by looking for slots that have `#` comments in `src/data/valid/Biosample-possibly-exhaustive.yaml`, or by looking for the following units

- `[arb'U]{unitless}`
- `[pH]`
- `{ratio}`

I have made heavy use of UCUM's `[arb'U]{some_annotation}` style for representing non-UCUM units. I think most of those are justified, but slots using these could probably all be changed to scalars

- `[arb'U]{people}`
- `[arb'U]{pets}`
- `[arb'U]{plants}`

### DIFF
--- a/src/data/valid/Biosample-possibly-exhaustive.yaml
+++ b/src/data/valid/Biosample-possibly-exhaustive.yaml
@@ -1,3 +1,6 @@
+# exhaustive as of 2025-07-08
+# but will become invalid when has_unit becomes required for QuantityValue
+# and may imply that some undesirable data patterns are OK
 id: nmdc:bsm-99-4444444
 type: nmdc:Biosample
 name: Sample Exhaustive Biosample instance. Although all of these values should pass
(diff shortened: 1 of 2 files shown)


turbomam:
all src/data/valid/Biosample-possibly-exhaustive.yaml QVs have UCUM legal units

still need to double check against MIxS and possibly remodel some

may want to reorganize the new src/scripts and/or delete the generated files in the root of the project

turbomam:
1. `flatten_quantity_values.py` extracted 173 QuantityValue structures from the YAML file and created quantity_values.tsv
2. `check_units.py` then read that TSV file and validated 51 unique units against the UOM website, achieving a 100% success rate with only 2 units skipped (mm[Hg] and {ratio} as dimensionless/custom units)

PR #2557, opened 2025-07-14
Deprecate `Metagenome Bins` FileTypeEnum permissible value

aclum:
This PR has the first stage of deprecation for `Metagenome Bins` as a FileTypeEnum permissible value.


turbomam (commented):
awesome. will we remove it from the schema at some point in the future?

aclum:
Yes, the next release following our deprecation protocol.

turbomam (approved):

PR #2564, opened 2025-07-16
Add alternative_names slot to class Biosample

aclum:
This PR added slot alternative_names to class Biosample and updates valid example data.


turbomam:
I guess the improved description (don't put ids with targeted slots in her) should go in the global definition

turbomam (approved):

PR #2570, opened 2025-07-17
Refactor sibling migrators into partials of new top-level migrator

eecavanna:
On this branch, I converted two top-level migrators (both involving the same schema version jump) into "partial" migrators, and implemented a new, top-level migrator that invokes them under the hood. This is the normal process in this situation—it was skipped for the `.0` release by accident, due to one of the migrators being misnamed.


eecavanna:
I am utilizing my high level of access to this repository to merge this PR into `main` "after hours" with review by anyone (except Copilot). My rationale is that I already discussed releasing schema version 11.9.0 today with @turbomam, who approved of it at a high level, and I think we both assumed there was not a bug in the migrators at the time. I am applying that approval to this PR.

PR #2578, opened 2025-07-17
Refactor to separate out AttributeValue classes and slots

aclum:
This PR refactors classes and slots that is_a AttributeValue for the purposes of being reused by bertron.


mslarae13 (approved):
LGTM!

src/schema/mixs.yaml
```diff
@@ -7,7 +7,8 @@ notes:
 id: https://raw.githubusercontent.com/microbiomedata/nmdc-schema/main/src/schema/mixs.yaml
 imports:
   - linkml:types
-  - attribute_values
+  - attribute_values_classes
```
turbomam, line 10: that change has to be configured in a helper file

src/schema/nmdc.yaml
```diff
@@ -14,8 +14,12 @@ license: https://creativecommons.org/publicdomain/zero/1.0/
 version: 0.0.0
 
 imports:
+  - linkml:types
+  - attribute_values_classes
   - annotation # also brings core and portal_*
   - workflow_execution_activity
+  - basic_classes
+  - basic_slots
```
turbomam, line 22: can't these be imported indirectly through other imports? i'd liek to keep it succinct.

src/schema/nmdc.yaml
```diff
@@ -1030,6 +1034,9 @@ enums:
         is_a: organic_layer
 
 slots:
+  biomaterial_purity:
```
turbomam, line 1037: this needs metadata

turbomam (changes requested):

src/schema/mixs.yaml
sierra-moxon: helper file?

src/schema/nmdc.yaml
sierra-moxon: Which sub-files would you like the basic_classes and basic_slots to come in via?

src/schema/nmdc.yaml
sierra-moxon: it sure does; but not in this PR.  This is a direct copy of what exists currently in the schema.

sierra-moxon:
> If the goal was to make attribute_values_classes.yaml independent, I still don't think that's the case
> 
> ```shell
> linkml generate linkml --format yaml src/schema/attribute_values_classes.yaml
> ```
> 
> > ValueError: No such slot name as an attribute of PersonValue ancestors or as a slot definition in the schema

Fixed.

src/schema/mixs.yaml
turbomam: `assets/yq-for-mixs_subset_modified.txt`

```
'.imports += ["attribute_values"]'
```

sierra-moxon: I just changed the name of the file back to attributes_values.yaml -- it didn't need to be changed.  The name change was from the initial attempt to separate classes from slots.  At the end of the day, we decided not to do this separation but the file name stuck.

turbomam (approved):

PR #2582, opened 2025-07-25
Generate JSON formatted schema definition

### LINKED ISSUE #2581: Provide schema in JSON format as published artifact
Currently the `nmdc-server` front-end code uses `nmdc-schema` for various things like pulling help text out of descriptions and annotations, getting enum permissible values, etc. It does this by:

1. Installing `nmdc-schema` as a front-end dependency through a somewhat non-standard mechanism (that's a separate issue that is orthogonal to this one)
2. Parsing the `nmdc_materialized_pattern.yaml` file and using the result

The issue with using the YAML file in step 2 is that it doesn't allow for type safety in the `nmdc-server` front-end code. This is because the YAML loading is done via Webpack which happens downstream of TypeScript's type checks. TypeScript does understand [JSON files as modules](https://www.typescriptlang.org/tsconfig/#resolveJsonModule), so if we could import a JSON file instead we'd have much better type checking. 

This task is to produce a JSON-formatted version of `nmdc_materialized_patterns.yaml` and ensure it is distributed as a published artifact.

cc: @samobermiller

pkalita-lbl:
Fixes #2581 

These changes generate a new JSON formatted version of the schema. This will allow for improvements in the `nmdc-server` front-end code. See issue for details.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,7 +96,9 @@ pydantic:
 all: site
 site: clean site-clean gen-project gendoc \
 nmdc_schema/gold-to-mixs.sssom.tsv \
-nmdc_schema/nmdc_materialized_patterns.schema.json nmdc_schema/nmdc_materialized_patterns.yaml \
+nmdc_schema/nmdc_materialized_patterns.schema.json \
+nmdc_schema/nmdc_materialized_patterns.yaml \
+nmdc_schema/nmdc_materialized_patterns.json \
 migration-doctests \
 prefixmaps \
 pydantic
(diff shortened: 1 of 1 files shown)


turbomam:
I love `yq` and approve this, but we already have this. Is it too bloated?

```make
nmdc_schema/nmdc_materialized_patterns.schema.json: nmdc_schema/nmdc_materialized_patterns.yaml
	$(RUN) gen-json-schema \
		--closed \
		--include-range-class-descendants \
		--top-class Database $< > $@
```

turbomam (approved):

pkalita-lbl:
> but we already have this

That's a different format. `gen-json-schema` generates [JSON Schema](https://json-schema.org/). What these changes produce is the LinkML schema definition serialized as JSON.

PR #2583, opened 2025-07-29
Modeling updates to execution_resource and processing_institution for WorkflowExecution

aclum:
This PR is to get the changes I want to the branch that I intended to work on. Once that is fixed, I'll do a PR to main.

This PR makes the following changes:
Makes execution_resource optional
removes ExecutionResourceEnum permissible values that are organizations
Adds EMSL-Tahoma as a ExecutionResourceEnum permissible value
Makes processing_institution required
adds NMDC to ProcessingInstitutionEnum
removes some MetagenomeSequencing invalid example data files

If these everyone agrees to these changes this PR will need an accompanying migrator.


kheal:
Seems reasonable.  I can give a more thorough review when it needs to be merged into main.

kheal (approved):

PR #2590, opened 2025-08-05
Add syntax constraint to `gene_function_id` for class `FunctionalAnnotationAggMember`

aclum:
This PR adds
- a `structured_pattern` syntax constraint to `gene_function_id` for class `FunctionalAnnotationAggMember`
- invalid example data files

Validation patterns come from `bioregistry.io`.


src/schema/nmdc.yaml
```diff
@@ -382,7 +382,11 @@ classes:
         range: AnnotatingWorkflow
       count: 
           description: The number of sequences (for a metagenome or metatranscriptome) or spectra (for metaproteomics) associated with the specified function.
-
+      gene_function_id:
+        structured_pattern:
+          syntax: "(COG:COG\\d+|PFAM:PF\\d{5}|KEGG.ORTHOLOGY:K\\d+)"
```
turbomam, line 387: My hunch is that this is appropriate

But if you aren't using any of the  shema's`settings` in the `syntax`, then you might as well make this a `pattern`, not a `structured_pattern`. If you were using `settings`, they would appear as symbols inside of `{}`s in the `syntax`.  I'm not sure how that combines with character count constraints like `{5}`

If that doesn't' make sense, then we can discuss in whatever channels you prefer

eecavanna (approved):
The migrator looks good to me. Once the comments about the "print" statement have been resolved, I'm comfortable with this migrator being merged in.

eecavanna:
(OOO, just resolving some notifications during some downtime)

nmdc_schema/migrators/migrator_from_11_9_1_to_11_10_0.py
```diff
@@ -0,0 +1,49 @@
+from nmdc_schema.migrators.migrator_base import MigratorBase
```
kheal, line 1: I think this migrator is meant to be in a different PR :).

src/schema/nmdc.yaml
```diff
@@ -382,7 +382,9 @@ classes:
         range: AnnotatingWorkflow
       count: 
           description: The number of sequences (for a metagenome or metatranscriptome) or spectra (for metaproteomics) associated with the specified function.
-
+      gene_function_id:
```
kheal, line 385: Is there already a valid data example?  If so it'd be helpful to add one with examples of these three.

turbomam (approved):

PR #2597, opened 2025-08-09
Update method signature to comply with base class

eecavanna:
The `Migrator` base class was recently updated so that its `upgrade` method has a `commit_changes` parameter. One of the newly-implemented migrators did not have that parameter in its `upgrade` method, even though that class inherited from `Migrator`. As a result, when calling the `upgrade` method, Python raised a `TypeError` (as shown [here](https://github.com/microbiomedata/nmdc-runtime/pull/1127#issuecomment-3169672830)).

On this branch, I added the `commit_changes` parameter to the `upgrade` method, along with a `TODO` about existing migrators.


eecavanna:
Once the tests pass here on GHA, I'll merge this in (without review) in order to unblock https://github.com/microbiomedata/nmdc-runtime/pull/1127.

PR #2620, opened 2025-08-25
Add structured aliases to InstrumentModelEnum permissible values

aclum:
This PR adds structured aliases from GOLD to InstrumentModelEnum permissible values


aclum:
We'd like to move this PR forward so @sujaypatil96 can move forward with pulling EMP500 amplicon records from GOLD.

PR #2653, opened 2025-10-03
Evaluate slot excuses for UCUM

mslarae13:
I added 2 columns to units/output/schema_units_excuses.tsv

the 1st is my suggested change. Either what the unit should be or how we should change the range.
the 2nd are notes about the decision, next steps, or questions.


mslarae13:
@turbomam I think the changes and comments on https://github.com/microbiomedata/nmdc-schema/blob/2652-analyze-slot-excuses-for-units/units/output/schema_units_excuses.tsv can be used to move ahead with the excuses slots. Will you be able to make the updates for the identified slots? And if so, can we do it here.

mslarae13:
PR in draft still .. pending response from @turbomam about (if there are any) next steps and if it should be done in this PR or separate.

turbomam:
Good review. Closing in favor of
- #2686 
- #2687

PR #2662, opened 2025-10-08
Add slot to capture long read sequencing DNA extraction method

mslarae13:
Closes https://github.com/microbiomedata/issues/issues/413 

- Consolidates the jgi-mg, jgi-mt, and emsl .yaml files into 1 user facility .yaml.
- Add a slot to capture DNA extraction method for long read sequences & differentiate short (MG) from long (MG-LR)
- Remove slots that NMDC doesn't need to store. These exist in the `submission-schema` & do not need to be stored by NMDC at the moment.


src/schema/portal_user_facility.yaml
```diff
@@ -0,0 +1,163 @@
...
+    slot_group: JGI-Metatranscriptomics
+    recommended: true
+  dna_lr_isolate_meth:
+    description: Describe the method/protocol/kit used to extract DNA/RNA.
+    title: DNA isolation method for long read sequencing
+    examples:
+      - value: phenol/chloroform extraction
+    rank: 16
+    string_serialization: '{text}'
+    slot_group: JGI-Metagenomics
```
mslarae13, line 60: @pkalita-lbl is this right? is there a separate slot group for the long read metagenomics in the submission portal? I don't think there is, but I'm not 100% on that.

And I think, once this is done, I'll need to update https://github.com/microbiomedata/submission-schema/blob/f8fd2a795b2ee006275e90272247bae20c47b4cd/src/nmdc_submission_schema/schema/nmdc_submission_schema_base.yaml#L727

ssarrafan:
pinged Mark to review

src/schema/portal_user_facility.yaml
```diff
@@ -0,0 +1,154 @@
...
+      as estimated % of the organism in that culture.
+    title: RNA expected organisms
+    examples:
+      - value: expected to contain microbes (59%) fungi (30%), viruses (10%), tadpoles
+          (1%)
+    rank: 14
+    string_serialization: '{text}'
+    recommended: true
+    deprecated: true; as of 2025-10-10, NMDC doesn't need to store this JGI metadata slot & it's now captured directly in the submission portal schema to support submissions
+  dna_lr_isolate_meth:
```
mslarae13: @turbomam this is the only new slot

PR #2671, opened 2025-10-13
Enforce ExternalIdentifier range for external identifier mixins

aclum:
This PR updates the default range to ExternalIdentifiers for mixins in the external_identifiers.yaml by adding is_a: external_database_identifiers. Previously these mixins showed the default range of string.


aclum:
remove default range from the external identifiers file, find the external identifiers slot and update that range to a curie.

turbomam (approved):

PR #2675, opened 2025-10-17
Raise minimum Python version to 3.10

pkalita-lbl:
Python 3.9 will be going end-of-life this month. These changes update the minimum Python version to 3.10. 

I updated the main testing workflow to run under Python 3.10 and 3.13 (previously 3.9 and 3.12). The first version of Python 3.14 was released just over a week ago, and it seems like at least one of this project's dependencies is not compatible with 3.14 yet (at least from what I saw in my brief local test). 

A number of the other GitHub Actions workflows were using Python 3.9. All of the non-test workflows have been updated to use Python 3.12.

### DIFF
--- a/.github/workflows/check-links.yaml
+++ b/.github/workflows/check-links.yaml
@@ -26,8 +26,10 @@ jobs:
         # Reference: https://github.com/actions/checkout
         uses: actions/checkout@v5
       - name: Set up Python
-        uses: actions/setup-python@v3
-        with: { python-version: 3.9 }
+        uses: actions/setup-python@v6
+        with:
+          python-version: '3.12'
+          cache: 'poetry'
       - name: Install Poetry
         uses: snok/install-poetry@v1.3
       - name: Install Python dependencies (including "docs" extras)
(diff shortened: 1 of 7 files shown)


.github/workflows/check-links.yaml
```diff
@@ -26,8 +26,10 @@ jobs:
         # Reference: https://github.com/actions/checkout
         uses: actions/checkout@v5
       - name: Set up Python
-        uses: actions/setup-python@v3
-        with: { python-version: 3.9 }
+        uses: actions/setup-python@v6
+        with:
+          python-version: '3.12'
```
eecavanna: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-file-input

FYI, this GHA can be configured to read the Python version spec from pyproject.toml.

pkalita-lbl: Oh interesting. I didn't know that.

Do you know what happens when `pyproject.toml` says `python = "^3.10"`? Does it just use 3.10 in that case? 

When we migrate this repo to `uv`, I would imagine we'd drop in a `.python-version` file (which Poetry doesn't support 😢) to specify the standard Python version for development.

PR #2695, opened 2025-10-21
Add slot to `WorkflowExecution` and `DataObject` classes to point to newer versions of associated `WorkflowExecution`

### LINKED ISSUE #2693: Add slot to `WorkflowExecution` and `DataObject` classes to point to newer versions of associated WorkflowExecution and DataObject
See https://github.com/microbiomedata/issues/discussions/1396 for full discussion.

The goal here is to have a slot in the schema so that (when populated) one is able to tell if a [WorkflowExecution](https://microbiomedata.github.io/nmdc-schema/WorkflowExecution/) or resultant [DataObject](https://microbiomedata.github.io/nmdc-schema/DataObject/) record has been replaced due to workflow re-run.

kheal:
This PR

- Adds slot `superseded_by` to the schema
- Adds the slot `superseded_by` to `WorkflowExecution` and `DataObject` classes
- Constrains the range of `superseded_by` to the specific type of `WorkflowExecution` subclass
- Adds example valid and invalid data
- Closes #2693

### DIFF
--- a/src/schema/basic_classes.yaml
+++ b/src/schema/basic_classes.yaml
@@ -693,9 +693,9 @@ slots:
   superseded_by:
     range: WorkflowExecution
     description: >-
-      Links a data object or workflow execution to a newer workflow execution that 
-      supersedes it, marking this record as outdated. The linked workflow execution 
-      produced improved or corrected results that should be used instead of this record.
+      Links a DataObject or WorkflowExecution record to a newer WorkflowExecution that 
+      supersedes it, marking this record as outdated. The linked WorkflowExecution or
+      resultant DataObjects should be used in favor of this record.
 
 enums:
   CreditEnum:


aclum:
I'd like to discuss DataObjects referring to a newer WorkflowExecution instead of or in addition to individual DataObjects over time mostly because workflow outputs can change over time there may not be 1:1 of the replacement files and or may involve increasingly complicated logic whereas the WorkflowExecution subclasses are more stable over time. 

Tagging @eecavanna  and @shreddd because this will need to be coordinated with corresponding changes to referential integrity checks in nmdc-runtime. If there is a newer version and something goes wrong and that version needs to be deleted the reference to that needs to be removed.

kheal:
@aclum  Now this slot only references the newer WorkflowExecution that makes the records "outdated".  I think this is much cleaner and more maintainable, thanks for the feedback.

aclum:
Looks great

PR #2723, opened 2025-11-24
Modeling updates to ecosystem_path_id 

aclum:
This PR adds ecosystem_path_id to class Biosample, updates the range to an integer, and updates example data.


turbomam:
> I've removed the swp file.

do you guys think we should add `*.swp` to `.gitignore`?

PR #2822, opened 2026-02-12
Add qc metrics to `MetaproteomicsAnalysis` and add permissible values to `FailureWhatEnum` to accompany these

### LINKED ISSUE #2821: Add slots to `MetaproteomicsAnalysis` to capture qc-related workflow data and associated permissible values to `FailureWhatEnum`

kheal:
Closes #2821 

This PR adds six new quality control metric slots to the `MetaproteomicsAnalysis` class:

- `peptide_to_spectrum_match_count` - Total number of MS2 spectra with FDR-passing peptide matches
- `peptide_to_spectrum_match_rate` - Percentage of peptide-to-spectrum matches (QuantityValue)
- `unique_peptide_seq_count` - Number of distinct peptide sequences identified
- `razor_protein_count` - Minimal protein set after Razor Protein parsimony
- `mean_peptide_count` - Average peptides per protein
- `total_protein_count` - Total distinct proteins identified

Also adds a permissible value to `FailureWhatEnum` to accompany these workflow statistics
- `low_peptide_to_spectrum_match_rate` - Rate of peptide assignment during the analysis is not sufficient.

Includes two examples file demonstrating usage of the new slots and an invalid example.

### DIFF
--- a/src/schema/basic_classes.yaml
+++ b/src/schema/basic_classes.yaml
@@ -1128,15 +1128,15 @@ enums:
         description: A process ran but did not produce any output. Ie binning ran but did not produce any medium or high quality bins.
       other:
         description: A lab process or analysis workflow has failed in a way that has not been captured by the available values yet. Please use slot 'qc_comment' to specify details.
-      low_peptide_to_spectrum_match_rate:
-        description: Rate of peptide assignment during the analysis is not sufficient.
       low_molecular_formula_assignment:
         description: Number of peaks with assigned molecular formulas during the workflow is not sufficient.
       low_metabolite_assignment:
         description: Number of metabolites identified during the Metabolomics or Lipidomics workflow is not sufficient.
       low_c13_isotopologue_assignment:
         description: Number of 13C isotopologues identified during the Metabolomics or Lipidomics workflow is not sufficient.
-
+      low_peptide_to_spectrum_match_rate:
+        description: Rate of peptide assignment during the analysis is not sufficient.
+        
   FailureWhereEnum:
     description: The permitted values for describing where in the process, either a lab or analysis workflow step, the failure occurred.
     comments:


kheal:
@SamuelPurvine I think this is ready for review.  The aim here is to add easy-to-generate workflow summary stats to the metadata _and_ expand the `FailureWhatEnum` to capture how those stats would trigger a failure.  I gave a whack at the permissible values to add for `FailureWhatEnum`, but open to more or different ones.

turbomam:
Nice work guys. Are you still thinking about making some of the ranges `QuantityValue` with unit `%`? What do you like about that more than a purely numerical slot with good documentation, examples and possibly https://linkml.io/linkml-model/latest/docs/minimum_value/ and https://linkml.io/linkml-model/latest/docs/maximum_value/ constraints of 0/100 or 0.0/1.0?

turbomam:
Thanks @kheal, your responses have been helpful. I appreciate what you're trying to accomplish with self-describing data in MongoDB. Thanks for adding `is_a: ms_analysis_qc_metric` to all the new slots and cleaning up the failure categorizations.

I had Claude Code help me research the patterns across the schema. Here's what I found:

Precedent in the schema: There are about 180 `QuantityValue` ranges  on MIxS or NMDC native slots where the unit can vary across instances (temperature in C/F/K, concentration in mg/L vs ppm, etc.). `workflow_execution_activity.yaml` currently has 41 bare numeric slots and zero `QuantityValue`. The pattern has been: `QuantityValue` for variable-unit measurements, bare numeric for fixed-unit QC metrics.

Precedent in this PR: 5 of your 6 new slots here use bare integer or float. The count slots from #2825 and #2828 are also bare integers. `peptide_to_spectrum_match_rate` would be the first `QuantityValue` in any workflow execution class.

Enforceability: With `range: float` we can add `minimum_value: 0` and `maximum_value: 100` and LinkML validates that for free. LinkML can't enforce min/max constraints on `has_numeric_value` inside a `QuantityValue`. 

I don't think we need to settle the QV-vs-bare-numeric question globally right now. But for this slot specifically, I'd suggest staying consistent with the rest of the workflow QC metrics:

```yaml
peptide_to_spectrum_match_rate:
    is_a: ms_analysis_qc_metric
    range: float
    minimum_value: 0
    maximum_value: 100
    description: >-
        The percentage of peptide-to-spectrum matches relative to the total number of spectra analyzed.
```

and adjusting the example data files

Going forward I'd like all schema contributors to show the research they did into the existing patterns  as part of their PRs. Something like "I checked `workflow_execution_activity.yaml` and saw that all QC metrics use bare numerics, but I think QV is better here because X." That way we're making deliberate decisions about when to break
from precedent instead of discovering the inconsistency in (or after) review.

PR #2901, opened 2026-03-23
2901 Make slot `name` on class Biosample required

### LINKED ISSUE #2878: enforcement of unique sample names
As documented in https://github.com/microbiomedata/issues/issues/1639 we have instances of duplicate values for `name` & `samp_name` in prod mongo biosample_set records. samp_name is defined as a uniq_key on class Biosample but this isn't enforced.

DOD:
 - [ ] decide which slots need to be required  & unique. Both slots are currently optional.
 - [ ] make the nmdc-schema updates
 - [ ] decide on how to enforce this. Eric had a suggestion to use mongo composite keys, Patrick had suggested custom linkml-validation in the schema.
 - [ ] implement enforcement

@eecavanna @shreddd @mslarae13 @pkalita-lbl @kheal @turbomam 

Assigning to Montana to start.

aclum:
This PR makes name on slot Biosample required. nmdc-runtime will enforce this as unique within a study in the runtime code. This PR also includes a migrator to check that all documents in the biosample_set collection have a value for `name`.


turbomam (approved):

kheal (approved):
LGTM.  Would be nice if the PR name was updated so it translates to nice release notes.  I left one comment re: an acronym in a description that felt unnecessary.

mslarae13:
Sorry for the late reply. I think this is fine, but maybe we should consider deprecating samp_name and just using name with a mapping to samp_name for MIxS? cuz both is confusing.
there might already be an issue for tis?

PR #2928, opened 2026-04-02
Update insdc identifier modeling

### LINKED ISSUE #2960: review and update modeling for external nucleotide identifiers
This came up when reviewing isolate modeling. We would like to generically represent external identifiers that represent assembly or annotation identifiers from INSDC.
Example: how should nmdc model GOLD's equivalent of genbank 16S ID.

relates to: #2884

aclum:
This PR updates the modeling of insdc sequence identifiers, adds pattern constraints, and adds a mapping for gold slots.


turbomam (commented):
pretty good. I haven't decided about the slot name specificity. I  made a couple of syntax fixes and added some annotations.

aclum:
@turbomam DDBJ prefixes are extremely extensive so was just going to have a typecode for genbank and ena for now. thoughts? https://www.ddbj.nig.ac.jp/insdc/prefix-e.html#:~:text=AAAA%2DAZZZ%20GenBank%20general%20WGS,JAAAAA%2DJZZZZZ%20GenBank%20general%20WGS

aclum:
@cmungall this is ready for review based on our discussion this afternoon.

aclum:
@eecavanna where are we at in terms of specifying collections which should be validated? Otherwise we need a no-op migrator to make sure no instances of insdc_assembly_identifiers get added before the next release (odds of this are extremely low).

eecavanna:
> @eecavanna where are we at in terms of specifying collections which should be validated? Otherwise we need a no-op migrator to make sure no instances of insdc_assembly_identifiers get added before the next release (odds of this are extremely low).

There is still no official way to do it (i.e. no change since the topic came up in PR https://github.com/microbiomedata/nmdc-schema/pull/2901#issuecomment-4139993406 on March 26). I'll give it some attention this month, as I convert the "standard migration notebook" into a Typer script/CLI app. In the meantime, I'll add a migrator to this PR. The migrator will have a placeholder where someone can add the names of the relevant collections. If I manage to figure out what those collections are, I'll add their names myself. I expect to have added the migrator within the next hour.

nmdc_schema/migrators/migrator_from_11_18_0_to_11_19_0.py
```diff
@@ -0,0 +1,33 @@
...
+    def upgrade(self, commit_changes: bool = False) -> None:
+        r"""
+        Migrates the database from conforming to the original schema, to conforming to the new schema.
+        """
+
+        # List of collection names that we want the migration orchestration notebook to validate.
+        self._collection_names = [
+            "biosample_set",
+            "workflow_execution_set",
+        ]
```
eecavanna, line 21: Hi @aclum and @turbomam — here's where you can list the names of the collections that you want the downstream "migration notebook" to validate during the migration process.

aclum: thanks @eecavanna

src/schema/external_identifiers.yaml
```diff
@@ -370,15 +372,21 @@ slots:
       - value: mgnify.analysis:MGYA00002270
         description: combined analyses (taxonomic, functional) of sample ERS438107
 
-  ## assemblies
-
-  assembly_identifiers:
+  ## nucelotide and protein sequences
```
turbomam:
```suggestion
  ## nucleotide and protein sequences
```

src/schema/external_identifiers.yaml
```diff
@@ -370,15 +372,21 @@ slots:
...
+  sequence_identifiers:
     abstract: true
+    description: Slot which contains information on nucleotide or protein sequences. 
+    
 
-  insdc_assembly_identifiers:
-    is_a: assembly_identifiers
-    pattern: "^insdc.sra:[A-Z]+[0-9]+(\\.[0-9]+)?$"
+  insdc_sequence_identifiers:
+    description: An identifer for nucleotide or protein sequences from an INSDC instituion.
```
turbomam:
```suggestion
    description: An identifier for nucleotide or protein sequences from an INSDC institution.
```

src/schema/workflow_execution_activity.yaml
```diff
@@ -66,7 +66,13 @@ classes:
         structured_pattern:
           syntax: "{id_nmdc_prefix}:wfmgas-{id_shoulder}-{id_blade}{id_version}$"
           interpolated: true
-
+      insdc_sequence_identifiers:
+         structured_pattern:
+           syntax: "^genbank:(GCF|GCA)_\\d{9}\\.\\d+$|^ena.embl:(GCF|GCA)_\\d{9}\\.\\d+$"
+           interpolated: true
+         see_also: 
+           - https://www.ncbi.nlm.nih.gov/datasets/docs/v2/data-processing/policies-annotation/genome-processing/version-status/
```
turbomam: these are all indented one character too much

src/schema/workflow_execution_activity.yaml
```diff
@@ -115,6 +121,12 @@ classes:
         structured_pattern:
           syntax: "{id_nmdc_prefix}:wfmtas-{id_shoulder}-{id_blade}{id_version}$"
           interpolated: true
+      insdc_sequence_identifiers:
+         structured_pattern:
+           syntax: "^genbank:(GCF|GCA)_\\d{9}\\.\\d+$|^ena.embl:(GCF|GCA)_\\d{9}\\.\\d+$"
+           interpolated: true
+         see_also: 
+           - https://www.ncbi.nlm.nih.gov/datasets/docs/v2/data-processing/policies-annotation/genome-processing/version-stat
```
turbomam: not ideal indentation... i think this all has one space to many too

turbomam (commented):
Thanks @aclum

I have some ideas of what this PR would do for data retrievers in the ideal world:

- capture the identifier value as a CURIE, with defined prefixes and other appropriate pattern constraints. This PR does that well.
- capture the identifier's owner, preferably in the CURIE prefix. I think this PR does that well too.
- capture or otherwise make it very easy for the user to get information about the sequence type. I thought that would require adding a new class with slots for the identifier and the type. Is your position that the class containing the slot (and possibly the pattern constraints on the initial characters) make that unambiguous? Or that making it unambiguous with an enumeration isn't feasible because the values in the source data are too noisy?
    - do you want to allow, require or block versioned accessions like `genbank:GCF_000000000.0`? Right now that's unclear to me. If allow, then we should show examples of both.

I also want to understand if you think `insdc_sequence_identifiers` should inherit from `external_database_identifiers`. Right now its root is `sequence_identifiers`.

I have a new issue to clean up the multiple different patterns within `external_database_identifiers`, so I'd like this PR to either follow new best practices or stay disconnected from `external_database_identifiers`. If you want to stay disconnected, then I'd like to put your new slot in a different YAML file.
https://github.com/microbiomedata/nmdc-schema/issues/3020

I made a few comments about misspellings and incorrect indentation too. You can be even more strict and say that this PR adds some unnecessary new lines, some of which have white-space on them too. I can easily reformat that if that's OK with you.

turbomam:
If you want `insdc_sequence_identifiers` -> `sequence_identifiers` to be in the `external_database_identifiers` hierarchy, then we should assert that with a `is_a` on `sequence_identifiers`

If we do that, then we wouldn't make any `range` or `multivalued` asserts because they would be inherited as `external_identifier` -> `uriorcurie`

As it is right now,   no range is asserted for `sequence_identifiers` so  `string` is inferred from the `default_range`.  No `range` is asserted for `insdc_sequence_identifiers`  either so `external_identifier` is inferred from the re inferred from the `insdc_identifiers` mixin

`multivalued: true` is asserted for `insdc_sequence_identifiers` but I think it would be better to move that up to `sequence_identifiers`, or for `sequence_identifiers` to inherit it from `external_database_identifiers`

aclum:
@turbomam i believe I've addressed your PR comments. For accession number vs version number I think we should allow both to follow allowed patterns with bioregistry.

turbomam (approved):
Approving. The structural fix (`is_a: external_database_identifiers` on `sequence_identifiers`) addresses the hierarchy concern in my earlier review — `insdc_sequence_identifiers` now follows the matrix shape (domain base + authority mixin) and inherits `range: external_identifier` and `multivalued: true` cleanly. I confirmed by running `linkml-validate` against the materialized schema that the per-class `slot_usage` patterns work as intended (16S accessions accepted on `OrganismSample`, GCF/GCA accepted on the assembly classes, mismatches rejected). Production MongoDB has zero documents using the old `insdc_assembly_identifiers` field, so the no-op migrator is safe.

eecavanna:
> Production MongoDB has zero documents using the old `insdc_assembly_identifiers` field, so the no-op migrator is safe.

In the context of this "(A) so (B)" statement; (A) does not, to me, imply (B).

Those kinds of documents could be inserted into the production database between now and when the no-op migrator is run. In addition, other instances of "the NMDC database" (other than our production one) that conform to the old schema may use the migrator to migrate the database to conform to the new schema.

The reason I consider the no-op migrator to be safe is that it identifies collections that the person running the migrator will dump and validate during the migration process. It is **not** that our production database happens to not contain such documents today.

PR #3019, opened 2026-04-27
Add new chemical entity permissible values for material processing metadata

bmeluch:
This PR adds five permissible values to `ChemicalEntityEnum` to accommodate material processing metadata for a new study/new protocols.

### DIFF
--- a/src/schema/core.yaml
+++ b/src/schema/core.yaml
@@ -157,6 +157,8 @@ enums:
         meaning: CHEBI:62947
       ammonium_bicarbonate:
         meaning: CHEBI:184335
+      ammonium_formate:
+        meaning: CHEBI:63050
       amitriptyline:
         meaning: CHEBI:2666
       Arg-C:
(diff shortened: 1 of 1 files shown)


src/schema/core.yaml
```diff
@@ -215,8 +219,14 @@ enums:
         meaning: CHEBI:43945
         comments:
           - A 1,1-bis(phosphonic acid) consisting of methane substituted by two phosphonic acid groups, also known as methylenediphosphonic acid
+      potassium_chloride:
+        meaning: CHEBI:32588
+      phosphate_buffered_saline:
```
kheal: `phosphate_buffered_saline` is a solution, not a compound, so I don't think it should be in this Enum.  We try to only have CHEBI: and MS: ontologies in this Enum.

Certainly annoying, but can we add its components to this Enum and model it like a solution (similar to what we do for mobile phase solutions for chromatography)?

Otherwise these look good to me.

bmeluch: We can, I don't have the exact recipe for the buffer so it'll be approximate

kheal (approved):

PR #3081, opened 2026-05-18
Define `ProcessedSample` structurally as `MaterialProcessing` output (#3051)

### LINKED ISSUE #3051: Define ProcessedSample
Add a definition for ProcessedSample.
Team members have indicating it should be defined as a sample in which the biomolecules are no longer subject to change as determined by cell lysis.

turbomam:
Closes #3051.

## What this adds

A `description` on the `ProcessedSample` class in `core.yaml`:

> "A Sample produced from another Sample through a MaterialProcessing event recorded in nmdc-schema."

The class previously had no description.

## Why this framing

The definition is structural: a ProcessedSample is whatever a MaterialProcessing event produces. It is verifiable from the data graph (does a producing MaterialProcessing event exist for this sample?) and does not conflict with any existing MaterialProcessing subclass output range.

The framing originally proposed in #3051 ("a sample in which the biomolecules are no longer subject to change as determined by cell lysis") was considered and rejected because it would invalidate roughly 7,800 of 19,572 existing prod ProcessedSamples — the outputs of SubSamplingProcess, Pooling, FiltrationProcess, DissolvingProcess, and ChemicalConversionProcess, none of which entail cell lysis.

Full rationale, alternative definitions considered, and the prod-data evidence: [NMDC ProcessedSample definition — design conversation (2026-05-19)](https://docs.google.com/document/d/10CoL3zLgunLpRTsw8GRJv0C6Cc_OEtPZpQTZ8KImrts/edit?usp=drivesdk)

## Generated files

Per `CLAUDE.md`, regenerated `nmdc_schema/` artifacts are not committed in development PRs; they will be updated immediately before merge or release.


aclum (changes requested):
We discussed in the metadata meeting that this would make a lot of existing samples invalid and my understanding is we still wanted to proceed  &  include a migrator to fix those samples. If this definition is enforced it will invalidate the valid example data that is currently in the schema, specifically Culturing is a subclass of MaterialProcessing and we currently have https://github.com/microbiomedata/nmdc-schema/blob/main/src/data/valid/Database-isolate-from-culture-collection.yaml.

If you don't have one already a claude skill to check for consistently with valid and invalid data would be useful. On a previously PR claude hallucinated a  definition of ProcessedSample which did not hold up against example data or data in mongo.

turbomam:
from @mslarae13 : distinguish between samples that we processed in some way that was required to ask a scientific question vs processing to make it amenable to some technology

turbomam:
After being processed, an output ProcessedSample could still be subjected to scientific question treatments

turbomam:
What has more of an impact on our choices of what classes to instantiate:
- hierarchy
- class name
- associated slots

we never want to repeat values in chained instances of the same class

turbomam:
From @kheal: is the defined by which sample instances could or would have a phenotype associated with them?

turbomam:
be very careful about defining classes based on what some person would intend do with the instances

turbomam:
don't forget about the idea to change the name of `Biosample` to `EnvironmentalSample`... not that it would necessarily solve the problems above

kheal:
> From @kheal: is the defined by which sample instances could or would have a phenotype associated with them?

this was more about 
- #3080

turbomam:
What limits do we have on the number of classes, especially in terms of operating with a data lakehouse?

turbomam:
could make `Sample` non-abstract

src/schema/core.yaml
```diff
@@ -1271,6 +1271,9 @@ classes:
     class_uri: nmdc:ProcessedSample
     is_a: Sample
     title: Processed Sample
+    description: >-
+      A Sample produced from another Sample through a MaterialProcessing event
+      recorded in nmdc-schema.
```
turbomam, line 1276: soften the requirement that the MaterialProcessing has been recorded in our database?

PR #3084, opened 2026-05-19
3017 update analysis types to support isolate modeling

### LINKED ISSUE #3017: Add isolate WGS permissible value to AnalysisTypeEnum
`AnalysisTypeEnum` currently covers metagenomics, metatranscriptomics, metabolomics, etc. — none of which apply to isolate whole-genome sequencing projects.

`submission-schema` added `IsolateInterface` and `JgiIsolateInterface` (submission-schema#418) but had to override `analysis_type` to `required: false` because no valid value exists.

Add a permissible value such as `isolate_genome_sequencing` (or agreed-upon equivalent) so that isolate submissions can record their analysis type, the `required: false` workaround can be removed, and `IsolateInterface` records are queryable by data type alongside other submission types.

aclum:
This PR updates AnalysisTypeEnum, adds analysis_type as a slot on class Organism sample and updates example data.


turbomam:
@copilot normalize the whitespace in and around any newly committed lines in this branch

src/schema/portal_enums.yaml
```diff
@@ -24,20 +24,34 @@ enums:
...
+        description: Short-read metatranscriptomic sequencing of RNA from a mixed community.
       natural organic matter:
+        description: Analysis of natural organic matter (NOM) from a mixed community, such as by Fourier-transform ion cyclotron resonance mass spectrometry (FTICR-MS).
       bulk chemistry:
+        description: Analysis of bulk chemical properties of a sample, such as pH, conductivity, total carbon, total nitrogen, etc.
       amplicon sequencing assay:
         meaning: OBI:0002767
         title: Amplicon sequencing assay
+      isolate genome sequencing:
+        meaning: OBI:0002117
```
turbomam: OBI:0002117 'whole genome sequencing assay' is broader than "isolate genome sequencing" so please use `broad_mappings` not `meaning`

turbomam (changes requested):
one meaning needs to be changed to a broad mapping. otherwise very nice and ready to go.

src/schema/portal_enums.yaml
aclum: I thought about that but this is the definition "A DNA sequencing assay that intends to provide information about the sequence of an entire genome of an **organism**", so meaning seemed more appropriate. @turbomam LMK if you still think this should switched to broad_mappings.

src/schema/portal_enums.yaml
```diff
@@ -24,20 +24,34 @@ enums:
...
         title: Metagenomics
       metagenomics_long_read:
         description: Long-read metagenomic sequencing
         title: Metagenomics (long read)
       metaproteomics:
+        description: Mass spectrometry-based analysis of proteins from a mixed community.
       metatranscriptomics:
+        description: Short-read metatranscriptomic sequencing of RNA from a mixed community.
       natural organic matter:
+        description: Analysis of natural organic matter (NOM) from a mixed community, such as by Fourier-transform ion cyclotron resonance mass spectrometry (FTICR-MS).
```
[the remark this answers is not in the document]
aclum: resolved by copilot commit.

mslarae13 (approved):

src/schema/portal_enums.yaml
```diff
@@ -24,20 +24,34 @@ enums:
   AnalysisTypeEnum:
     permissible_values:
       metabolomics:
+        description: Mass spectrometry-based analysis of metabolites from a mixed community.
```
kheal: We would use this same term for community or isolates.

src/schema/portal_enums.yaml
mslarae13: discussed 2026-05 isolate and organism aren't the same thing. OBI doesn't appear to distinguish between the 2. So we change this to broad_mappings and if Mark gets updates from OBI we can re-visit.

src/schema/portal_enums.yaml
aclum: definitions update for metabolomics, lipidomics, and NOM.

turbomam (approved):

PR #3094, opened 2026-05-21
Change gc_content's range to `float` and add minimum and maximum values

### LINKED ISSUE #3095: Enforce a range on gc_content
This slot should be constrained to 0-100%

aclum:
This PR updates the range of gc_content from a QuantityValue to a float so we can enforce max and min values.


aclum:
@turbomam how do you want to enforce range for GC content? It should be 0-100% so this build should fail w/my changes to the example file. Because of the class redirect AI is telling me we either need a linkml rule or a special subclass. Or I suppose we could make this a float.

cc @mslarae13

turbomam:
We add can add minimum and maximum values

https://linkml.io/linkml-model/latest/docs/minimum_value/

turbomam:
@copilot give NMDC's gc_content slot a [minimum_value/](https://linkml.io/linkml-model/latest/docs/minimum_value/) of 0 and maximum value of 100

turbomam:
Duh I rushed past the pact that it's a `QuantityValue` slot. I don't have a quick solution for this. I think all percentage slots should have integer or float ranged with minimum and maximum values but that will require migration**S**

mslarae13:
> Duh I rushed past the pact that it's a `QuantityValue` slot. I don't have a quick solution for this. I think all percentage slots should have integer or float ranged with minimum and maximum values but that will require migration**S**

@turbomam I think we have other % slots where we've implemented the standard units that we could reference.

aclum:
This slot hasn't been used yet so can we just change it to a float?

aclum:
@turbomam what do you want to do here?

turbomam:
Going with @aclum's suggestion: switch `gc_content` to `range: float` with `minimum_value: 0` and `maximum_value: 100`. Since this is now a scalar slot, use LinkML's `unit` metaslot for the percentage:

```yaml
gc_content:
  range: float
  minimum_value: 0
  maximum_value: 100
  unit:
    ucum_code: "%"
```

Background:
- `gc_content` isn't in any released schema version (only on main and `v11.19.2-rc.1`); the runtime API currently serves `11.19.1`.
- 0 `gc_content` documents in my local Mongo snapshot.

So nothing should have been written with the old `QuantityValue` shape. To keep the "migrator for any shape change" discipline, the migrator just counts `gc_content` documents and fails if it finds any. Worth running against prod and dev anyway.

Follow-ups, separate issues:
- Migrate the other 8 percent slots from `QuantityValue` + `storage_units` annotation to scalar + `unit` metaslot.
- Adopt LinkML's `unit` metaslot more broadly on existing type-ranged slots that carry a unit.
- LinkML feature request so a parent slot's `minimum_value`/`maximum_value` can constrain the inner `has_numeric_value` without writing a rule.

turbomam (approved):

turbomam:
@aclum what do you think of this PR now?

aclum:
LGTM, someone else needs to approve this because I authored this PR.

turbomam:
No migrator needed. `gc_content` and the `organism_set` collection exist only on `main`; neither is in `nmdc-schema` 11.19.1, which is what `nmdc-runtime` pins. So the API-bound schema has never had this slot, there's no `organism_set` data in prod to strand, and `gc_content` ships as `float` from its first release. (Confirms that the slot hasn't been used yet.)

turbomam:
@aclum if we can't get anybody else to approve, I make a motion that our mutual approval is good enough

aclum:
Agreed

PR #3106, opened 2026-05-26
Add sequel_IIe to InstrumentModelEnum

sujaypatil96:
## Summary
- Adds `sequel_IIe` permissible value to `InstrumentModelEnum` in `src/schema/basic_classes.yaml`, with alias `PacBio Sequel IIe`
- PacBio Sequel IIe is a distinct instrument from the original Sequel and Sequel II — it shares Sequel II's chemistry/optics but adds on-instrument compute for HiFi (CCS) read generation
- No OBI term currently exists for Sequel IIe, so the entry omits `meaning:` — consistent with the existing `revio` entry

## Test plan
- [x] `make all` builds cleanly
- [x] `make test` passes
- [ ] New value appears in generated artifacts (JSON Schema, OWL)

### DIFF
--- a/src/schema/basic_classes.yaml
+++ b/src/schema/basic_classes.yaml
@@ -1278,7 +1278,11 @@ enums:
         aliases:
           - PacBio Sequel II
         meaning: OBI:0002633
-      
+
+      sequel_IIe:
+        aliases:
+          - PacBio Sequel IIe
+
       revio:
         aliases:
           - PacBio Revio


sujaypatil96:
Background: there is a study that we are trying to ingest from NCBI (https://www.ncbi.nlm.nih.gov/bioproject/?term=PRJNA1071982), the SRA records for which use the following `instrument_model` values – "Illumina NovaSeq 6000", "Sequel IIe", "PromethION".

Currently, we have `sequel_II` in the `InstrumentModelEnum` but not `sequel_IIe` which needs to be added.

For example, https://www.ncbi.nlm.nih.gov/biosample/SAMN39873784 is linked to this SRA record: https://www.ncbi.nlm.nih.gov/sra/SRX23628290

turbomam:
Follow-up issues filed for the broader mapping gaps surfaced while reviewing this PR:

- #3107 — Add OBI `meaning:` CURIEs to the three unmapped umbrella sequencer PVs (`novaseq`, `hiseq`, `nextseq`) that already have OBI terms
- #3108 — Promote `exact_mappings` to `meaning:` across all enum PVs where the mapping is 1:1

OBI NTR for Sequel IIe is drafted and ready to file at obi-ontology/obi (modeled on obi-ontology/obi#849). PacBio Revio also has no OBI term.

turbomam:
OBI NTRs filed for both missing PacBio sequencers:
- obi-ontology/obi#1984 — Sequel IIe
- obi-ontology/obi#1985 — Revio

turbomam:
OBI series grouping NTRs filed to bring PacBio and ONT in line with the existing Illumina series pattern:
- obi-ontology/obi#1986 — PacBio SMRT sequencer series instrument
- obi-ontology/obi#1987 — Oxford Nanopore sequencer series instrument

turbomam (approved):

src/schema/basic_classes.yaml
```diff
@@ -1278,7 +1278,11 @@ enums:
         aliases:
           - PacBio Sequel II
         meaning: OBI:0002633
-      
+
+      sequel_IIe:
```
aclum, line 1282: The only difference between these is if data processing is on instrument or not so sequel_IIe should be an alias of sequel_II IMO

PR #3121, opened 2026-05-28
Assert PV `is_a` hierarchies in `InstrumentModelEnum`

### LINKED ISSUE #3120: assert that Sequel IIe seqeuncer PV `is_a` Sequel II
also
- apply that pattern to existing backlog of PVs were appropriate
- communicate this new pattern to data portal developers etc

turbomam:
Closes #3120.

## Why

Should `sequel_IIe` be folded into `sequel_II` as an **alias**? No. `aliases` assert that two values are *equivalent*, but NCBI's controlled vocabulary treats "Sequel IIe" and "Sequel II" as **separate** values. An alias would assert a false equivalence and lose the distinction NCBI needs to forward a user-supplied "Sequel IIe" faithfully.

`is_a` is the right tool: keep `sequel_IIe` as its own PV (so it forwards to NCBI without lossy normalization) while recording that it is a *specialization* of `sequel_II`, so NMDC search and rollups can still treat them together. This PR applies the same idea to the Illumina families: relate a specific model to its series without collapsing them into one value.

## What

PV-level `is_a` added in `InstrumentModelEnum`:

- `sequel_IIe` → `sequel_II` (the literal ask)
- `novaseq_6000`, `novaseq_x` → `novaseq`
- `hiseq_1000/1500/2000/2500/3000/4000`, `hiseq_x_ten` → `hiseq`
- `nextseq_500/550/1000` → `nextseq`

Applied where the value's OBI `meaning:` supports it: OBI groups the numbered Illumina models under a `... series instrument` class (e.g. `Illumina HiSeq 2500` under `Illumina HiSeq series instrument`), so those get `is_a`. **Skipped** where OBI treats values as siblings (`PacBio Sequel II` is not under `PacBio Sequel`) and peer models with no shared parent (Orbitrap Exploris 120/240/480, Solarix 7T/12T/15T, the Agilent GC-MS pair). The one editorial link OBI can't adjudicate is `sequel_IIe is_a sequel_II` (no Sequel IIe term in OBI), kept as a hardware refresh. Decision rule documented in `CONTRIBUTING.md`. Follows existing precedent in `StationaryPhaseEnum` and `SamplePortionEnum`.

## Impact

- **No validation change, no data migration.** Enums compile to a flat string list in JSON Schema and `str` enums in Pydantic; PV `is_a` is metadata only. Verified: generated `InstrumentModelEnum` is an unchanged flat 41-value list. Asserting `sequel_IIe is_a sequel_II` does **not** make `sequel_II` accepted where `sequel_IIe` is.
- **Downstream (data portal / nmdc-runtime / nmdc-server):** non-breaking. The hierarchy is newly *available* for dropdown grouping and rollup queries; nothing requires action. Communicating this availability to portal developers is tracked in #3120.
- **OWL:** deprioritized. `linkml generate owl --enum-inherits-as-subclass-of` would surface this as `subClassOf`, but that flag is intentionally **not** enabled. Documented in `CLAUDE.md`.

## Notes

Verified with `linkml generate json-schema src/schema/nmdc.yaml` (clean build, `is_a` targets resolve; generated `InstrumentModelEnum` is an unchanged flat 41-value list). Full `make all` not run locally yet. No generated `nmdc_schema/` artifacts committed (per `CLAUDE.md`, those land at release time).


turbomam:
we need to make sure that people maintaining other microbiomedata repos that could take advantage of this, like nmdc-server, know about this

sujaypatil96 (approved):
This is a great solution to organize PV's and also make sure we are retaining specificity on instrument values in upstream sources like NCBI.

<s>I also love the comment on "sequel_IIe" noting the distinction b/w it and "sequel_II" and why we chose to add that as a separate PV.</s>

Discussion/policy on whether or not to propagate every documentation/schema change decisions in metamodel `comments` block.

src/schema/basic_classes.yaml
```diff
@@ -1283,8 +1295,21 @@ enums:
         meaning: OBI:0002633
 
       sequel_IIe:
+        is_a: sequel_II
         aliases:
           - PacBio Sequel IIe
+        comments:
```
aclum: I don't think we need to store every AI modeling  justification within the schema itself. Can this be abstracted into a general schema skill about modeling instead or be part of the PR comment rather than detailed  comments within the schema?

aclum (approved):

eecavanna:
> Applied only where one value is a true specialization (a variant/refresh, or a numbered model under a generic family-head PV).

This seems a little perilous to me.

I think marketers can switch model names and numbers for any reason they want (e.g. supply chain changes; desire to convey freshness to the market).

Also, sometimes the first version of something ends up having the product line be named after it. Example: Sony named the first `PlayStation` "PlayStation". They then came out with a successor called "PlayStation 2". In my opinion, a "PlayStation 2" is a `PlayStation` (i.e. a member of that product line), but is not a "PlayStation" (i.e. arguably not a variant of the original one). The first iPhone, iPod, XBox, etc. are in a similar situation.

I suggest pushing back on Claude to document more specific criteria. I worry about someone on our team making (and having to defend) this determination about someone's product line in the future.

Maybe there are few enough products involved here that it is, indeed, feasible for someone on our team to determine this for every new PV. Also, maybe the hierarchy is obvious to people that use these instruments (I am not among them).

src/schema/basic_classes.yaml
turbomam: thanks @aclum . when you post comments like these, would you please include the part that you're most concerned about about?

turbomam: Kept as a permissible value distinct from sequel_II rather than
            folded into it as an alias. The Sequel IIe differs from the Sequel II
            essentially only in onboard electronics, so merging them would improve
            the NMDC search experience (a IIe query would match II records).
            However, NCBI treats "Sequel II" and "Sequel IIe" as separate
            controlled-vocabulary values, so we retain the distinction here in
            order to forward user-supplied "Sequel IIe" through to NCBI records
            without lossy normalization. The is_a assertion records that the IIe
            is a specialization of the II so that NMDC search/rollup can still
            treat them together.

turbomam: I'm going to take your suggestion of reflecting it in the PR comments and therefor remove the comment alltogheter.

turbomam:
Thanks @eecavanna, that helped improve the PR

I won't pretend we have a perfect algorithm for this, but part of this modeling is acknowledging and using the fact that entities can be related to one another.

The criterion is no longer left solely to judgment. It's written down in `CONTRIBUTING.md` (Modeling Best Practice), and mirrored for agents in `CLAUDE.md`. We can revise those over time.

>assert PV `is_a` only for a real specialization (a model within a series, or a refresh of a model), judged on domain knowledge with the ontology `meaning:` (OBI) as corroborating evidence, not the sole authority; when you can't establish the relationship, leave it flat.`

Your "a series is contingent" point is in the rule explicitly: 

> a "series" is a curatorial grouping (e.g. an OBI `... series instrument` class), not a vendor-guaranteed category, so we corroborate against the ontology, not vendor naming. 

It also captures the discussions between @aclum, @sujaypatil96 and me that started this: 

>`aliases` assert equivalence, `is_a` records specialization. `sequel_IIe` stays a distinct PV with `is_a: sequel_II` because NCBI treats "Sequel IIe" and "Sequel II" as separate values, so an alias would be a false equivalence.

I kept this short rather than write a formal criteria essay. The larger pieces are tracked separately: 
- microbiomedata/nmdc-schema#3124
- microbiomedata/issues#1745
- microbiomedata/nmdc-schema#3123

PR #3127, opened 2026-06-03
Add `maximum_value` to `estimated_size` (closes #3122)

### LINKED ISSUE #3122: `estimated_size`: align bounds and unit with JGI esplims
Surfaced 2026-05-28 reviewing JGI esplims (`pps/esp/esplims` on `code.jgi.doe.gov`, `extensions/server/submission_runtimes.py:516-589`, `check_genome_size`).

## Current nmdc-schema state

`estimated_size` (mixs.yaml): `range: integer`, `minimum_value: 1`, no `maximum_value`. Last touched in #3091 (integer + drop bp-suffix machinery). Unit: bp (implicit; comment says "base pair scale/units"; no UCUM).

## What JGI esplims enforces

Category-conditional, captured in **Mb** (confirmed by `aga_utils.py:529` `tax_genome_size_est_mb` and `reads_requested.py:69` `'Est. Genome Size (Mb)'`):

| Biosafety Material Category | Max |
|---|---|
| Alga, Animal, Plant, Protist | 100,000 Mb (= 1e11 bp) |
| Archaea, Bacteria, Fungi, Plasmid, Virus, Synthetic Construct | **5,000 Mb** (= 5e9 bp) |

Also required-conditional: when one of the above categories is set and `estimated_size` is empty, esplims rejects.

## Questions

1. **Unit**: keep bp (current), switch to Mb (matches esplims and what submitters fill on the JGI form), or carry a unit annotation and let the portal convert?

2. **Max**: a single loose 1e11 bp ceiling lets oversized microbe submissions pass NMDC and gate-fail at JGI later; a category-aware rule (5,000 microbe / 100,000 eukaryote, in Mb) mirrors esplims and catches the error earlier. `JgiIsolateInterface` in submission-schema already imports `biosafety_mat_cat`, so a class-level LinkML rule keyed on it is feasible.

3. **Min**: esplims has no explicit lower bound (just must-be-specified-when-conditional). Current `minimum_value: 1`. Keep, or drop to 0?

## Refs

- submission-schema #422 (wire isolate biology slots), #423 (interfaces), #429 (requiredness audit, row for JGI #12)
- @aclum raised the bounds question 2026-05-21 in submission-schema PR #423

cc @aclum

turbomam:
Adds `maximum_value: 100000000000` (1e11 bp = 100,000 Mb) to the `estimated_size` slot in `mixs.yaml`.

This is a loose sanity ceiling matching the JGI esplims upper bound for eukaryotes. Tighter per-category limits (e.g., 5,000 Mb for microbes) are enforced in submission-schema via class rules, not here.

### DIFF
--- a/src/schema/mixs.yaml
+++ b/src/schema/mixs.yaml
@@ -7115,6 +7115,7 @@ slots:
     maximum_value: 100000000000
     comments:
       - JGI reports values in megabases (Mb); NMDC stores them in base pairs (bp).
+      - The maximum_value here (1e11 bp = 100,000 Mb) is the eukaryote ceiling from JGI esplims. The tighter microbe ceiling (5,000 Mb) and the category-conditional required rule are enforced in submission-schema via class rules on IsolateInterface, keyed on biosafety_mat_cat.
     examples:
       - object: 300000
   ploidy:
--- a/src/schema/mixs.yaml
+++ b/src/schema/mixs.yaml
@@ -7112,6 +7112,7 @@ slots:
     slot_uri: MIXS:0000024
     range: integer
     minimum_value: 1
+    maximum_value: 100000000000
     comments:
       - JGI reports values in megabases (Mb); NMDC stores them in base pairs (bp).
     examples:


turbomam:
i think we may want a minimum size like 200 too

turbomam:
we discussed retaining the bp units in nmdc-schema/Mb in submission-schema but it's not to late to change that plan

I'm not 100% convinces that any submission-schema or nmdc-server code looks at the `unit` metaslot that we are using here (in contrast to the units annotation). But I should double check.

Issue #2970, opened 2026-04-10
Remove remaining RDF generation and triplestore remnants


turbomam:
## Context

NMDC no longer maintains an RDF representation of instance data or a triplestore. PR #2929 removed Apache Jena and Fuseki from the Dockerfile and docker-compose.yml. OWL generation of the schema itself is kept (`--include owl`). But several RDF remnants remain.

### Makefile build changes

- **Drop `--include rdf`** from both `linkml generate project` invocations (Makefile:103 and Makefile:127). This eliminates `project/rdf/` and all `examples/output/*.ttl` files.
- **Remove `rdf` from `PROJECT_FOLDERS`** (Makefile:244) — currently `jsonldcontext jsonschema owl python rdf`, should become `jsonldcontext jsonschema owl python`.

### Dead Makefile targets (project.Makefile)

- `local/nmdc_materialized.ttl` (~line 247) — converts schema to RDF Turtle, no downstream consumer
- `local/nmdc-no-use-native-uris.owl.ttl` (~line 243) — OWL variant without native URIs, no downstream consumer
- `local/mongo_as_nmdc_database_rdf_safe.yaml` (~line 109) — "RDF safe" migration step, part of the dead dump chain (#2969)

### Files to delete

- `project/rdf/` directory — generated RDF context, will stop being generated
- `examples/output/*.ttl` (~120 files) — RDF serialization of every example, nobody consumes these
- `assets/my_emsl_prefix.ttl` — EMSL prefix declarations in Turtle, no references in code

### .gitignore cleanup

- `tdbcontent` and `tdbcontent/` — Jena TDB2 data directory entries, nothing creates these anymore

### What stays

- `--include owl` — OWL generation of the schema is still wanted
- `rdflib` dependency in pyproject.toml — used by generated `nmdc.py` (linkml-runtime requirement)
- `--include jsonldcontext` — JSON-LD context generation

## Related

- #2929 — Removed Jena/Fuseki from Dockerfile and docker-compose.yml
- #2969 — Replace dump-validate-migrate chain (covers the `rdf_safe` target)
- #2749 — Review and clean up RDF/OWL generation in build process
- #2781 — Remove `--include rdf` from gen-project command (this issue supersedes it)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

turbomam:
[...]

PR #3132, opened 2026-06-05
Remove SPARQL query files and unreferenced OWL customization artifact

turbomam:
Removes 11 `.rq` files from `assets/sparql/` and `assets/my_emsl_prefix.ttl`.

The SPARQL files were written for a triplestore workflow (Fuseki, TDB) that is no longer part of the NMDC stack. `assets/my_emsl_prefix.ttl` is an orphaned custom prefix file with no references in any Makefile, script, or schema file.

OWL generation targets (`--include owl` in `gen-project`) and OWL generation monitoring issues are not affected.

Partial implementation of https://github.com/microbiomedata/nmdc-schema/issues/2970


turbomam (commented):
Fixed in the follow-up commit: removed the 'Approach A: RDF/Triplestore Analysis' section from `units/docs/README.md`, including the embedded SPARQL query and the `assets/sparql/mongodb-slots-to-units.rq` file path reference. Also updated the `mongodb-slots-to-units.csv` status note to say the triplestore workflow is retired rather than claiming the file is regenerable via SPARQL.

turbomam:
Follow-up commit 025eb6828 addresses the Copilot finding: removed the 'Approach A: RDF/Triplestore Analysis' section from `units/docs/README.md`, including the embedded SPARQL query text and the `assets/sparql/mongodb-slots-to-units.rq` file path reference. Also updated the `mongodb-slots-to-units.csv` status line to say the triplestore workflow is retired.

PR #3174, opened 2026-06-10
Implement function that returns mapping from class name to typecodes

### LINKED ISSUE #3079: Make external function for grabbing all valid NMDC typecode values
As referenced in this issue in nmdc-server, https://github.com/microbiomedata/nmdc-server/issues/2134, we need a way to use nmdc-schema as the source of truth for what a valid NMDC typecode is. We currently do a version of this [here in the `make_typecode_to_class_map.py` script](https://github.com/microbiomedata/nmdc-schema/blob/adf5c5405c6773be488ef02a96fdc76ace309fa9/src/scripts/make_typecode_to_class_map.py#L103) for generating [this page in the schema docs](https://microbiomedata.github.io/nmdc-schema/typecode-to-class-map/).

Right now, we have a hardcoded enum in nmdc-server that is based on the above docs page. It would be good if instead we could have a function that we import from nmdc-schema that can pull these values straight from the schema as is done for that docs page.

eecavanna:
On this branch, I implemented a function in `nmdc_schema/id_helpers.py` named `get_class_name_to_typecode_map`, which people (e.g. developers of downstream applications) can use to get a mapping from schema class name to the typecode(s) compatible with the `id` slot, if any, of that schema class.

A note regarding maintenance: The doctests currently reference specific classes that exist in the schema. When I had an option, I tried to choose classes I thought were unlikely to be removed from the schema.

Fixes #3079

### DIFF
--- a/nmdc_schema/id_helpers.py
+++ b/nmdc_schema/id_helpers.py
@@ -3,11 +3,13 @@ This module contains helper functions related to schema element identifiers.
 """
 
 import re
+from functools import lru_cache
 from typing import Dict, List, Optional
 
 from nmdc_schema.get_nmdc_view import ViewGetter
 
 
+@lru_cache
 def get_compatible_typecodes(slot_pattern: str) -> List[str]:
     r"""
     Returns the typecodes, if any, that the schema says values of the specified `id` field can contain.
@@ -61,6 +63,7 @@ def get_compatible_typecodes(slot_pattern: str) -> List[str]:
     return typecodes
 
 
+@lru_cache
 def get_typecode_for_future_ids(slot_pattern: str) -> Optional[str]:
     r"""
     Returns the typecode, if any, that schema authors want future values of the specified `id` field to contain.
@@ -80,6 +83,7 @@ def get_typecode_for_future_ids(slot_pattern: str) -> Optional[str]:
     return compatible_typecodes[0] if len(compatible_typecodes) > 0 else None
 
 
+@lru_cache
 def get_class_name_to_typecode_map() -> Dict[str, List[str]]:
     r"""
     Returns a dictionary in which each key is the name of a schema class and each value is a list of all the typecodes
--- a/nmdc_schema/id_helpers.py
+++ b/nmdc_schema/id_helpers.py
@@ -92,15 +92,15 @@ def get_class_name_to_typecode_map() -> dict[str, list[str]]:
     True
     >>> all(isinstance(v, list) for v in d.values())  # all dict values are lists
     True
-    >>> "NamedThing" in d  # schema class lacking an `id` field compatible with any typecodes
+    >>> "NamedThing" in d  # schema class lacking an `id` slot compatible with any typecodes
     True
     >>> d["NamedThing"]
     []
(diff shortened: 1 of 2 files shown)


eecavanna:
@codex review

nmdc_schema/id_helpers.py
```diff
@@ -76,3 +78,56 @@ def get_typecode_for_future_ids(slot_pattern: str) -> Optional[str]:
...
+    >>> d["Biosample"]
+    ['bsm']
+    >>> "MassSpectrometry" in d  # schema class having an `id` slot compatible with multiple typecodes
+    True
+    >>> sorted(d["MassSpectrometry"])
+    ['dgms', 'omprc']
+    """
+
+    # Make a dictionary in which each key is the name of a schema class and each value is the definition of that class.
+    view_getter = ViewGetter()
```
turbomam, line 114: do you think this would be run frequently enough to justify something like a `functools.lru_cache` ?

src/scripts/make_typecode_to_class_map.py
```diff
@@ -96,6 +96,9 @@ def get_collection_names_for_class_name(class_name: str) -> List[str]:
 def main():
     r"""
     Generates a Markdown document containing a table that can be used to map typecodes to schema classes.
+
+    TODO: Refactor this function to use the newly-implemented `get_class_name_to_typecode_map` function
```
turbomam, line 100: what do you think of making an issue for that TODO?

turbomam (approved):
I think you may have a mixture of PEP 585 builtins for typing (dict[str, list[str]]) and typing.List/Optional

nmdc_schema/id_helpers.py
eecavanna: Good idea. I don't think so, but I see no harm in adding that. I'll add it now.

src/scripts/make_typecode_to_class_map.py
eecavanna: Yes, I'll create one.

eecavanna:
> I think you may have a mixture of PEP 585 builtins for typing (dict[str, list[str]]) and typing.List/Optional

I'll address this now.

---

Update: Done via commit https://github.com/microbiomedata/nmdc-schema/pull/3174/commits/d5e5281b2bcd3d85a8e2b32139d3771fe44353b7.

codytodonnell (approved):

nmdc_schema/id_helpers.py
eecavanna: Done via commit https://github.com/microbiomedata/nmdc-schema/pull/3174/commits/6263aba9dd34f205bafcc738446436b5792fac82. Resolving conversation.

src/scripts/make_typecode_to_class_map.py
eecavanna: Issue created at: https://github.com/microbiomedata/nmdc-schema/issues/3178. Resolving conversation.

eecavanna:
Thanks for reviewing this and giving me the time to implement it. I'm ready for it to be merged in.

PR #3187, opened 2026-06-16
Add  INSDC run identifiers slot

### LINKED ISSUE #3186: Add insdc run identifiers
We support a number of insdc identifiers but don't currently have a slot for sra run identifiers

aclum:
This PR adds support for INSDC run identifiers and adds the slot to class DataObject.


src/schema/external_identifiers.yaml
```diff
@@ -321,11 +321,23 @@ slots:
...
+    description: INSDC identifiers for the unique sequencing result for a specific sample
+    see_also:
+      - https://www.ncbi.nlm.nih.gov/sra/docs/submitmeta/#sra-metadata-experiment
     is_a: external_database_identifiers
     pattern: "^insdc.sra:(E|D|S)RX[0-9]{6,}$"
     mixins:
       - insdc_identifiers
 
+  insdc_run_identifiers:
+    description: Identifiers for the manifest of data file(s) that are derived from sequencing a library described by the associated INSDC experiment.
```
jasmynp, line 333: Is this description supposed to be for the insdc_experiment_identifers? Swap the two descriptions?

jasmynp (commented):
might need a fix under external_identifiers.yaml

src/schema/external_identifiers.yaml
aclum: no, this is for the run, the definition is basically verbatim from NCBI's documentation. See the link via see_also

jasmynp: ok i see that now. My confusion came from our earlier discussion regarding which experiment being synonymous with our "manifest", so the description of the run using the same terminology confused me.

jasmynp (commented):
After Alicia looks at my latest reply to her comment regarding the descriptions, looks good to me, but I'll defer to Mark or Sujay to look this over as well.

turbomam:
I gave both example files final new lines and escaped the literal `.` in the validation pattern. I added issues to generalize those

- https://github.com/microbiomedata/nmdc-schema/pull/3187/changes/6f521b3d85451753aa6dc85279117a4d2e575f38
- https://github.com/microbiomedata/nmdc-schema/issues/3192
- https://github.com/microbiomedata/nmdc-schema/issues/3193

src/schema/external_identifiers.yaml
samobermiller: I also remain confused about the connection to manifest, and how this will be invoked when running sra data in automation. but the slots added here still make sense, so I don't think it should hold anything up

samobermiller (approved):

sujaypatil96 (approved):
I like how comprehensive this PR is:

- [x] Added new slot `insdc_run_identifiers` and most populated possible metaslots
- [x] Asserted slot on `DataObject` class
- [x] Added prefix expansion
- [x] Added valid and invalid examples

src/schema/external_identifiers.yaml
```diff
@@ -321,11 +321,23 @@ slots:
...
+    see_also:
+      - https://www.ncbi.nlm.nih.gov/sra/docs/submitmeta/#sra-metadata-experiment
```
turbomam: for `insdc_run_identifiers`, wouldn't this be a better link?

```suggestion
      - https://www.ncbi.nlm.nih.gov/sra/docs/submitmeta/#linking-metadata-and-data-run
```

turbomam (approved):
I have a suggested change to the `see_also`s but otherwise this is good

https://github.com/microbiomedata/nmdc-schema/pull/3187#discussion_r3430641780

PR #3208, opened 2026-06-18
Make yq download architechure-aware and version-pinned

### LINKED ISSUE #3207: Dockerfile downloads amd64 yq from latest; make it arch-aware and version-pinned
The `Dockerfile` downloads `yq` unconditionally as `yq_linux_amd64` and from `latest`. On Apple Silicon the container is arm64 (confirmed: server `os/arch linux/arm64`), so the amd64 binary runs only under emulation. A test build succeeded with `yq` emulated, so it is not a hard failure today, but the build should not depend on emulation being available, and the original amd64 assumption (#1744) predates common arm64 Macs.

Proposed: download the architecture-appropriate, version-pinned binary, e.g. `yq_linux_$(dpkg --print-architecture)` at a fixed release tag. Found while testing the Docker build during the 11.20.2 release.

turbomam:
On Apple Silicon the build container is arm64, so the hardcoded `yq_linux_amd64` ran only under emulation. This fetches the architecture-matching binary via `dpkg --print-architecture` and pins the release (v4.53.3, what `latest` resolved to during testing) so the build no longer depends on emulation being available or on `latest` drifting.

Closes #3207.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,8 @@ RUN apt-get update && \
 
 # Download and install yq.
 # Reference: https://github.com/mikefarah/yq#install
-RUN wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -O /usr/bin/yq && \
+RUN ARCH="$(dpkg --print-architecture)" && \
+    wget "https://github.com/mikefarah/yq/releases/download/v4.53.3/yq_linux_${ARCH}" -O /usr/bin/yq && \
     chmod +x /usr/bin/yq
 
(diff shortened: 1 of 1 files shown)


turbomam:
@codex review

eecavanna (approved):
I didn't test this (I'm reviewing it on iPhone). The problem and solution make sense to me, and nothing in the diff looks problematic to me.

pkalita-lbl (approved):

eecavanna:
Approved and ready for merge.

PR #3214, opened 2026-06-18
3213 update LibraryPreparation modeling for insdc compatibility and example data for MicroFlora Danica

aclum:
This PR updates the modeling of library preparation to add slots from MixS and NCBI and adds an example data from MicroFlora Danica with suggested naming conventions for `name` and `description` slots

PR #3215, opened 2026-06-23
INSDC library modeling: pin SRA source, enforce + document the SRA DataObject pattern

turbomam:
Follow-ups on top of https://github.com/microbiomedata/nmdc-schema/pull/3214 (3213 update LibraryPreparation modeling for insdc compatibility and example data for MicroFlora Danica) to make the INSDC library modeling align with the SRA standard and be clear to schema users for interchange and validation. Targets your branch so you stay the owner of the feature.

### What and why

- **Pin the SRA citation.** The `see_also` links pointed at the `master` branch of `enasequence/webin-xml` (a moving target). Pinned to tag `2.1.0`. I fetched the XSD at that tag and confirmed it resolves with the same enum values that the schema mirrors.

- **Enforce the SRA retrieval contract.** Added a rule on `DataObject`: when `data_object_type` is `SRA toolkit-accessible sequence data`, `insdc_run_identifiers` is required. Before this, the schema allowed such an object with neither a URL nor an accession, leaving no way to retrieve it. This uses the same `preconditions`/`postconditions` idiom already in the schema (DOI and chromatography rules). It only fires for the new SRA type, so it does not affect existing data.

- **Document the pattern for users.** Added a comment on the enum value explaining how to model SRA-fetched data: one run-level `DataObject`, accession in `insdc_run_identifiers`, `url`/`md5_checksum`/`file_size_bytes` omitted because retrieval is by accession through the SRA Toolkit, not by a stable URL.

- **Mark the partial enums deliberate.** Noted that `LibraryStrategyEnum` and `LibrarySelectionEnum` are curated subsets of the INSDC vocabulary. Verified against the XSD: `LibrarySourceEnum` already matches it exactly (9/9); strategy is 4/40 and selection is 6/31, with exact token casing in all cases.

- **Typos + example alignment.** Fixed `avaliable` and `rRN`.

### Please confirm: the example DataObject rename

I changed the example DataObject from `name: file manfiest for run...` / `Manifest file for the SRR27943095 run` to a sequence-data wording, because the name and description said "manifest" while `data_object_type` says `SRA toolkit-accessible sequence data`, and a manifest (a listing of files) is a different thing from the sequence data. You have called it a "manifest" consistently since the first commit, so if you genuinely meant a manifest file, the right fix is the opposite of mine: change or add a `data_object_type` rather than the name. Your call.

### Not done locally

I did not run `make all && make test`; CI is the gate. The new rule is satisfied by the example (it has `insdc_run_identifiers`).

### DIFF
--- a/src/schema/basic_classes.yaml
+++ b/src/schema/basic_classes.yaml
@@ -536,6 +536,18 @@ classes:
         required: true
       data_category:
         required: true
+    rules:
+      - title: sra_toolkit_object_requires_run_identifier
+        description: >-
+          If data_object_type is "SRA toolkit-accessible sequence data", then insdc_run_identifiers is required, because such objects are retrieved by INSDC run accession through the SRA Toolkit rather than by a stable URL.
+        preconditions:
+          slot_conditions:
+            data_object_type:
+              equals_string: SRA toolkit-accessible sequence data
+        postconditions:
+          slot_conditions:
+            insdc_run_identifiers:
+              required: true
 
   DataEmitterProcess:
     class_uri: nmdc:DataEmitterProcess
--- a/src/schema/basic_slots.yaml
+++ b/src/schema/basic_slots.yaml
@@ -944,7 +944,6 @@ enums:
         description: Files that are available for download via SRA Toolkit by providing an INSDC accession
         comments:
            - File format will depend on options specified to SRA Toolkit
-           - Represent such an object as a single run-level DataObject. Put the run accession in insdc_run_identifiers. url, md5_checksum, and file_size_bytes are typically omitted and are not required, because retrieval is by accession through the SRA Toolkit rather than a single stable URL; they may still be recorded for specific files produced after retrieval.
         see_also:
           - https://github.com/ncbi/sra-tools/wiki
 
--- a/src/schema/basic_slots.yaml
+++ b/src/schema/basic_slots.yaml
@@ -944,7 +944,7 @@ enums:
         description: Files that are available for download via SRA Toolkit by providing an INSDC accession
         comments:
            - File format will depend on options specified to SRA Toolkit
-           - Represent such an object as a single run-level DataObject. Put the run accession in insdc_run_identifiers; url, md5_checksum, and file_size_bytes do not apply because retrieval is by accession through the SRA Toolkit, not by a stable URL.
+           - Represent such an object as a single run-level DataObject. Put the run accession in insdc_run_identifiers. url, md5_checksum, and file_size_bytes are typically omitted and are not required, because retrieval is by accession through the SRA Toolkit rather than a single stable URL; they may still be recorded for specific files produced after retrieval.
         see_also:
           - https://github.com/ncbi/sra-tools/wiki
 
--- a/src/schema/basic_slots.yaml
+++ b/src/schema/basic_slots.yaml
@@ -57,19 +57,19 @@ slots:
     description: Sequencing technique intended for this library
     range: LibraryStrategyEnum
     see_also:
-      - https://github.com/enasequence/webin-xml/blob/master/src/main/resources/uk/ac/ebi/ena/sra/schema/SRA.experiment.xsd
+      - https://github.com/enasequence/webin-xml/blob/2.1.0/src/main/resources/uk/ac/ebi/ena/sra/schema/SRA.experiment.xsd
 
   library_source:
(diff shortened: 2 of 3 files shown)


src/schema/basic_slots.yaml
```diff
@@ -937,9 +941,10 @@ enums:
           - https://github.com/ncbi/sra-tools/wiki/HowTo:-fasterq-dump
 
       SRA toolkit-accessible sequence data:
-        description: Files that are avaliable for download via SRA Toolkit by providing an INSDC accession
+        description: Files that are available for download via SRA Toolkit by providing an INSDC accession
         comments:
            - File format will depend on options specified to SRA Toolkit
+           - Represent such an object as a single run-level DataObject. Put the run accession in insdc_run_identifiers; url, md5_checksum, and file_size_bytes do not apply because retrieval is by accession through the SRA Toolkit, not by a stable URL.
```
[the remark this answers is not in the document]
turbomam: Good catch, softened in 87e8c7e52. The guidance now says url, md5_checksum, and file_size_bytes are typically omitted and not required (not that they are invalid), and notes they may still be recorded for specific files produced after retrieval.

turbomam:
@sujaypatil96 @aclum what do you think of the changes I added here on top of 
- https://github.com/microbiomedata/nmdc-schema/pull/3214

src/schema/basic_slots.yaml
```diff
@@ -937,9 +941,10 @@ enums:
           - https://github.com/ncbi/sra-tools/wiki/HowTo:-fasterq-dump
 
       SRA toolkit-accessible sequence data:
-        description: Files that are avaliable for download via SRA Toolkit by providing an INSDC accession
+        description: Files that are available for download via SRA Toolkit by providing an INSDC accession
         comments:
            - File format will depend on options specified to SRA Toolkit
+           - Represent such an object as a single run-level DataObject. Put the run accession in insdc_run_identifiers. url, md5_checksum, and file_size_bytes are typically omitted and are not required, because retrieval is by accession through the SRA Toolkit rather than a single stable URL; they may still be recorded for specific files produced after retrieval.
```
aclum: I recommend removing this extra comment

src/schema/basic_classes.yaml
```diff
@@ -536,6 +536,18 @@ classes:
         required: true
       data_category:
         required: true
+    rules:
```
aclum, line 539: this is a good addition

aclum (approved):

src/schema/basic_classes.yaml
sujaypatil96: Requiring `insdc_run_identifiers` when `data_object_type="SRA toolkit-accessible sequence data"` makes sense to me too.

src/schema/basic_slots.yaml
turbomam: Removed in 19dca6876.

src/schema/basic_slots.yaml
```diff
@@ -57,19 +57,19 @@ slots:
...
     see_also:
-      - https://github.com/enasequence/webin-xml/blob/master/src/main/resources/uk/ac/ebi/ena/sra/schema/SRA.experiment.xsd
+      - https://github.com/enasequence/webin-xml/blob/2.1.0/src/main/resources/uk/ac/ebi/ena/sra/schema/SRA.experiment.xsd
 
   library_selection:
     description: Library selection or enrichment method used
     range: LibrarySelectionEnum
     see_also:
-      - https://github.com/enasequence/webin-xml/blob/master/src/main/resources/uk/ac/ebi/ena/sra/schema/SRA.experiment.xsd
+      - https://github.com/enasequence/webin-xml/blob/2.1.0/src/main/resources/uk/ac/ebi/ena/sra/schema/SRA.experiment.xsd
```
sujaypatil96, line 72: Awesome, using tags to point to repo files is better practice (than using links to branches), so +1 for this change too!

src/schema/basic_slots.yaml
sujaypatil96: Yup, agreed! PV `comments` is not the best place for these notes. The translator/ETL pipeline using this PV should document the use of it as needed.

sujaypatil96 (approved):
Agreed with a lot of changes on this PR. Left a couple of notes, but everything LGTM! 🚀

PR #3218, opened 2026-06-23
Retarget the post-11.20.2 no-op migrator to 11.21.0

turbomam:
The next release is a minor bump to **11.21.0**, not a patch (11.20.3), because it adds new modeling (library-preparation slots and enums in #3214, plus a conditional DataObject rule in #3215).

#3216 added a no-op migrator named `migrator_from_11_20_2_to_11_20_3`. This renames it to `migrator_from_11_20_2_to_11_21_0` and sets `_to_version = "11.21.0"` so the chain link matches the version that will actually be released. The two backfill migrators (11.20.0 to 11.20.1, 11.20.1 to 11.20.2) are unchanged.

Still a no-op: every change from 11.20.2 to 11.21.0 is additive, so no data valid under 11.20.2 becomes invalid. Should merge before the v11.21.0 release so no phantom 11.20.3 migrator remains on main.

sorry about that, @eecavanna

### DIFF
--- a/nmdc_schema/migrators/migrator_from_11_20_2_to_11_21_0.py
+++ b/nmdc_schema/migrators/migrator_from_11_20_2_to_11_21_0.py
@@ -5,14 +5,14 @@ class Migrator(MigratorBase):
     r"""
     Migrates a database between two schemas.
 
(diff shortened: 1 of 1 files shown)


eecavanna (approved):
Looks good to me. Thanks for doing this.

Issue #3227, opened 2026-06-25
Add a `badges` slot to Biosample and Study for metadata-quality recognition


turbomam:
We want to recognize biosamples and studies that meet defined metadata-quality bars, as an alternative to a single numeric score or star rating. This is the concrete schema mechanism under the metadata quality rating system in https://github.com/microbiomedata/issues/issues/1583, and the successor to the older star idea in https://github.com/microbiomedata/mixs-6-2-release-candidate/issues/48. Score-based prior art: the Metadata Coverage Index (Liolios et al. 2012, https://pmc.ncbi.nlm.nih.gov/articles/PMC3558968/), a plain field-fill percentage. A badge is more expressive: it names a topic and a level instead of collapsing to one number.

Proposal: add a `badges` slot to Biosample and Study.

Implemented in https://github.com/microbiomedata/nmdc-schema/pull/3230 (slot, enum, badge subsets) plus an attachment PR (#3232 or #3233).

First PR scope: start with a multivalued enum slot. The initial implementation adds `badges` as a multivalued slot whose range is a new enum, and this repo provides that initial enum: a flat list of topic-plus-level values, e.g. `biogeochemistry_silver`, `biogeochemistry_gold`, `land_use_and_vegetation_silver`. Multivalued so a record can hold more than one badge (different topics, or a topic at a level).

Open design question (beyond the first PR): the longer-term range.

- Option A (the starting point): enum of flat badge values. Simple and queryable, but combinatorial as topics x levels grow.
- Option B: a list of a small wrapper class (e.g. `Badge`) with `topic`, `level`, and possibly `awarded_by`, `date`, and the slot group judged against. Avoids the combinatorial enum and supports per-badge provenance.

We start with Option A and can migrate to Option B later if topics and levels grow independently.

## Permissible values mirror the subsets

The enum is organized by topic, and each topic is exactly a badge subset defined in https://github.com/microbiomedata/nmdc-schema/issues/3228. A permissible value encodes a topic and a level, so a PV is `<subset_topic>_<level>`. Starting from the two first-PR subsets in 3228:

- `biogeochemistry` -> `biogeochemistry_silver`, `biogeochemistry_gold`
- `land_use_and_vegetation` -> `land_use_and_vegetation_silver`, `land_use_and_vegetation_gold`

[...]

PR #3230, opened 2026-06-26
Add badge subsets, MetadataBadgeEnum, and the badges slot

turbomam:
Defines the schema terminology for biosample metadata-quality badges: the badge subsets, the enum, the `badges` slot, its attachment to `Biosample`, and a valid example. It does not compute badges; award logic belongs in the runtime.

Revised 2026-08-13 to match the metadata quality squad decisions of 2026-08-05 and 2026-08-12 and @bmeluch's review, so it no longer resembles its original description:

- Badges are boolean. The silver and gold levels are gone.
- Ships the three badges scoped for this quarter: `biogeochemistry`, `host_information`, `expert_curation`. `land_use_and_vegetation` and the other candidate axes move to a follow-up.
- `biogeochemistry` (44 slots) and `host_information` (37 slots) are completeness badges, each declared as a subset carrying a `badge_minimum_slots` annotation of 2. Membership is the squad's Badge Subsets Scratchpad supersets, each slot verified as a `Biosample` slot. Twelve members are NMDC-native and carry `in_subset` in their own definitions, since the yq customization file only rewrites `mixs.yaml`.
- `expert_curation` is awarded from `ProvenanceMetadata.source_system_of_record`, not from slot completeness, so it has an enum value but no subset and no bar. The sync test exempts it through an explicit list rather than a silent gap.

The bar is an absolute count of populated slots and never a proportion of subset size, so adding a slot to a subset can only make a badge easier to earn. Without that property a subset that grows would revoke badges records already display, which users would read as a badge vanishing for no reason.

`badges` is attached to `Biosample` only. Study badges are out of scope for this quarter, so the `Study` attachment from the original plan is deliberately absent.

Three decisions are recorded as issues rather than settled here, because they are squad calls and none of them blocks the terminology:

- https://github.com/microbiomedata/nmdc-schema/issues/3339 whether required slots belong in badge subsets. Two `host_information` members are `required: true` in submission-schema's `PlantAssociatedInterface`.
- https://github.com/microbiomedata/nmdc-schema/issues/3340 the bar of 2 needs a per-biosample distribution against these subset sizes.
- https://github.com/microbiomedata/nmdc-schema/issues/3341 `src/schema/subsets.md` is maintainer-facing, so it sits outside `src/docs/` to stay off the published site.

Closes https://github.com/microbiomedata/nmdc-schema/issues/3228 (Define slot groups (subsets) that badge criteria are evaluated against)

[...]

PR #3233, opened 2026-06-26
Attach badges slot to ProvenanceMetadata (alternative)

turbomam:
Adds the `badges` slot to ProvenanceMetadata. Records reach badges through their
`provenance_metadata`.

## Relationships
- Requires https://github.com/microbiomedata/nmdc-schema/pull/3230 (defines the slot, enum, and subsets); merge that first. This PR is based on that branch.
- Alternative to https://github.com/microbiomedata/nmdc-schema/pull/3232 (badges on Biosample and Study). Do not merge both.
- Best combined with https://github.com/microbiomedata/nmdc-schema/pull/3231 (provenance_metadata on NamedThing): without it, badges reach only the classes that currently carry provenance_metadata; with it, badges reach every NamedThing.

Closes https://github.com/microbiomedata/nmdc-schema/issues/3227 (together with https://github.com/microbiomedata/nmdc-schema/pull/3230)


turbomam:
Scope caveat for this variant. Combined with #3231 (provenance_metadata on NamedThing), attaching badges to ProvenanceMetadata makes the `badges` slot reachable on all 55 NamedThing subclasses, not just submitter records. That includes ontology-term classes where a metadata-quality badge has no meaning (OntologyClass, NcbiTaxon, FunctionalAnnotationTerm, OrthologyGroup) plus the process, configuration, and data classes (DataObject, Configuration, the various *Process classes, etc.). Badges were conceived for Biosample and Study.

So this route is the more powerful one but over-broad. The direct-attachment alternative #3232 scopes badges to exactly Biosample and Study. If we keep the provenance route, we would likely want to constrain where `badges` is meaningful rather than leave it on every NamedThing. Recording this to inform the #3232 vs #3233 choice.

turbomam:
Closing in favor of the scoped approach in https://github.com/microbiomedata/nmdc-schema/pull/3232 (Attach badges slot to Biosample and Study). This PR attached badges to ProvenanceMetadata as the broader alternative; we are going with direct attachment to the record classes instead. Branch left in place for now in case we want to revisit.

PR #3265, opened 2026-07-10
Potential fix for 1 code quality finding

turbomam:
_This PR applies 1/1 suggestions from code quality [AI findings](https://github.com/microbiomedata/nmdc-schema/security/quality/ai-findings)._


turbomam (commented):
Verified, this is a correct fix. `make_document` currently ends its output with a stray blank line because `md_footer` is an empty string and still gets joined with `\n\n`; filtering out empty parts removes it. The `make_document` doctest only asserts the return type, so it still passes. Safe to merge.

PR #3267, opened 2026-07-10
Remove inline example from sample storage temperature description

Yurii201811:
## Summary

- add a durable MIxS customization for `samp_store_temp` that keeps its description definition-only
- retain `-80 Cel` in the slot’s structured `examples` field
- leave generated `src/schema/mixs.yaml` untouched, following the repository’s current generated-schema policy

Closes #2466

## Verification

Applied the exact new expression from `assets/yq-for-mixs-customizations.txt` to a local copy of the current generated schema with yq 4.53.3, then asserted:

- description: `Temperature at which the sample was stored.`
- first example: `-80 Cel`
- `git diff --check`

PR #3275, opened 2026-07-13
Remove redundant inline examples from MIxS slot descriptions

turbomam:
## Why

Follows the convention added in #3273 (CONTRIBUTING.md, "Put example values in the `examples` metaslot, not in prose"), applied to MIxS-derived slots. Where a description repeats a value that is already in the slot's `examples`, the copy in the prose is a duplicate and is removed.

## What

Via `assets/yq-for-mixs-customizations.txt` (MIxS slots are changed there, not by hand-editing generated `mixs.yaml`), strip the redundant inline value from eight slot descriptions whose `examples` already carry it:

`basin`, `root_med_carbon`, `root_med_solid`, `samp_store_temp`, `api`, `root_med_macronutr`, `root_med_micronutr`, `root_med_suppl`.

`samp_store_temp` is the slot from the externally-opened, now-closed #3267. `src/schema/mixs.yaml` is regenerated; the MIxS source is pinned (GSC mixs v6.3.0), so the diff is exactly these eight descriptions.

## Deliberately not in scope

This PR **adds no examples and changes no example types**, to stay clear of the range/type audit in #2950. Left for that work:

- Adding examples to slots that only carry the value inline and have no `examples` block (`field`, `reservoir`, `samp_well_name`, and the measurement slots).
- Fixing examples whose value disagrees with the description (`iwf`, `root_med_ph`, `root_med_regl`, `samp_transport_cond`) and the `elev` example (`100 meter` on a `float` range).
- Retyping scalar examples on wrapper-class ranges. Note the roadmap here: `TextValue` is slated to be replaced by plain `string`, and `TimestampValue` by JSON-friendly temporal ranges or rigorously-patterned strings. Their current scalar `value:` examples are therefore already forward-compatible and should not be objectified. The real objectification target is the `QuantityValue` slots (plus a decision on `ControlledTermValue`/`GeolocationValue`).

Draft pending review.

### DIFF
--- a/assets/yq-for-mixs-customizations.txt
+++ b/assets/yq-for-mixs-customizations.txt
@@ -587,20 +587,23 @@
 'del(.slots.host_body_product.examples)'
 '.slots.host_body_product.examples.[0].value = "mucus [UBERON:0000912]"'
 
-# Relocate misplaced example values out of descriptions into the examples metaslot.
-# Scoped to TextValue name/label slots; see CONTRIBUTING.md "Put example values in the
-# examples metaslot, not in prose". Measurement/QuantityValue slots are deliberately
-# excluded here: their example handling belongs with the units work (issue #2950).
-'.slots.field.description |= "Name of the hydrocarbon field"'
-'.slots.field.examples = [{"value": "Albacora"}]'
-'.slots.reservoir.description |= "Name of the reservoir"'
-'.slots.reservoir.examples = [{"value": "Carapebus"}]'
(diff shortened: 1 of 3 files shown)


turbomam:
Thanks to @Yurii201811 for the nudge to do this

turbomam:
Superseded by a cleaner split: docs rendering is #3276, the redundant-inline-example strips move to a dedicated PR, and objectification + unit/discrepancy work follow as separate PRs.

Issue #1233, opened 2023-10-25
MVP polymorphic planned_process_set


turbomam:
also removing all `slot_usage`s and these slots from `Database`
- date_created
- etl_software_version
----

see also
- #1153 
- #1197 
- #1154

turbomam:
remove directly asserted metaslot values from slots that mix in `object_set`? Try on `biosample_set`.

turbomam:
GOLD has some `..._type` slots that we shouldn't mess with, in addition to the MIXS  `..._type` slots

Most of these could be renamed following a `..._category` pattern. They should all use enumerated ranges.

- neon*type
----

- analysis_type
- compression_type
- container_type
- data_object_type
- device_type
- dna_cont_type
- feature_type
- library_type
- omics_type
- rna_cont_type
- sample_type

Issue #1048, opened 2023-07-26
all classes should have a required type-designating slot, with range CURIe and `designates_type` = True


turbomam:
make these assertions as high in the class hierarchy as possible

turbomam:
may need to restructure hierarchy

turbomam:
also review any nmdc-schema (not MIxs) slots whose name includes the word 'type'

turbomam:
This may decrease the number of `..._set` slots required for class `Database` and therefore the number of MongoDB collections.

But that would require a review of code that infers the type of something based on the collection that it came from.

aclum:
overlap with  #1030

mslarae13:
@turbomam  is this still in progress or is it done?

mslarae13:
@turbomam status update?

turbomam:
> @turbomam status update?

It's merged into the main branch of berkeley-schema-fy24

Issue #248, opened 2022-01-26
deprecate type slot


wdduncan:
During the January release, we encountered problems due the `type` slot not have a value.

Kitware reads the values of the `type` slot when populating the the postgres tables. The `type` slot is also required for change sheets to work.

So, make the `type` slot required.

Also, create enums to constrain the range of the `type` slot.

cc @dehays @turbomam @sujaypatil96

wdduncan:
Think about this a bit ... we could use the `pattern` attribute to constrain the values.

wdduncan:
Per meeting today:
We are not going to require slots, but we should create enums for the range of the `type` slot.

wdduncan:
mistakenly closed

wdduncan:
After discussing this more with @dehays, we think it is best to deprecate the `type` slot. The reason is there is inconsistent usage and confusion about what the values should be.   

Originally, the `type` slot was used to capture the information about the python class used to create the instance data. This allowed me to reference the type of python object that was used to create a JSON record.  

When a linkml schema (such as the `nmdc-schema`) is converted into python, the linkml class names are convert into python classes with named changed into [PascalCase](https://techterms.com/definition/pascalcase) and prefixed with nmdc. For example, the `nom analysis activity` class in the `nmdc-schema` is converted into `NomAnalysisActivity` in the `nmdc.py` file. In the type field, this is class is then represented as `nmdc:NomAnalyisActivity` in order to make it explicit that the class comes from the nmdc namespace. However, this usage seems to be causing confusion.

PR #3352, opened 2026-08-14
Remove see_also links that point at closed issues

### LINKED ISSUE #3345: Remove see_also links that point at closed issues (19 entries)
The merged schema carries 19 `see_also` entries that point at closed issues and pull requests. A reader who follows one from a documentation page lands on something closed, with no indication of whether the link explains the current design or records an abandoned proposal.

| element kind | element | link |
| --- | --- | --- |
| class | Culturing | https://github.com/microbiomedata/nmdc-schema/issues/2978 |
| class | Isolation | https://github.com/microbiomedata/nmdc-schema/issues/2978 |
| class | NcbiTaxon | https://github.com/microbiomedata/nmdc-schema/issues/2959 |
| class | NcbiTaxon | https://github.com/microbiomedata/nmdc-schema/issues/2971 |
| class | Organism | https://github.com/microbiomedata/nmdc-schema/issues/2959 |
| class | Organism | https://github.com/microbiomedata/nmdc-schema/issues/2971 |
| class | OrganismSample | https://github.com/microbiomedata/nmdc-schema/issues/2961 |
| slot | classified_as | https://github.com/microbiomedata/nmdc-schema/issues/2959 |
| slot | classified_as | https://github.com/microbiomedata/nmdc-schema/issues/3016 |
| slot | gold_organism_identifiers | https://github.com/microbiomedata/nmdc-schema/issues/2973 |
| slot | salinity_category | https://github.com/microbiomedata/nmdc-metadata/pull/297 |
| slot | type | https://github.com/microbiomedata/nmdc-schema/issues/1048 |
| slot | type | https://github.com/microbiomedata/nmdc-schema/issues/1233 |
| slot | type | https://github.com/microbiomedata/nmdc-schema/issues/248 |
| subset | badge_topic | https://github.com/microbiomedata/nmdc-schema/issues/3228 |
| subset | biogeochemistry | https://github.com/microbiomedata/nmdc-schema/issues/3228 |
| subset | biogeochemistry | https://github.com/microbiomedata/nmdc-schema/issues/3326 |
| subset | host_information | https://github.com/microbiomedata/nmdc-schema/issues/3228 |
| subset | host_information | https://github.com/microbiomedata/nmdc-schema/issues/3326 |

Five other `see_also` GitHub links point at issues that are still open and are not in scope here.

**Proposed**: remove them from the schema, not from the documentation templates, so the change is visible in review and travels with the element it describes.

Two cases deserve a second look before deletion rather than a blanket sweep, because the link is doing work a reader cannot get elsewhere:

- `type` carries three closed links (https://github.com/microbiomedata/nmdc-schema/issues/248 deprecate type slot, https://github.com/microbiomedata/nmdc-schema/issues/1048 all classes should have a required type-designating slot, https://github.com/microbiomedata/nmdc-schema/issues/1233 MVP polymorphic planned_process_set). Together they are the reason `type` exists in its current form. If they go, the reasoning should land in the slot's `comments` instead of vanishing.
- The five badge subset links point at issues closed *by* the work that created those subsets. They are provenance for a design decision made yesterday, which is the most defensible kind of closed link.

A rule worth setting at the same time, since this will recur: `see_also` is for a reader who wants more context now, so it should point at durable material (a doc page, a specification, an ontology term) rather than at issue threads. Where an issue is the only record, its content belongs in `comments` on the element.

Related: a check that flags `see_also` links resolving to closed issues would keep this from accumulating again, though it needs network access and would not suit every CI job.

turbomam:
Removes the 19 `see_also` entries listed in https://github.com/microbiomedata/nmdc-schema/issues/3345, all of which pointed at closed issues or pull requests.

Distinct GitHub references in the merged schema go from 19 to 8. Ten `see_also` keys had no surviving items and were dropped with them. All edits are in hand-maintained source files; none of these references lived in the generated `mixs.yaml`, so no customization lines or regeneration were involved.

**The `type` slot's reasoning is preserved rather than deleted.** Its three links (https://github.com/microbiomedata/nmdc-schema/issues/248 deprecate type slot, https://github.com/microbiomedata/nmdc-schema/issues/1048 all classes should have a required type-designating slot, https://github.com/microbiomedata/nmdc-schema/issues/1233 MVP polymorphic planned_process_set) were together the only record of why the slot exists in its current form. That is now a comment on the slot:

> Deprecating this slot was proposed and rejected: without it, a document read back from a polymorphic MongoDB collection cannot be resolved to the class it instantiates. It is required on every class for that reason, rather than as a convention inherited from LinkML.

**One entry was deliberately kept and needs a reviewer's opinion.** `emsl_project_identifiers` has:

```yaml
see_also:
  - https://github.com/microbiomedata/nmdc-schema/issues/927#issuecomment-1802136437
```

The issue is closed, so the rule as stated would remove it, but this is a deep link to a specific comment rather than to a thread, and that comment is where the reasoning for the identifier pattern actually lives. Deleting it loses content with nowhere else to point. Happy to remove it if reviewers prefer the rule applied uniformly; it is a one-line change.

Three closed-issue references remain in the schema and are out of scope, because they sit in prose rather than in `see_also`: two in class comments and one inside a `description`. Those read as narrative context and are not presented to a user as a link to follow.

Verification: every file in `src/schema/` parses, the schema regenerates, and the full test suite passes (87 tests).

Closes https://github.com/microbiomedata/nmdc-schema/issues/3345

### DIFF
--- a/src/schema/attribute_values.yaml
+++ b/src/schema/attribute_values.yaml
@@ -259,10 +259,12 @@ slots:
     notes:
       - makes it easier to read example data files
       - required for polymorphic MongoDB collections
-    see_also:
-      - https://github.com/microbiomedata/nmdc-schema/issues/1048
-      - https://github.com/microbiomedata/nmdc-schema/issues/1233
-      - https://github.com/microbiomedata/nmdc-schema/issues/248
+    comments:
+      - >-
+        Deprecating this slot was proposed and rejected: without it, a document
+        read back from a polymorphic MongoDB collection cannot be resolved to the
+        class it instantiates. It is required on every class for that reason,
+        rather than as a convention inherited from LinkML.
     structured_aliases:
       - literal_form: workflow_execution_class
         contexts:
--- a/src/schema/basic_classes.yaml
+++ b/src/schema/basic_classes.yaml
@@ -106,9 +106,6 @@ classes:
         with range OntologyClass. The intent is that these slots point to NcbiTaxon instances specifically,
         but LinkML does not currently support constraining the term slot of a ControlledIdentifiedTermValue
         to a particular OntologyClass subclass.
-    see_also:
-      - https://github.com/microbiomedata/nmdc-schema/issues/2959
-      - https://github.com/microbiomedata/nmdc-schema/issues/2971
     close_mappings:
       - biolink:OrganismTaxon
 
@@ -269,9 +266,7 @@ classes:
     exact_mappings:
       - COB:0000022
     see_also:
-      - https://github.com/microbiomedata/nmdc-schema/issues/2959
       - https://github.com/microbiomedata/nmdc-schema/issues/2803
-      - https://github.com/microbiomedata/nmdc-schema/issues/2971
 
   PlannedProcess:
     description: >-
@@ -718,9 +713,6 @@ slots:
         The global range stays OntologyClass. Organism is currently the only class that
         narrows classified_as to NcbiTaxon, via slot_usage. Extending that narrowing to
         other classes is tracked in https://github.com/microbiomedata/nmdc-schema/issues/3016.
-    see_also:
-      - https://github.com/microbiomedata/nmdc-schema/issues/2959
-      - https://github.com/microbiomedata/nmdc-schema/issues/3016
     narrow_mappings:
       - biolink:in_taxon
 
--- a/src/schema/basic_slots.yaml
+++ b/src/schema/basic_slots.yaml
@@ -64,8 +64,6 @@ subsets:
         permissible value has a subset here. expert_curation is awarded from
         ProvenanceMetadata.source_system_of_record rather than from slot
         completeness, so it has no badge_topic subset.
-    see_also:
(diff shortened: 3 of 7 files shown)


turbomam:
Both suppressed comments looked at, one taken and one declined.

**Taken**: the `euryhaline` typo. Fixed. `salinity_category` was already in this diff, so it belongs here.

**Declined for this PR**: switching `structured_aliases` from `contexts` to `source`. The premise that `source` is what the schema uses elsewhere is only half right. Both are established here: 25 entries use `source`, 14 use `contexts`, and the `contexts` group is internally consistent (the JGI JAT macro file, and the Illumina instrument aliases in `basic_classes.yaml`).

They also mean different things in LinkML. `source` records where an alias came from; `contexts` records the settings in which the alias applies. A bitbucket YAML file that uses `workflow_execution_class` as its field name is arguably a context rather than a source, so this is a semantic call across 14 entries in several files, not a consistency cleanup.

Out of scope for a PR that removes closed `see_also` links. Worth its own issue if the convention should be settled; happy to file one.

PR #3396, opened 2026-09-09
Add new classes for SIP metagenomics data modeling

aclum:
This PR adds classes, slots, and example data pertaining to SIP metagenomics modeling. 
This modeling reuses slots defined https://doi.org/10.1093/gigascience/giae071
see also: [MIMS-MISIP](https://genomicsstandardsconsortium.github.io/mixs/0010014/) checklist

PR #3400, opened 2026-09-09
Get the SIP branch green: regenerate mixs.yaml, objectify examples, repair the import TSV

turbomam:
Fixups on https://github.com/microbiomedata/nmdc-schema/pull/3396, targeting that branch so @aclum can take or leave them. Both of its failing "Build and test" jobs are green here.

Five changes are mine rather than @aclum's and need her agreement:

1. **`isotopolog` is constrained** on `LabelingProcess`: `has_raw_value` required, and `term.id` must be `PUBCHEM.COMPOUND:<CID>` with no CID 0. Nothing was enforced before, not even an empty object. This uses `range_expression`, new to this schema, so it ships with two valid and four invalid examples.
2. **`isotopolog_dose` becomes a `QuantityValue`** with `[ppm]`, matching the other SIP measurements.
3. **`gradient_pos_density` stores `g/cm3`.** `g/mL` is not a `UnitEnum` value, so that line could never have worked.
4. **`g/mL` is added as an alias** of `g/cm3`, which touches a shared enum.
5. **The `MIXS:0001345` mapping is removed.** @turbomam's call; it mapped a class to a slot.

One correction worth naming: both new classes had a `class_uri` local name that did not match the class name, which is what made the type designator reject the example data.

The rest is mechanical. Regenerating `mixs.yaml` so the three yq lines from 2f51c1418 take effect, 14 malformed rows in the import TSV, and example data that could not have validated.

Happy to give the reasoning behind any of these.

https://claude.ai/code/session_01NDbXnpG2GdSN3Dq3sFuzPR


turbomam:
Update after the first push: the `MIXS:0001345` mapping is gone rather than relocated.

It arrived on this branch as a `narrow_mappings` on the `Protocol` class. `MIXS:0001345` is `sip_method`, which is a MIxS slot, and the repo conventions say a mapping relates elements of the same kind, so a class mapped to a slot is not allowed. My first pass moved it to the `protocol_link` slot, which `FractionationProcess` already inherits from `PlannedProcess`, since that is type-correct and preserves the intent. @turbomam decided the mapping is not wanted at all, so commit 417570124 removes it.

`src/schema/basic_classes.yaml` now differs from `main` only by the two `PlannedProcessTypeEnum` entries this feature adds. @aclum, if you want the alignment recorded in some other form, say so and it goes back in.

The body above has been updated to match.

assets/yq-for-mixs-customizations.txt
```diff
@@ -1001,6 +1007,12 @@
 '.slots.host_tot_mass.examples = [{"object": {"type": "nmdc:QuantityValue", "has_raw_value": "2500 g", "has_numeric_value": 2500, "has_unit": "g"}}]'
 '.slots.host_wet_mass.examples = [{"object": {"type": "nmdc:QuantityValue", "has_raw_value": "1500 g", "has_numeric_value": 1500, "has_unit": "g"}}]'
 '.slots.humidity_regm.examples = [{"object": {"type": "nmdc:TextValue", "has_raw_value": "25 gram per cubic meter;R2/2018-05-11T14:30/2018-05-11T19:30/P1H30M"}}]'
+'.slots.internal_standard.examples = [{"object": {"type": "nmdc:TextValue", "has_raw_value": "https://doi.org/10.1101/2022.12.20.521340"}}]'
+'.slots.isotopolog.examples = [{"object": {"type": "nmdc:ControlledTermValue", "has_raw_value": "toluene [PUBCHEM.COMPOUND:1140]", "term": {"id": "PUBCHEM.COMPOUND:1140", "type": "nmdc:OntologyClass"}}}]'
+'.slots.isotopolog_atom_frac.examples = [{"object": {"type": "nmdc:TextValue", "has_raw_value": "0.95"}}]'
```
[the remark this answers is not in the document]
turbomam: Correct, and fixed in 6ad8a8320. The yq customization now overrides the `comments` on `isotopolog` to keep the first sentence and drop "If it's an undefined mixture, enter 0", so the generated documentation no longer advertises a value that the NMDC form cannot accept.

Only the false instruction was removed. How NMDC should represent an undefined mixture is a real question and it is still open, listed in the pull request body for @aclum. `ControlledTermValue` leaves `term` optional, so a raw-value-only object is one candidate, but that is a modeling decision rather than a fixup.

assets/yq-for-mixs-customizations.txt
```diff
@@ -773,7 +774,8 @@
 '.slots.phosphate.annotations.storage_units = {"tag": "storage_units", "value": "umol/L"}'
 '.slots.silicate.annotations.storage_units = {"tag": "storage_units", "value": "umol/L"}'
 '.slots.isotopolog_incu_time.annotations.storage_units = {"tag": "storage_units", "value": "h"}'
-'.slots.gradient_pos_density.annotations.storage_units = {"tag": "storage_units", "value": "g/mL"}'
```
aclum, line 776: keep as g/mL, add as a permissible value instead of updating to g/cm3. g/mL for more appropriate for fluids.

aclum:
@turbomam  this is not a passing PR and the some of the invalid examples are bogus src/data/invalid/LabelingProcess-isotopolog-no-raw-value.yaml for example, it says has_raw_value is required when it isn't.

turbomam:
Both fair. Taking them in order.

The red CI was mine. I added `g/mL` as a permissible value per your review, and my edit left `g/cm3`'s `description` line under the new key, so the enum had two `description` keys and `linkml generate` died with `ValueError: Duplicate key: "description"`. Fixed in d17d19263. I pushed the broken commit because I read a piped grep's exit status instead of make's, which is on me.

On the invalid example, you are right and it is worth separating two things I bundled together.

Nothing in the schema requires `has_raw_value`. It is optional on `AttributeValue` and stays optional on `ControlledTermValue`, `ControlledIdentifiedTermValue` and `QuantityValue`. That file is invalid only because of the `range_expression` this PR adds, which requires it for `isotopolog` on `LabelingProcess` alone. So it is not testing schema policy, it is testing a rule I invented in the same PR, and I should have said so rather than filing it next to the namespace examples as though it were the same kind of thing.

The namespace half is the part I would defend. It stops `CHEBI:60004`, `pubchem.compound:1140` and `PUBCHEM.COMPOUND:0`, which is @cmungall's proposal from https://github.com/GenomicsStandardsConsortium/mixs/issues/577 applied here rather than waiting on MIxS.

The `has_raw_value` half only exists to give the undefined-mixture case somewhere to put the description, since `term` is omitted there. If you would rather not require it, I will drop that condition and the example with it, and the constraint becomes namespace-only. Your call.

src/data/invalid/LabelingProcess-isotopolog-cid-zero.yaml
aclum, line 1: remove this invalid example, there is no way for linkml to report this as invalid.

src/data/invalid/LabelingProcess-isotopolog-no-raw-value.yaml
aclum, line 1: remove, has raw_value is not required for ControlledTermValues

src/data/invalid/LabelingProcess-isotopolog-wrong-namespace.yaml
aclum, line 1: remove this example, because of the abstraction there is no way to put a constraint on the id value w/linkml

src/data/valid/LabelingProcess-isotopolog-undefined-mixture.yaml
aclum, line 1: Im not sure this is how we want to handle this case.

aclum (approved):

PR #3422, opened 2026-09-16
Rename and define the SIP classes and slots that nmdc-schema owns

turbomam:
Names and definitions for the SIP elements that nmdc-schema defines itself. Nothing imported from MIxS is renamed or redefined; changes to those belong upstream, as the typo fixes in https://github.com/GenomicsStandardsConsortium/mixs/issues/1408 do. All of these elements are new on this branch and appear in no release and no data, so renaming them costs nothing now and would cost a migrator later.

Each rename is its own commit, so any one of them can be dropped without touching the others.

- `IsotopeMetadata` to `IsotopologAddition`. Each entry is one isotopolog, or one heterogeneous material, added during the labeling, with the isotope it carries, its label status, the atom fraction and position of the label, and the dose. The name uses the word the sibling MIxS slots already use (`isotopolog_label`, `isotopolog_dose`, and so on), and it fits an entry at natural abundance and an entry that is cells or biomass rather than a substance. "Metadata" says nothing in a schema where every class is metadata, and the old definition ("metadata about an isotope labeling process") described the process rather than the entry.
- `isotope_label_metadata` to `isotopolog_additions`, one entry per isotopolog and isotope. The two SIP slot definitions also move out from under the header for legacy MIxS slots, where they had been placed.
- `stable_isotope_probe` to `stable_isotope_probing` in `ProtocolCategoryEnum`. The category names the technique, not a probe.
- `GradientFractionationProcess` to `DensityGradientFractionationProcess`. The class is defined as separation on a density gradient and carries the isopycnic centrifugation aliases; other fractionations, including the size exclusion used in proteomics, do not fit it, and the plainer name claimed them. The `ProtocolForEnum` value follows.

Definitions rewritten: `IsotopeLabelingProcess` said the process labels a substance, which is backwards (labeled substances are added to a sample so that isotope uptake can be traced), said "stable or radioactive" while `IsotopeEnum` holds only 13C, 15N, 2H and 18O, and did not cover natural-abundance controls. `DensityGradientFractionationProcess` now says what is separated, by what property, and that each output fraction records its gradient position.

One comment each on the MIxS slots `isotopolog_atom_frac` and `isotopolog_atom_pos`, added through the yq customizations: their MIxS descriptions call for pipe-separated values ordered like `isotopolog`, which is a spreadsheet convention; in NMDC each entry holds one value of each. The descriptions themselves are untouched.

This branch is based on https://github.com/microbiomedata/nmdc-schema/pull/3421 and carries its two commits until that one merges.

Validation: `make test` passes locally (95 tests), `linkml-lint` reports no problems, and every SIP example file validates under the new names. Generated artifacts under `nmdc_schema/` are not committed.


assets/yq-for-mixs-customizations.txt
```diff
@@ -456,7 +456,7 @@
 '.slots.watering_regm.range |= "TextValue"'
 '.slots.window_open_freq.range |= "TextValue"'
 '.slots.window_size.range |= "TextValue"'
-'.slots.isotopolog.range |= "ControlledTermValue"'
+'.slots.isotopolog.range |= "ControlledIdentifiedTermValue"'
```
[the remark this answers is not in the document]
turbomam: That commit belongs to https://github.com/microbiomedata/nmdc-schema/pull/3421, on which this PR is stacked; it shows here only until 3421 merges, and the scope statement in the body is about this PR's own commits. On migration: the `isotopolog` slot is in no release (v11.23.0 has no such slot), so no conformant data can hold the term-less shape. Same answer as on 3421.

src/schema/nmdc.yaml
```diff
@@ -1170,19 +1192,34 @@ enums:
...
+    range: IsotopologAddition
+    multivalued: true
+    inlined_as_list: true
+
+  hetero_isotopolog:
+    title: heterogeneous isotopolog
+    description: >-
+      A labeled input that has no PubChem compound identifier, described in free text: an undefined mixture such as root exudates or plant litter, or labeled cells or biomass such as T4 phages. A defined compound with a PubChem compound identifier goes in isotopolog instead.
+    comments:
+      - Exactly one of isotopolog and hetero_isotopolog is given on each IsotopologAddition entry; the rules on that class enforce it.
```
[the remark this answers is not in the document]
turbomam: Agreed. Fixed in d0ff50ae4, which describes the slot as an input rather than a labeled input and rewords the generated `isotopolog` comment the same way; that also covers the suppressed comment on assets/yq-for-mixs-customizations.txt:1018.

src/schema/nmdc.yaml
```diff
@@ -640,21 +641,42 @@ classes:
...
           slot_conditions:
             term:
               range_expression:
                 slot_conditions:
                   id:
                     pattern: '^PUBCHEM\.COMPOUND:[1-9]\d*$'
+    rules:
+      - title: hetero_isotopolog_required_when_isotopolog_absent
+        description: >-
+          If isotopolog is absent, then hetero_isotopolog is required, so that every entry names the input in one of the two ways.
```
turbomam, line 654: On the suppressed finding here: fixed in 295aa5c88. The rule description now says input rather than labeled input, since an entry can be at natural abundance.

src/schema/nmdc.yaml
```diff
@@ -640,21 +641,42 @@ classes:
...
+          slot_conditions:
+            isotopolog:
+              value_presence: ABSENT
+        postconditions:
+          slot_conditions:
+            hetero_isotopolog:
+              required: true
+      - title: hetero_isotopolog_absent_when_isotopolog_present
+        description: >-
+          If isotopolog is present, then hetero_isotopolog must be absent, so that no entry names the input in both ways.
```
turbomam, line 665: On the suppressed finding here: fixed in 295aa5c88, same wording change as the first rule.

src/schema/nmdc.yaml
```diff
@@ -1170,19 +1192,35 @@ enums:
         is_a: organic_layer
 
 slots:
+  # SIP (stable isotope probing) slots defined by NMDC. The MIxS MISIP slots they group are imported in mixs.yaml.
+  isotopolog_additions:
+    description: >-
+      The isotopologs, or heterogeneous materials, added during this isotope labeling process, one entry per isotopolog and isotope.
+    range: IsotopologAddition
+    multivalued: true
+    inlined_as_list: true
```
turbomam, line 1201: On the suppressed finding here: I marked `isotopolog_additions` recommended in 11dda857b, so a process record with no additions now draws a validation warning. I did not make it required. @aclum's to-do on https://github.com/microbiomedata/nmdc-schema/pull/3396 is to make `isotopolog` recommended rather than required because published studies often do not report the substrate, and a process record can exist before its additions are curated. The class definition describes what the process does, not what a record must contain. If we want to require at least one entry, that is `required: true` plus `minimum_cardinality: 1` and a counter-example, and it should be @aclum's call.

assets/yq-for-mixs-customizations.txt
```diff
@@ -1014,11 +1014,12 @@
...
+'.slots.isotopolog.comments = ["Give the PubChem compound as PUBCHEM.COMPOUND:<CID>, both in has_raw_value and in term.id. An input with no PubChem compound identifier, such as root exudates or cells, goes in hetero_isotopolog instead; NMDC does not use a sentinel identifier for that case."]'
 '.slots.isotopolog_atom_frac.description |= sub("should be orded the same"; "should be ordered the same")'
 '.slots.isotopolog_atom_pos.description |= sub("should be orded the same"; "should be ordered the same")'
+'.slots.isotopolog_atom_frac.comments += ["In NMDC each IsotopologAddition entry describes one added input, an isotopolog or a hetero_isotopolog, and carries at most one atom fraction, so the pipe-separated ordering described for spreadsheets does not apply."]'
+'.slots.isotopolog_atom_pos.comments += ["In NMDC each IsotopologAddition entry describes one added input, an isotopolog or a hetero_isotopolog, and carries at most one set of labeled atoms, so the pipe-separated ordering described for spreadsheets does not apply."]'
```
turbomam: On the suppressed finding here: fixed in d6a855bdc. Both comments now say that each `IsotopologAddition` entry describes one added input, which may be an `isotopolog` or a `hetero_isotopolog`, and carries at most one value of this slot.

src/schema/nmdc.yaml
```diff
@@ -606,9 +606,9 @@ classes:
     class_uri: nmdc:IsotopeLabelingProcess
     is_a: MaterialProcessing
     description: >-
-      A material processing that labels a substance with a stable or radioactive isotope(s).
+      A material processing in which substances or materials, isotopically labeled or at natural abundance, are added to a sample so that uptake of the isotope by the resident organisms can be traced.
```
turbomam, line 609: On the suppressed finding here: fixed in d6a855bdc. The definition no longer says one or more; it describes what the process does, and `isotopolog_additions` stays recommended, see the thread above on why it is not required.

src/schema/nmdc.yaml
```diff
@@ -640,21 +641,42 @@ classes:
...
+              value_presence: ABSENT
 
-  GradientFractionationProcess:
-    class_uri: nmdc:GradientFractionationProcess
+  DensityGradientFractionationProcess:
+    class_uri: nmdc:DensityGradientFractionationProcess
     is_a: MaterialProcessing
     description: >-
-      A material processing that separates a sample into multiple fractions based on a density gradient.
+      A material processing that separates a sample, typically extracted nucleic acids, into fractions by buoyant density, usually by isopycnic centrifugation. The output fractions can record their gradient position, buoyant density and relative nucleic acid amount.
```
turbomam, line 679: On the suppressed finding here: fixed in d6a855bdc. The definition now says the process separates a sample, typically extracted nucleic acids, so it no longer claims an input restriction that `has_input` does not enforce. Whether `has_input` should be narrowed to `procsm` is a modeling change for @aclum to decide.

turbomam, line 679: On the second suppressed finding at this line: fixed in d6a855bdc. The definition now says the output fractions can record gradient position, buoyant density and relative amount, rather than that each one does, since `gradient_position` is optional on `ProcessedSample`.

src/schema/nmdc.yaml
```diff
@@ -625,13 +625,14 @@ classes:
       has_output:
         required: true
 
-  IsotopeMetadata:
-    class_uri: nmdc:IsotopeMetadata
+  IsotopologAddition:
+    class_uri: nmdc:IsotopologAddition
```
turbomam, line 629: On the suppressed finding here: declined. A migrator in this repo transforms a database from one release of the schema to the next (nmdc_schema/migrators/README.md), and "current version" in CONTRIBUTING.md means the released schema, v11.23.0 today, which has none of these names. No database has ever been validated against `IsotopeMetadata` or `isotope_label_metadata`; they exist only in commits on the unreleased branch this PR targets. Renaming them before that branch reaches a release is the way to avoid a migrator, not a reason to write one.

PR #3441, opened 2026-09-23
Test that badge subsets only use what the nmdc-runtime badge job can read

### LINKED ISSUE #3440: Test that badge subsets only use what the nmdc-runtime badge job can read
The badge job in nmdc-runtime (https://github.com/microbiomedata/nmdc-runtime/pull/1631) reads the badge subsets from this schema. `tests/test_badge_subset_sync.py` checks the subsets against the enum, but not against what the job expects. Four gaps, checked on `main` at `c2ee3fc2d`:

1. The bar check uses `int(value)`, so `badge_minimum_slots: "2"` and `2.0` both pass. The job requires an actual integer and throws `ValueError` on either.
2. Nothing requires a badge subset member to be a Biosample slot. For a slot that isn't, `SchemaView.induced_slot(slot, "Biosample")` returns the default range `string` instead of failing (linkml-runtime 1.11.1), so the job would check a slot biosamples never have. None exist today.
3. Nothing limits the ranges of badge subset slots. The job has rules for `TextValue`, `QuantityValue`, `ControlledTermValue`, `ControlledIdentifiedTermValue`, `string`, `float` and enums, and throws on anything else. None exist today.
4. The job awards `expert_curation` when `source_system_of_record` equals the literal `NMDC_Submission_Portal`. Renaming that permissible value would stop the badge without any error.

Add a test for each. The failure messages for 3 and 4 should say that `nmdc_runtime/site/ops/badges.py` has to change too.

Job code: https://github.com/microbiomedata/nmdc-runtime/blob/f1f77cb4e93e2e21917d1fec7e448eca2ec673bd/nmdc_runtime/site/ops/badges.py
Related: https://github.com/microbiomedata/nmdc-runtime/issues/1633

turbomam:
Closes https://github.com/microbiomedata/nmdc-schema/issues/3440

The badge job added in https://github.com/microbiomedata/nmdc-runtime/pull/1631 reads the badge subsets from this schema, and a schema change it can't read would either stop the job or quietly skip a badge. These tests catch that at build time: an integer bar (the old check used `int()`, which accepts `"2"` and `2.0`; the job rejects both), members that are Biosample slots, ranges the job has a rule for, and the `NMDC_Submission_Portal` value the job matches for `expert_curation`.

Validation: each new check fails against a copy of the materialized schema mutated to break it, and `main`'s version of the test passes with the bar written as `"2"` or `2.0`. Full `pytest tests` passes locally.

Related: https://github.com/microbiomedata/nmdc-runtime/issues/1633


turbomam:
Replying to the two findings in the Copilot overview:

- Failure handling for invalid types: fixed in the latest commit. If `source_system_of_record` stopped being enum-ranged, `get_enum` raised `No such enum: "string"` and the test errored. It now fails with a message that names the badge job. Checked against a copy of the materialized schema with that range set to `string`.
- Scalar-string `in_subset`: not changed. SchemaView turns `in_subset: biogeochemistry` into `['biogeochemistry']`, so the slot is still counted as a member. Checked the same way, with `ph` given a scalar `in_subset`.

eecavanna:
Will look at this next quarter if badge stuff still in scope.

turbomam:
How coupled to the runtime is it really? It doesn't parse any of the runtime code. So that needs better documentation (about being used for "helpful" warning messages)

turbomam:
Is there any way that the schema could be more definitive about paths in a document that might be schema valid but still essentially "empty" for some purpose, like calculating badge eligibility?

